PR #16, opened 2017-08-17
circuit breaker implemenation

jagadish-kb:
Defined default circuit breaker configurations for all outbound http subsystems. The config is same for all downstream systems for now, however there is provision in the config to override any specific value for each subsystem/api within it.


lib/utils/HttpClient.js
```diff
@@ -18,39 +19,56 @@ class HttpClient {
...
+      const commonConfig = _.assign({},
+        _.omit(httpCircuitConfig, 'apis'),
+        _.omit(apiConfig, 'api_overrides'));
+      logger.silly(`Creating command '${baseUrl}_base_circuit' with options:${JSON.stringify(commonConfig)}`);
+      const commandKey = _.toLower(`${baseUrl}_base_circuit`);
+      this.commandMap[baseUrl] = {};
+      this.commandMap[baseUrl].BASE_CMD = this.createCommand(commandKey, commonConfig);
+      _.each(apiConfig.api_overrides, (apiOverrideConfig, httpMethod) => {
+        const apiConfig = _.assign({}, commonConfig, _.omit(apiOverrideConfig, 'method_overrides'));
+        const commandKey = _.toLower(`${baseUrl}_${httpMethod}_CIRCUIT`);
```
kamath-prasad: can we already use `circuit` instead of `CIRCUIT` since we do toLower ?

lib/utils/HttpClient.js
```diff
@@ -18,39 +19,56 @@ class HttpClient {
...
+        this.commandMap[baseUrl][commandKey] = this.createCommand(commandKey, apiConfig);
+        logger.silly(`Created command '${commandKey}' with options: ${JSON.stringify(apiConfig)}`);
+        _.each(apiOverrideConfig.method_overrides, (methodOverRideConfig, path) => {
+          const methodConfig = _.assign({}, apiConfig, methodOverRideConfig);
+          const commandKey = _.toLower(`${baseUrl}_${httpMethod}_${path}_CIRCUIT`);
```
kamath-prasad: same as above.

lib/utils/HttpClient.js
```diff
@@ -64,15 +82,34 @@ class HttpClient {
...
+      }
+      baseUrl = `${urlComponents.protocol}//${urlComponents.hostname}:${urlComponents.port}`;
+      path = urlComponents.pathname;
+      if (this.commandMap[baseUrl] === undefined) {
+        logger.silly('setting up command factory for url :-', baseUrl);
+        this.buildCommandFactory(baseUrl);
+      }
+    }
+    if (this.commandMap[baseUrl]) {
+      if (this.commandMap[baseUrl][`${baseUrl}_${httpMethod}_${path}_CIRCUIT`]) {
```
kamath-prasad: will checking index with captialized `CIRCUIT` work here and below since we create the command map using lower case indexes in buildCommandFactory ?

lib/logger.js
```diff
@@ -13,7 +13,7 @@ const transports = [
     prettyPrint: true,
     level: config.log_level || 'info',
     silent: false,
-    colorize: true,
+    colorize: (config.colorize_log !== undefined && config.colorize_log === false) ? false : true,
```
jagadish-kb: by default this wont be defined. So only in cases where it is explicity defined and set to false, this is to be turned off. Existing behavior will not be affected by this code, this is only required in local and hence changes will be done only in brokers settings.yml and no changes required in bosh release.

config/circuit-breaker-config.yml
```diff
@@ -0,0 +1,44 @@
...
+# If an API URL is not configured in the below, then all calls to that subsystem bypass the circuit breaker.
+# One circuit is created for each of the configured API URL (Base Circuit) & additional circuits are created if configuration are overridden.
+# Appropriate circuit is chosen from bottom-up i.e. if for a particular path in API if configuration is overridden, then corresponding circuit with overriden config is applied, else Base circuit is applied)
+http:
+  error_threshold_percentage: 49 #error percentange at which the circuit trips
+  service_timeout: 180000 #timeout time by which the request must be served
+  request_volume_threshold: 10 #Number of requests post which metrics checks are applied by circuit breaker
+  sleep_window_in_ms : 120000 #ms the circuit trips before it allows for a single request to go through in half-open circuit state
+  statistical_window_length: 120000 #ms the metrics are caclucated for, after this the metrics are reset.
+  statistical_window_number_of_buckets: 10 #statistical window length is sliced into this configured number of buckets
```
jagadish-kb: Those do not use HTTPClient, they internally use PackageCloud which internally has its own client. Defining circuit breaker for them would be a seperate excercise very similar to Riemann. In this checkin, only all those subsystems with which SF interacts directly via REST are addressed. Namely CF, BOSH, Quota Mgmt & SF itself.

config/circuit-breaker-config.yml
```diff
@@ -0,0 +1,44 @@
...
+    "https://uaa.bosh-lite.com":
+      name: "CF UAA"
+    "https://service-fabrik-broker.bosh-lite.com":
+      name: "Service Fabrik APIs"
+    "https://10.0.2.2:9293":
+      name: "Service Broker APIs"
+    "https://sap-provisioning.sb-iot-sap.cfapps.stagingaws.hanavlab.ondemand.com":
+      name: "Quota Management Oauth"
+    "https://tenant-onboarding-develop.cfapps.stagingaws.hanavlab.ondemand.com":
+      name: "Qutoa Management APIs"
```
jagadish-kb: Ok

lib/utils/HttpClient.js
jagadish-kb: ok.

lib/utils/HttpClient.js
jagadish-kb: ok

lib/utils/HttpClient.js
jagadish-kb: good catch.

lib/logger.js
kamath-prasad: Ok , got it.

config/circuit-breaker-config.yml
kamath-prasad: Oh , my bad forgot it :) .. then we are fine .

Issue #30, opened 2017-08-29
Come up with admin API to get the BOSH director config for a giving service's BOSH deployment


kamath-prasad:
**Problem Description**

- Currently after the support for Multi-BOSH - https://github.com/SAP/service-fabrik-broker/pull/7 the deployments for service instances could be created across any of the BOSH's configured . 
- Hence from a service's POV there is no single place to look for the BOSH where the deployment is present in order to verify the correctness of the deployment.

**Proposed Solution**

- Come up with a admin API in broker which takes the BOSH deployment name as input.
- Output should contain the details of the BOSH director under which the deployment is created.

shashankmjain:
Can you please check if Services should be allowed to call back the Service Fabrik  broker. I had already raised this concern via email. Can you check this. As far as I remember the security groups are configured in a way, so that services cant call back the Broker. Must have been for a purpose. May be now we decided to change that.

kamath-prasad:
@shashankmjain : This API would not be exposed to the services . Broker component will expose this functionality via scripts and services can only use it in their unit/integration tests with the broker.

PR #31, opened 2017-08-29
Closes-Bug: #30 Admin api for director config

muralisc:
Description:

With service fabrik now supporting multiple BOSH, there is a need to
find the director for the deployment by using the deployment name.
The code changes in this commit introduces a new admin endpoint

```
/admin/deployments/<deployment_name>/config
```

Which returns the config of the director when a valid deployment is
sent as parameter. Else returns error code of 400 and an empty json.

### DIFF
--- a/lib/routes/admin.js
+++ b/lib/routes/admin.js
@@ -52,6 +52,9 @@ router.route('/deployments/:name/backup')
 router.route('/deployments/:name/backup/status')
   .get(controller.handler('getLastOobBackupStatus'))
   .all(middleware.methodNotAllowed(['GET']));
+router.route('/deployments/:name/director')
+  .get(controller.handler('getDeploymentDirectorConfig'))
+  .all(middleware.methodNotAllowed(['GET']));
 router.route('/deployments/:name/restore')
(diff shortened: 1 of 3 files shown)


lib/routes/admin.js
```diff
@@ -52,6 +52,9 @@ router.route('/deployments/:name/backup')
 router.route('/deployments/:name/backup/status')
   .get(controller.handler('getLastOobBackupStatus'))
   .all(middleware.methodNotAllowed(['GET']));
+router.route('/deployments/:name/config')
```
kamath-prasad: @muralisc / @jagadish-kb  : should we have bosh in the url path or keep it generic as it is now ? Because tomorrow we might have some other configs as well , correct ?

muralisc: yeah makes sense . Now its not clear which config it is.
```
/deployments/<name>/directorconfig
```

jagadish-kb: how about this : `/deployments/<name>/director` obviously from the URL you cant get anything other than config, so no need to explicitly specify it.

PR #33, opened 2017-08-30
Fix #32 (Deployments in locked state even after successful backup)

### LINKED ISSUE #32: Locks on deployment not released after successful backups
On successful completion of backup, FabrikStatusPoller clears the poll & then tries to unlock the deployment via CC. However if there is an error while unlocking the deployment (either due to BOSH issue or CC issue), the deployment is left locked and poller no longer retries. This keeps the lock on the deployment for the max configured duration and no subsequent updates/backup/deletes can happen.

Fix:
1. Do not clear the poller till the lock is successfully released and retry unlock operation in case of failures in pollers next regular check.

jagadish-kb:
1. Poller is stopped (or cleared) only on successfully unlocking
deployment.
2. Restart of poller staggers the getLock calls on the deployment
to ensure we dont overload bosh with a huge numbe of getProperty
requests.

closes #32

### DIFF
--- a/lib/EventLogInterceptor.js
+++ b/lib/EventLogInterceptor.js
@@ -211,29 +211,34 @@ class EventLogInterceptor {
   }
 
   publishAndAuditLogEvent(url, method, request, response, checkResponseBody, instanceType) {
-    const eventConfig = _.clone(this.getEventConfig(url, method));
-    if (eventConfig === undefined) {
-      return undefined;
+    try {
+      const eventConfig = _.clone(this.getEventConfig(url, method));
+      if (eventConfig === undefined) {
+        return undefined;
+      }
+      if (checkResponseBody) {
+        eventConfig.check_res_body = true;
+      }
+      const opStatus = this.getOperationStatus(response, eventConfig, response.body);
+      instanceType = instanceType === undefined ? CONST.INSTANCE_TYPE.DIRECTOR : instanceType;
+      instanceType = instanceType.length > 0 ? `${instanceType}.` : instanceType;
+      const eventInfo = this.getEventInfo(instanceType,
+        eventConfig.event_name,
+        eventConfig.tags,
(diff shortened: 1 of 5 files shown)


lib/fabrik/FabrikStatusPoller.js
```diff
@@ -124,17 +135,38 @@ class FabrikStatusPoller {
...
+                  .delay(config.backup.lock_check_delay_on_restart * numberOfSFDeployment)
+                  .then(() => bosh
+                    .director
+                    .getLockProperty(deploymentName)
+                    .then(lockInfo => this.startIfNotLocked(lockInfo, operation))
+                    .catch(err => logger.error(`Error occurred while setting poller `, err)));
+              }
+            }))
+            .catch(errors.Timeout, (err) => {
+              logger.info('Error occurred while fetching deployments ...', err);
```
kamath-prasad: This should be logger.error ?

lib/EventLogInterceptor.js
```diff
@@ -211,29 +211,34 @@ class EventLogInterceptor {
...
-    }
-    if (checkResponseBody) {
-      eventConfig.check_res_body = true;
+    try {
+      const eventConfig = _.clone(this.getEventConfig(url, method));
+      if (eventConfig === undefined) {
+        return undefined;
+      }
+      if (checkResponseBody) {
+        eventConfig.check_res_body = true;
```
subhankarc: Should we use underscore here ?

jagadish-kb: We use camelCase for JS variable names and snake case for config values. [Ideally config values are loaded from YML]

subhankarc: 👍

kamath-prasad:
Looks good. Will merge after the bosh release changes are also reviewed.

PR #35, opened 2017-08-31
Closes #34 - Circuit breaker not configured for Bosh calls

### LINKED ISSUE #34: Circuit breaker not configured for BOSH calls from service-fabrik-broker
Circuit breaker config for BOSH are not enforced. 
PFB excerpts from the broker logs:
```
2017-08-22T06:34:16.190Z - warn: Circuit breaker not defined for : undefined_/deployments/service-fabrik-0266-399479fc-b517-4cbb-ab40-7b3e59aa284b/diff_POST
2017-08-22T06:34:16.191Z - warn: Circuit breaker not defined for : undefined_/deployments/service-fabrik-0477-236f25fb-5bfe-4331-a3e5-655e96a97fed/diff_POST
```

jagadish-kb:
Base URL in BoshDirector client was set for each of the requests which was not handled in HTTPClient's circuit breaker configuration/lookup. This is now fixed.

### DIFF
--- a/config/settings.yml
+++ b/config/settings.yml
@@ -493,6 +493,15 @@ development: &development
   <<: *defaults
   log_level: debug
   password: 'secret'
+  directors:
+  - <<: *director
+    cpi: aws_cpi
+    primary: true
+    support_create: true
+    name: bosh
+    uuid: b7e651e2-3afe-471f-9bca-93048fa591f4
+    infrastructure: &test_infrastructure
+      <<: *director_infrastructure
   agents:
     <<: *agents
     blueprint: &agent_blueprint1
@@ -679,7 +688,7 @@ production:
 test:
   <<: *defaults
   log_level: debug
-  enable_circuit_breaker: false
+  # enable_circuit_breaker: false
   external:
     <<: *external
     protocol: 'http'
--- a/lib/utils/HttpClient.js
+++ b/lib/utils/HttpClient.js
@@ -34,10 +34,10 @@ class HttpClient {
 
   buildCommandFactory(baseUrl) {
     const httpCircuitConfig = _.get(config, `circuit_breaker.http`);
-    if (httpCircuitConfig) {
+    if (baseUrl && httpCircuitConfig) {
       const apiConfig = httpCircuitConfig.apis[baseUrl];
       if (apiConfig === undefined) {
-        logger.warn(` Circuit breaker not defined for URL : ${baseUrl}`);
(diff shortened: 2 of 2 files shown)


config/settings.yml
```diff
@@ -493,6 +493,15 @@ development: &development
   <<: *defaults
   log_level: debug
   password: 'secret'
+  directors:
```
subhankarc: Can you please check if boshlite2 setting(https://github.com/SAP/service-fabrik-broker/blob/master/config/settings.yml#L533) is affected for this change or needs changes?

lib/utils/HttpClient.js
```diff
@@ -162,13 +163,13 @@ class HttpClient {
...
   request(options, expectedStatusCode) {
-    const command = this.getCommand(options.url, options.method);
+    const command = this.getCommand(options, options.method);
     if (command) {
       logger.silly(`command config for ${options.url}:`, _.omit(command, 'Promise'));
       return Promise.try(() => command.execute(options, expectedStatusCode));
     }
     if (this.PARSED_URL_MAP[`${options.url}_${options.method}`] === undefined) {
-      logger.warn(`Circuit breaker not defined for : ${this.baseUrl}_${options.url}_${options.method}`);
+      logger.warn(`Circuit breaker not defined for : ${options.url} HTTP Method:${options.method}`);
```
subhankarc: Shouldn't it be debug similar to https://github.com/SAP/service-fabrik-broker/pull/35/files#diff-b0429a6c736f8cb3d4e258a10501fb13R40

jagadish-kb: No. I explictly kept this at warn as this helps in identifying URLs that are not configured for Circuit breaker easily. The line pointed by you gets invoked repeatedly every time so that was getting logged every time. However this line gets only logged only once. Post that `this.PARSED_URL_MAP` is updated with dummy object and this never gets logged. So having just one log message is OK.

config/settings.yml
jagadish-kb: `director` is no longer a valid property. It must be `directors`. Since i was testing in development, i only fixed that config. Please note this should have been updated as part of the BOSH user story. I have anyways updated the same for other envs as well.

lib/utils/HttpClient.js
subhankarc: Understood. That should be fine.

config/settings.yml
subhankarc: Thanks @jagadish-kb

PR #37, opened 2017-09-04
Closes #36 - SF MongoDB bind was attempted even during udpate

### LINKED ISSUE #36: ServiceFabrik tries to create binding with SF Mongodb even on update
On update of SF mongodb, binding is being created again. Ideally we must bind to the SF mongodb only once during the initial creation and post that we must always retrieve the binding and use the same and not try to create it every time even on each update.

jagadish-kb:
DBManager initiates connection to DB by getting the DB co-orindates
from the property stored on deployment. Once db bind info is retrieved
it is then stored locally & on any future ocassions which requires
db connections to be reset again, the same bind info is used.

However during start if BOSH is down, then db bind info is not
fetched & during an update request after it completes successfully
if bind info is not found in cache its assumed that its a creation
flow and tries creating the bind once again which errors as reported.

To fix this couple of things have been done:
1. During start if unexpected errors occur then initialize of DB
manager is retried till it succeeds (with a delay)
2. Post create/update operation before creating the binding first
a check is done to see if binding already exists and only then
creation of binding is done.
3. Retry is introduced even for erros occuring during polling of
create/update task status.
4. If bind info cannot be retrieved post update/create bosh errors
even in such scenarios retry happens till it succeeds.

Hopefully with all the retries in place it must certainly fix the
DBManager failures due to BOSH errors.

Fixes #36


lib/fabrik/DBManager.js
```diff
@@ -42,86 +41,128 @@ const DB_STATES = {
...
-              logger.error(`mongodb property - config.mongodb.deployment_name is undefined. Deployment name must be defined.`);
-              this.dbState = DB_STATES.NOT_CONFIGURED;
-            }
+      return utils
+        .retry(() => this
+          .directorManager
+          .getBindingProperty(config.mongodb.deployment_name, CONST.FABRIK_INTERNAL_MONGO_DB.BINDING_ID), {
+            maxAttempts: 5,
+            minDelay: 5000,
+            predicate: (err) => !(err instanceof ServiceBindingNotFound)
```
kamath-prasad: Will we be able to log this error ? Or will it be logged by DirectorManager ?

lib/fabrik/DBManager.js
```diff
@@ -42,86 +41,128 @@ const DB_STATES = {
...
-      .startUp(config)
-      .catch(() => this.dbState = DB_STATES.CONNECTION_FAILED);
+    return utils
+      .retry(() => dbConnectionManager
+        .startUp(config), {
+          maxAttempts: 5,
+          minDelay: 5000
+        })
+      .then(() => this.dbInitialized = true)
+      .catch((err) => {
```
kamath-prasad: Do we need to log this error ?

lib/fabrik/DBManager.js
```diff
@@ -42,86 +41,128 @@ const DB_STATES = {
...
+
+  initialize() {
+    return Promise.try(() => {
+        if (_.get(config, 'mongodb.provision.plan_id') === undefined && _.get(config, 'mongodb.url') === undefined) {
+          logger.warn('Mongodb not configured. Either DB URL or Mongo Plan Id must be configured for enabling MongoDB usage with ServiceFabrik.', _.get(config, 'mongodb.url'));
+          this.dbState = DB_STATES.NOT_CONFIGURED;
+          return;
+        }
+        if (_.get(config, 'mongodb.provision.plan_id') !== undefined && _.get(config, 'mongodb.provision.network_index') === undefined) {
+          logger.warn('Plan Id is defined. Must also define network segment index where mongodb is to be deployed, else DB create/update will fail');
```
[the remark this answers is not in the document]
jagadish-kb: Index Id is required during create/update operation and not during initialisation. In createUpdate flow if this config is not present exception is thrown, however during initialization for whatsoever reason if this config is missed and DB connect can happen then this will just log the warning and proceed.

lib/fabrik/DBManager.js
jagadish-kb: This error gets logged in initialize method [here](https://github.com/jagadish-kb/service-fabrik-broker/blob/7aa3fda8089470c5e7f413fd66665fe097802b49/lib/fabrik/DBManager.js#L86)

lib/fabrik/DBManager.js
jagadish-kb: Error bubbles up and gets logged at one place centrally [here](https://github.com/jagadish-kb/service-fabrik-broker/blob/7aa3fda8089470c5e7f413fd66665fe097802b49/lib/fabrik/DBManager.js#L88)

lib/bosh/BoshDirectorClient.js
```diff
@@ -374,7 +373,9 @@ class BoshDirectorClient extends HttpClient {
 
   }
 
-  pollTaskStatusTillComplete(taskId, pollInterval, timeout) {
+  pollTaskStatusTillComplete(taskId, pollInterval, timeout, maxErrorRetry) {
+    let errorRetries = 0;
+    maxErrorRetry = maxErrorRetry || 3;
```
sauravmndl: The default value 3 is hard coded, is it better to to put in constant so that somewhere can be used also .. or configurable.
so that in similar cases can be used!

test/acceptance/service-fabrik-admin.internal-db.lifecycle.spec.js
```diff
@@ -28,6 +28,7 @@ describe('service-fabrik-admin', function () {
 
     before(function () {
       config.mongodb.provision.plan_id = 'bc158c9a-7934-401e-94ab-057082a5073f';
+      mocks.director.getBindingProperty(CONST.FABRIK_INTERNAL_MONGO_DB.BINDING_ID, {}, config.mongodb.deployment_name, 'NOTFOUND');
```
sauravmndl: 'NOTFOUND' --> can we define a constant? or fine ??

test/acceptance/service-fabrik-admin.internal-db.lifecycle.spec.js
```diff
@@ -63,6 +64,7 @@ describe('service-fabrik-admin', function () {
       it('should provision service fabrik internal mongodb when deployment not found', function (done) {
         const WAIT_TIME_FOR_ASYNCH_CREATE_DEPLOYMENT_OPERATION = 200;
         this.timeout(2000 + WAIT_TIME_FOR_ASYNCH_CREATE_DEPLOYMENT_OPERATION);
+        mocks.director.getBindingProperty(CONST.FABRIK_INTERNAL_MONGO_DB.BINDING_ID, {}, config.mongodb.deployment_name, 'NOTFOUND');
```
sauravmndl: 'NOTFOUND' -->same as above.

lib/bosh/BoshDirectorClient.js
```diff
@@ -405,9 +406,12 @@ class BoshDirectorClient extends HttpClient {
             }
           })
           .catch(err => {
+            errorRetries++;
             logger.error(`error occurred while fetching state of task id: ${taskId}`, err);
```
sauravmndl: May be we can include errorRetries (the count in present loop) in log..... so in a huge log file it will be easy to debug.

sauravmndl:
Overall looks good, only few findings mentioned above!

test/acceptance/service-fabrik-admin.internal-db.lifecycle.spec.js
jagadish-kb: Thats the name of the error class, no need to define it in constants.

test/acceptance/service-fabrik-admin.internal-db.lifecycle.spec.js
jagadish-kb: Thats the name of the error class, no need to define it in constants.

lib/fabrik/DBManager.js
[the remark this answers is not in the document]
kamath-prasad: Ok , got it.

lib/fabrik/DBManager.js
kamath-prasad: ok.

lib/fabrik/DBManager.js
kamath-prasad: ok.

kamath-prasad:
@jagadish-kb : Will merge this PR . Since it is on master it will not be part of 9A . Hence request you to raise a PR on 9A rel branch too , else let this fix go as part of 9B ?!

Issue #22, opened 2017-08-21
Support for backup and restore on Azure


sauravmndl:
Service Fabrik Broker should have the capability for backup and restore on Azure.
It should support ondemand/schedule backup. All operations related to backup and restore
should be also supported on Azure.

PR #40, opened 2017-09-06
closes issue #22 Feature Azure Backup and Restore (#9)

sauravmndl:
Description:

1. Added new class lib/iaas/AzureClient.js
2. Refactored lib/iaas/CloudProviderClient.js
3. Introduced parent class lib/iaas/BaseCloudClient.js and made above
   two as sub clas
4. Updated package.json for new node modules
5. Added Unit Tests
6. Added dummy config for Azure in config/settings.yml
7. Renamed BlockStorageClient.js --> ComputeClient.js and added section
   for Azure


lib/iaas/AzureClient.js
```diff
@@ -0,0 +1,192 @@
...
+    );
+    this.computeClient = this.constructor.createComputeClient(_
+      .chain(this.settings)
+      .set('provider', this.provider)
+      .value()
+    );
+    Promise.promisifyAll(this.computeClient.snapshots);
+
+    this.storage.on('log::*', function onmessage(message, obj) {
+      const event = _.nth(this.event.split('::'), 1);
```
sauravmndl: Need to verify if log event is supported

lib/iaas/AzureClient.js
```diff
@@ -0,0 +1,192 @@
...
+
+  static createStorageClient(options) {
+    return Promise.promisifyAll(
+      azureStorage.createBlobService(options.storageAccount, options.storageAccessKey));
+  }
+
+  static createComputeClient(settings) {
+    return ComputeClient.createComputeClient(settings).value();
+  }
+
```
sauravmndl: extra new line not required

lib/iaas/AzureClient.js
```diff
@@ -0,0 +1,192 @@
...
+}
+
```
sauravmndl: extra new line

lib/iaas/BaseCloudClient.js
```diff
@@ -0,0 +1,89 @@
...
+  }
+
+  static createComputeClient() {
+    throw new NotImplementedBySubclass('createComputeClient');
+  }
+
+  static get providerErrorTypes() {
+    return CloudProviderError;
+  }
+
```
sauravmndl: extra new line

lib/iaas/ComputeClient.js
```diff
@@ -0,0 +1,48 @@
...
+
+  constructor(settings) {
+    this.settings = settings;
+  }
+
+  deleteSnapshot(settings) {
+    throw errors.NotImplemented(`ComputeClient is not supported for ${settings.provider}`);
+  }
+}
+
```
sauravmndl: extra new line

lib/iaas/AzureClient.js
```diff
@@ -0,0 +1,192 @@
...
+  list(container, options) {
+    if (arguments.length < 2) {
+      options = container;
+      container = this.containerName;
+    }
+    const prefix = options ? options.prefix : null;
+    //TODO Need to update following line with value from 'options' e.g. options.marker
+    let continuationToken = null;
+    //TODO : maxResult is 5000 
+    //https://azure.github.io/azure-storage-node/BlobService.html#listBlobsSegmentedWithPrefix__anchor
```
sauravmndl: Finalise the approach on pagination.

lib/iaas/AzureClient.js
```diff
@@ -0,0 +1,192 @@
...
+      .catch(BaseCloudClient.providerErrorTypes.Unauthorized, err => {
+        logger.error(err.message);
+        throw new Unauthorized('Authorization at the cloud storage provider failed');
+      })
+      .catch(BaseCloudClient.providerErrorTypes.Forbidden, err => {
+        logger.error(err.message);
+        throw new Forbidden('Authentication at the cloud storage provider failed');
+      })
+      .catchThrow(BaseCloudClient.providerErrorTypes.NotFound, new NotFound(`Object '${file}' not found`));
+  }
```
sauravmndl: the error message can be better.

lib/iaas/AzureClient.js
```diff
@@ -0,0 +1,192 @@
...
+      container = this.containerName;
+    }
+    logger.debug(`Deleting file ${file} in container ${container} `);
+    return this.storage
+      .deleteBlobAsync(container, file)
+      .then(response => logger.info(`Response of blob deletion from cloud storage for ${file} :`, response))
+      .catch(BaseCloudClient.providerErrorTypes.Unauthorized, err => {
+        logger.error(err.message);
+        throw new Unauthorized('Authorization at the cloud storage provider failed');
+      })
```
sauravmndl: Keep a reference of file name and may be also container name in error message, otherwise difficult to track in huge log file.

lib/iaas/AzureClient.js
```diff
@@ -0,0 +1,192 @@
...
+      .catch(BaseCloudClient.providerErrorTypes.Forbidden, err => {
+        logger.error(err.message);
+        throw new Forbidden('Authentication at the cloud storage provider failed');
+      })
```
sauravmndl: keep reference of blob name and container

lib/iaas/AzureClient.js
jagadish-kb: rename this variable to eventLogLevel

jagadish-kb: what is this.event? shouldn't you be splitting the message instead?

lib/iaas/AzureClient.js
```diff
@@ -0,0 +1,192 @@
...
+      ], event) ? event : 'debug';
+      if (obj) {
+        logger.log(level, message, obj);
+      } else {
+        logger.log(level, message);
+      }
+    });
+  }
+
+  getContainer(container) {
```
jagadish-kb: rename variable from `container` to `containerName`

lib/iaas/AzureClient.js
```diff
@@ -0,0 +1,192 @@
...
+      container = this.containerName;
+    }
+    const prefix = options ? options.prefix : null;
+    //TODO Need to update following line with value from 'options' e.g. options.marker
+    let continuationToken = null;
+    //TODO : maxResult is 5000 
+    //https://azure.github.io/azure-storage-node/BlobService.html#listBlobsSegmentedWithPrefix__anchor
+    //If want to fetch more use 'continuationToken' returned in result
+    return this.storage  
+      .listBlobsSegmentedWithPrefixAsync(container, prefix, continuationToken)
```
jagadish-kb: An implementation similar to `ResourceStream` which internally iterates over all results and promisify the method must be done to handle pagination instead of doing it like this Or look at implemenation of `BackupStore.listFilenames` which invokes itself recursively till all the pages are read.

lib/iaas/AzureClient.js
```diff
@@ -0,0 +1,192 @@
...
+      file = container;
+      container = this.containerName;
+    }
+    logger.debug(`Deleting file ${file} in container ${container} `);
+    return this.storage
+      .deleteBlobAsync(container, file)
+      .then(response => logger.info(`Response of blob deletion from cloud storage for ${file} :`, response))
+      .catch(BaseCloudClient.providerErrorTypes.Unauthorized, err => {
+        logger.error(err.message);
+        throw new Unauthorized('Authorization at the cloud storage provider failed');
```
jagadish-kb: Better to send `err.message` to this constructor instead of this hard coded message. There is anyways a hardcoded message in errors.js for this class in case its input as undefined.

lib/iaas/AzureClient.js
```diff
@@ -0,0 +1,192 @@
...
+      })
+      .catch(BaseCloudClient.providerErrorTypes.Forbidden, err => {
+        logger.error(err.message);
+        throw new Forbidden('Authentication at the cloud storage provider failed');
```
jagadish-kb: same as [above](https://github.com/SAP/service-fabrik-broker/pull/40/files#r137982476)

lib/iaas/AzureClient.js
```diff
@@ -0,0 +1,192 @@
...
+        .createWriteStreamToBlockBlob(
+          options.container, options.remote);
+
+      uploadedStream.on('error', onerror);
+      uploadedStream.on('finish', onfinish);
+      uploadedStream.end(buffer);
+    });
+  }
+
+  uploadJson(container, file, data) {
```
jagadish-kb: What is the difference between upload vs uploadJson? are both of them needed?

lib/iaas/AzureClient.js
```diff
@@ -0,0 +1,192 @@
...
+    return this
+      .upload({
+        container: this.containerName,
+        remote: file
+      }, new Buffer(JSON.stringify(data, null, 2), 'utf8'));
+  }
+
+  downloadJson(container, file) {
+    if (arguments.length < 2) {
+      file = container;
```
jagadish-kb: everywhere replace container with containerName to be consistent.

lib/iaas/AzureClient.js
```diff
@@ -0,0 +1,192 @@
...
+      container = this.containerName;
+    }
+    return this
+      .download({
+        container: this.containerName,
```
jagadish-kb: this is incorrect. Here instead of `this.containerName` it should be just `containerName` as you can specify in method if you want to read from a diff container

lib/iaas/AzureClient.js
```diff
@@ -0,0 +1,192 @@
...
+        remote: file
+      })
+      .then((data) => {
+        const response = JSON.parse(data);
+        response.trigger = response.trigger === CONST.BACKUP.TRIGGER.MANUAL ? CONST.BACKUP.TRIGGER.SCHEDULED : response.trigger;
+        //The above conversion is done to handle existing CRON Jobs which set this trigger as 'manual' even for scheduled Jobs
```
jagadish-kb: Remove these comments as its not applicable in case of AZURE and also these checks are no longer required. This was only required in case of earier implementation where you had scheduled backups being set as manual in meta info.

lib/iaas/AzureClient.js
```diff
@@ -0,0 +1,192 @@
...
+      })
+      .catchThrow(SyntaxError, new NotFound(`Object '${file}' not found`));
+  }
+
+  deleteSnapshot(snapshotName) {
+    return Promise
+      .try(() => {
+        return this.computeClient
+          .snapshots
+          .deleteMethodAsync(this.settings.resource_group, snapshotName)
```
jagadish-kb: Basic validation of this.settings in constructor and logging warning/errors will help in identifying the issues early on in case of some misses in setting right props.

lib/iaas/AzureClient.js
```diff
@@ -0,0 +1,192 @@
...
+          .then(retval => logger.info(`Deleted snapshot ${snapshotName}`, retval))
```
jagadish-kb: what is retVal ? is it a boolean? In that case log this in a tap as the response from deleteSnapshot will not be a boolean but undefined because of above logger. In either case better to use tap if you just want to log response.

lib/iaas/index.js
```diff
@@ -2,15 +2,20 @@
...
 const config = require('../config');
 const CloudProviderClient = require('./CloudProviderClient');
+const AzureClient = require('./AzureClient');
 const BackupStore = require('./BackupStore');
 const BackupStoreForServiceInstance = require('./BackupStoreForServiceInstance');
 const BackupStoreForOob = require('./BackupStoreForOob');
+const BaseCloudClient = require('./BaseCloudClient');
 
-const cloudProvider = new CloudProviderClient(config.backup.provider);
+const cloudProvider = config.backup.provider.name === 'azure' ?
```
jagadish-kb: Better to move this conditional creation into `BaseCloudClient` new factory method which provides right implementation based on input provider name.

test/iaas.AzureClient.spec.js
```diff
@@ -0,0 +1,366 @@
...
+    const scheduled_data = {
+      trigger: CONST.BACKUP.TRIGGER.SCHEDULED,
+      type: 'online',
+      state: 'succeeded',
+      backup_guid: backup_guid,
+      started_at: filename.isoDate(moment().subtract(1, 'days').toISOString()),
+      agent_ip: mocks.agent.ip,
+      service_id: service_id
+    };
+
```
jagadish-kb: unnecessary new lines

test/iaas.AzureClient.spec.js
```diff
@@ -0,0 +1,366 @@
...
+                <Metadata>  
+                  <Color>green</Color>  
+                  <BlobNumber>02</BlobNumber>  
+                </Metadata>  
+              </Blob>
+            </Blobs>  
+            <NextMarker />   
+          </EnumerationResults>`
+        };
+        mocks.azureClient.list(settings.container, undefined, expectedResponse);
```
jagadish-kb: Is xml the expected response type? most APIs now have support for JSON can you check if you can tweak any input configs for it to directly return JSON instead of XML. [XML parsing is slower than compared to JSON parsing]

test/iaas.AzureClient.spec.js
```diff
@@ -0,0 +1,366 @@
...
+          headers: {
+            'transfer-encoding': 'chunked',
+            'content-md5': '+v4bYMJBB8zY9FYiE+RISQ==',
+            'last-modified': 'Thu, 05 Jan 2017 03:13:04 GMT',
+            etag: '"0x8D43518C381D87C"',
+            server: 'Windows-Azure-Blob/1.0 Microsoft-HTTPAPI/2.0',
+            'x-ms-request-id': '5dc67ddb-0001-003d-7701-672321000000',
+            'x-ms-version': '2016-05-31',
+            'x-ms-request-server-encrypted': 'false',
+            date: new Date().toISOString(),
```
jagadish-kb: Do you have to mock it with all these headers explicitly? can you not just mock with URL? Reason being if we upgrade the azure library and if the lib starts sending in new headers then these tests would also fail we might have to fix for errors which are non-functional. So please verify if these mocks can be generic. 
Also in the above i see date is being set will this match the date sent at request time? for matching dates always use sinon faketimers in tests.

test/iaas.AzureClient.spec.js
```diff
@@ -0,0 +1,366 @@
...
+          body: scheduled_data
+        }, 2);
+        return client.uploadJson(blobName, scheduled_data)
+          .then(result => {
+            expect(result).to.eql(scheduled_data);
+            mocks.verify();
+          });
+      });
+
+      it('blob upload should fail: PreconditionFailed', function () {
```
jagadish-kb: from the TC its not clear what precondition is being met. Can you please add details either in the title or as comments?

test/iaas.AzureClient.spec.js
sauravmndl: Azure supports XML as of now. See https://docs.microsoft.com/en-us/rest/api/storageservices/list-blobs. I didn't find json example response anywhere.

test/iaas.AzureClient.spec.js
```diff
@@ -0,0 +1,366 @@
...
+          .catch(err => {
+            expect(err.statusCode || err.code).to.eql(412);
+            mocks.verify();
+          });
+
+      });
+
+      it('blob download should be successful', function () {
+        mocks.azureClient.download(`/${settings.container}/${blobName}`, scheduled_data, {
+          'content-type': 'application/octet-stream',
```
jagadish-kb: Are all these response headers required? Most of the param values wont have right values. So if not required, please remove them.

test/iaas.AzureClient.spec.js
```diff
@@ -0,0 +1,366 @@
...
+        return client.downloadJson(`${blobName}`)
+          .catch(err => {
+            expect(err).to.be.instanceof(NotFound);
+            mocks.verify();
+          });
+      });
+
+      it('blob download should fail on non json body: NotFound error', function () {
+        mocks.azureClient.download(`/${settings.container}/${blobName}`, 'Hello World', {
+          'content-type': 'application/octet-stream',
```
jagadish-kb: same as above

test/mocks/azureClient.js
```diff
@@ -0,0 +1,140 @@
...
...n\":\"1462557169\",\"not_before\":\"1462553269\",\"resource\":\"https://management.core.windows.net/\",\"access_token\":\"eyJ0eXAiOiJKV1QiLCJhbGciOiJSUzI1NiIsIng1dCI6Ik1uQ19WWmNBVGZNNXBPWWlKSE1iYTlnb0VLWSIsImtpZCI6Ik1uQ19WWmNBVGZNNXBPWWlKSE1iYTlnb0VLWSJ9.eyJhdWQiOiJodHRwczovL21hbmFnZW1lbnQuY29yZS53aW5kb3dzLm5ldC8iLCJpc3MiOiJodHRwczovL3N0cy53aW5kb3dzLm5ldC83MmY5ODhiZi04NmYxLTQxYWYtOTFhYi0yZDdjZDAxMWRiNDcvIiwiaWF0IjoxNDYyNTUzMjY5LCJuYmYiOjE0NjI1NTMyNjksImV4cCI6MTQ2MjU1NzE2OSwiYXBwaWQiOiJiOWU2ZTA3Yi1jNDNlLTQ3MzEtODVjYS05ODE3ODkyNzI0Y2QiLCJhcHBpZGFjciI6IjEiLCJpZHAiOiJodHRwczovL3N0cy53aW5kb3dzLm5ldC83MmY5ODhiZi04NmYxLTQxYWYtOTFhYi0yZDdjZDAxMWRiNDcvIiwib2lkIjoiNGUwNDNmODYtYjMzZC00YzNiLThjNTYtNWM3NTkyOGEzNzBlIiwic3ViIjoiNGUwNDNmODYtYjMzZC00YzNiLThjNTYtNWM3NTkyOGEzNzBlIiwidGlkIjoiNzJmOTg4YmYtODZmMS00MWFmLTkxYWItMmQ3Y2QwMTFkYjQ3IiwidmVyIjoiMS4wIn0.rROtJoq7sd0BIfLtS-Ra9-xIN9lQPXnN0NvR8BWgEP8imEp1M3ryN8v1IUldiQeb6yhsf0Jg1QqA_vI8HUvCpDoCA7MvmSrMCVfrY7hKFDl9cJeFMILVXjWRbCHp29k0A_pz1r8au07_MmGGrEpBpc0Z5P1rr9qwOe4SmLCuC1poyNuERpwMtHXS4MwJ7mjz9OHsn_pvVxV9TMA-lcBxqfsxWPf8kbmHiFeyLonBftD-h0X7wjwJ-EGK6WSQ-bBtWyJNyJ6sDxqSUB4WpuywVVe_pCPcnKLhqD5tbW_thkyQG7nJ4bM_qqXvKIEtgxO0HGDIA7KS5AdMgBx8uNAC7g\"}', {
```
jagadish-kb: are the mock response params expires_on, not_before used? if not, then remove them. If they are used then certainly hardcoding them might make tests fail later.

test/mocks/azureClient.js
```diff
@@ -0,0 +1,140 @@
...
+      'content-length': '1234'
+    });
+}
+
+function download(remote, body, headers) {
+  let status = 200;
+  if (body instanceof Error && body.status) {
+    const err = _.pick(body, 'status', 'message');
+    status = err.status;
+    headers['content-type'] = 'text/html';
```
jagadish-kb: I thought we request for content-type of xml, is response coming back in html?

lib/iaas/AzureClient.js
sauravmndl: done

lib/iaas/AzureClient.js
sauravmndl: done

lib/iaas/BaseCloudClient.js
sauravmndl: done

lib/iaas/AzureClient.js
sauravmndl: If you feel it is not a must required.... then can we keep the name intact. In cloudproviderclient class also we are using same name.

lib/iaas/AzureClient.js
sauravmndl: should we fetch all in this this method itself ? ..... In case of BackupStore.listFilenames(), the caller (BackupStore) is doing the accumulation. CloudProviderClient also returning batch-by-batch. Not whole. If we do accumulation here, then it behaves differently than CoudProviderClient. Problem is in BackupStore.listFilenames() we have done package cloud specific code. Now somehow we should harmonise that, in order to support azure library structure. According to me this change is not must as part of initial release as Azure supports by default 5000 and very unlikely we'll be having 5000 backups in a space with backup schedule option once per day (instead of 3 per day which is case in AWS). But I agree it is better to have now only. We can discuss face to face  once and take decision.

lib/iaas/AzureClient.js
sauravmndl: err.message only gives 'Unauthorized'. I'll put a reference of blob name and container in the statement.

sauravmndl: updated the message

lib/iaas/AzureClient.js
sauravmndl: done

lib/iaas/ComputeClient.js
sauravmndl: done

lib/iaas/AzureClient.js
sauravmndl: Ideally  not required. uploadJson(): converts json to buffer and upload() uses only buffer. Is okay to have as someone may need to use upload() directly to feed a buffer when data not in json format.

lib/iaas/AzureClient.js
sauravmndl: done

lib/iaas/AzureClient.js
sauravmndl: done

lib/iaas/AzureClient.js
sauravmndl: done

lib/iaas/AzureClient.js
sauravmndl: done

lib/iaas/AzureClient.js
sauravmndl: As discussed .... Don't want to return response to caller. Will keep as it is.

test/iaas.AzureClient.spec.js
sauravmndl: done

test/mocks/azureClient.js
sauravmndl: done changes

test/iaas.AzureClient.spec.js
sauravmndl: These are the response headers from Azure and independent of library.

test/iaas.AzureClient.spec.js
sauravmndl: done

test/iaas.AzureClient.spec.js
sauravmndl: removed many headers.

test/iaas.AzureClient.spec.js
sauravmndl: removed many headers.

test/mocks/azureClient.js
sauravmndl: done

lib/iaas/AzureClient.js
sauravmndl: done

test/iaas.AzureClient.spec.js
sauravmndl: Though removed many headers.

lib/iaas/index.js
sauravmndl: done

lib/iaas/AzureClient.js
sauravmndl: done

lib/iaas/AzureClient.js
sauravmndl: this should be based on event only. If you see next statement... it is looking for log level. But the problem is I inherited the code from CloudProviderClient.js which uses package cloud. But I am using azure storage lib which also has event listener. But I didn't see any example how to use this. Planning to remove this listener.

lib/iaas/AzureClient.js
sauravmndl: if the resource group is not present we should return error. Shall i return the error? in that case broker monit process will not start which is required also in my opinion. But thing is all other config params are check as at the beginning the connection clients are created using them. Only this resource group is pending :P

sauravmndl: Done check for resource_group

lib/iaas/AzureClient.js
sauravmndl: @subhankarc  Yes, for this we need to change in BackupStore.listFilenames(). But want to do that diligently as we have to test with > 5000 blobs and need to incorporate unit test for that as well. As we are going staging this sprint, my suggestion is to keep a track of this and will do it properly. Also, as I mentioned Azure supports 5000 instead of 1000 which is case for AWS and with frequency once per day instead of thrice which is case for AWS. Please let me know if this is fine.

lib/iaas/AzureClient.js
```diff
@@ -0,0 +1,183 @@
...
+  remove(container, file) {
+    if (arguments.length < 2) {
+      file = container;
+      container = this.containerName;
+    }
+    logger.debug(`Deleting file ${file} in container ${container} `);
+    return this.storage
+      .deleteBlobAsync(container, file)
+      .then(response => logger.info(`Response of blob deletion from cloud storage for ${file} in container ${container}:`, response))
+      .catch(BaseCloudClient.providerErrorTypes.Unauthorized, err => {
```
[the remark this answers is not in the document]
sauravmndl: Unauthorized, Forbidden, NotFound are explicitly handled for blob deletion and other errors are automatically handled in upper layer as most of the cases in our code base.

lib/iaas/AzureClient.js
sauravmndl: As mentioned in one of previous comments, keeping the name 'container' for both the subclass of BaseCloudClient : AzureClient and CloudProviderClient.

lib/iaas/AzureClient.js
jagadish-kb: Created a ticket #46

lib/iaas/AzureClient.js
jagadish-kb: ok

lib/iaas/AzureClient.js
jagadish-kb: ok

lib/iaas/AzureClient.js
jagadish-kb: ok

test/iaas.AzureClient.spec.js
jagadish-kb: ok

PR #44, opened 2017-09-08
Fixes #43 Retrying unlock of deployment leads to a domino effect on BOSH

### LINKED ISSUE #43: Fabrikstatus poller retries unlocking of deployments which runs beyond polltime
Fabrikstatus poller retries unlocking in case of an error in an exponential backoff manner and by the time the second retry is to be done, the next run of fabrik status poller comes over and it fails & even it retries. This leads to a domino effect of unlocking across retries and next runs of fabrik status poller.

Fix: Do not try retry of unlock operation as FabrikStatus poller is cancelled only after successful unlock, so if it fails, the next run anyways tries it again.

Additionally, also cache the `/getDeployments` calls to bosh as its called for every broker update operation and not caching this leads to too many calls to bosh for the same & this operation is quite expensive. [While caching ensure cache misses leads to once again querying bosh for deployments]

jagadish-kb:
On unlock error, retry is implemented. However if BOSH is already down
performing unlock operation results in a CF UPDATE which internally
invokes /getDeployments on BOSH which is an expensive call. The fix is:
1. Stop retry of unlock operation in individual run of poller
2. Find deploymentName for the instance Id initially in the BOSH cache
and then if there is a cache miss goto BOSH to getDeployments
3. FabrikStatusPoller also gets all the deploymentNames from the BOSH Cache

Closes #43

### DIFF
--- a/lib/bosh/BoshDirectorClient.js
+++ b/lib/bosh/BoshDirectorClient.js
@@ -33,8 +33,13 @@ class BoshDirectorClient extends HttpClient {
     this.populateCache();
   }
 
-  clearCache() {
-    this.cache = {};
+  clearCache(config) {
+    if (config) {
+      logger.info('clearing cache for config - ', config.name);
+      _.each(this.cache, (value, key) => value === config ? delete this.cache[key] : '');
+    } else {
+      this.cache = {};
+    }
   }
 
   static getInfrastructure() {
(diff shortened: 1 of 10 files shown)


lib/bosh/BoshDirectorClient.js
```diff
@@ -64,26 +69,54 @@ class BoshDirectorClient extends HttpClient {
...
-            }
-          }));
-      })
-      .tap(() => {
+      .map(config.directors,
+        (directorConfig) => this.getDeploymentsByConfig(directorConfig))
+      .finally(() => {
         this.cacheLoadInProgress = false;
-        logger.debug('Cached Deployments:', _.keys(this.cache));
+        logger.info('Clearing cacheLoadInProgress flag. Bosh DeploymentName cache is loaded.');
```
sauravmndl: For which instance/deployment keep a reference.

lib/bosh/BoshDirectorClient.js
```diff
@@ -64,26 +69,54 @@ class BoshDirectorClient extends HttpClient {
...
+  getDeploymentNamesFromCache(boshName, attempt) {
+    return Promise.try(() => {
+      if (this.cacheLoadInProgress) {
+        if (!attempt) {
+          attempt = 1;
+        } else if (attempt > CONST.BOSH_POLL_MAX_ATTEMPTS) {
+          throw errors.Timeout.toManyAttempts(CONST.BOSH_POLL_MAX_ATTEMPTS, new Error('Fetching deployments from Cache is taking too long.'));
+        }
+        logger.info(`Cache load in progress. GetDeploymentNames will be delayed by 500 ms - current attempt ${attempt}`);
+        return Promise.delay(500 * attempt).then(() => this.getDeploymentNamesFromCache(boshName, ++attempt));
```
sauravmndl: 500 put in a constant and use that in logger statement in above also.

lib/bosh/BoshDirectorClient.js
```diff
@@ -93,15 +126,13 @@ class BoshDirectorClient extends HttpClient {
...
+        logger.silly('found director in cache...', cache_val.name);
         return cache_val;
       }
       logger.debug('cache miss for..', deploymentName);
       if (this.cacheLoadInProgress) {
         logger.debug('Cache load in progress.. deferring execution..', this.cacheLoadInProgress);
-        return Promise.delay(100).then(() => {
-          return this.getDirectorConfig(deploymentName);
-        });
+        return Promise.delay(500).then(() => this.getDirectorConfig(deploymentName));
```
sauravmndl: 500 put in constant.

lib/bosh/BoshDirectorClient.js
jagadish-kb: Will take it up post hotfix.

lib/bosh/BoshDirectorClient.js
jagadish-kb: Will take it up post hotfix.

Issue #52, opened 2017-09-16
Update of instances creates a new instance on active BOSH.


muralisc:
**Description**

Update of instance created by service-fabrik fails but creates a new instance on the currently active
BOSH.
Expected Behaviour: Update should not fail , No new instance should be created. 

**Steps to reproduce the issue:**
1. Deploy a dedicated instance in BOSH
2. Change service-fabrik code to work with multiple bosh
3. Do update of the instance.

PR #53, opened 2017-09-16
Closes #50 #52

### LINKED ISSUE #50: Cannot choose Bosh director while creating service instance
**Description**
With the current implementation during instance creation service-fabrik the active bosh director.
This behaviour may need to be overridden in some case.

To facilitate this a parameter could be passed along with cf create command which will specify the bosh director name.

muralisc:
This commit enables

* service-fabrik-broker to recognize json parameter
  passed to `cf create-service` command and deploy to the BOSH specified.

  ```
  {
   bosh_director_name: <bosh>,
   username: <user_name>,
   password: <password>
  }
  ```

  This bosh_director_name if provided is used to chose the bosh where
  the deployment is to be created.  If no director_name is provided the
  default bosh is used for create.

* With multiple BOSH updates to instances in old BOSH were failing and
a new instance was created in the currently active BOSH.
Any create/update request first checks to see the BOSH in which the
instance exist and then perform update on that BOSH.

### DIFF
--- a/lib/bosh/BoshDirectorClient.js
+++ b/lib/bosh/BoshDirectorClient.js
@@ -304,7 +304,13 @@ class BoshDirectorClient extends HttpClient {
   createOrUpdateDeployment(manifest, opts) {
     const query = opts ? _.pick(opts, 'recreate', 'skip_drain', 'context') : undefined;
     const deploymentName = yaml.safeLoad(manifest).name;
-    const config = _.sample(this.activePrimary);
+    const boshDirectorName = _.get(opts, 'parameters.bosh_director_name');
+    let config;
+    if (boshDirectorName) {
+      config = this.getConfigByName(boshDirectorName);
+    } else {
+      config = _.sample(this.activePrimary);
+    }
     if (config === undefined || config.length === 0) {
       throw new errors.NotFound('Did not find any bosh director config which supports creation of deployment');
     }
--- a/lib/bosh/BoshDirectorClient.js
+++ b/lib/bosh/BoshDirectorClient.js
@@ -305,30 +305,39 @@ class BoshDirectorClient extends HttpClient {
     const query = opts ? _.pick(opts, 'recreate', 'skip_drain', 'context') : undefined;
     const deploymentName = yaml.safeLoad(manifest).name;
     const boshDirectorName = _.get(opts, 'parameters.bosh_director_name');
-    let config;
-    if (boshDirectorName) {
-      config = this.getConfigByName(boshDirectorName);
-    } else {
-      config = _.sample(this.activePrimary);
-    }
-    if (config === undefined || config.length === 0) {
-      throw new errors.NotFound('Did not find any bosh director config which supports creation of deployment');
-    }
-    return this
-      .makeRequestWithConfig({
-        method: 'POST',
-        url: '/deployments',
-        headers: {
-          'Content-Type': 'text/yaml'
-        },
-        qs: query,
-        body: _.isObject(manifest) ? yaml.safeDump(manifest) : manifest
-      }, 302, config)
-      .tap(() => {
-        logger.info(`Cached ${deploymentName} at director: ${config.name} ${config.url}`);
-        this.cache[deploymentName] = config;
+    return Promise.try(() => {
+        if (boshDirectorName) {
+          logger.info(`Deploying through ${boshDirectorName}`);
+          return this.getConfigByName(boshDirectorName);
+        } else {
+          return this
+            .getDirectorConfig(deploymentName) // if update
+            .catch(() => _.sample(this.activePrimary)); // if create
+        }
       })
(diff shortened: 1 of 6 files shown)


muralisc:
Test failing due to npm install issue.

lib/fabrik/DirectorManager.js
```diff
@@ -186,6 +188,21 @@ class DirectorManager extends BaseManager {
           return;
         }
       })
+      .then(() => {
+        const username = _.get(args, 'parameters.username');
+        const password = _.get(args, 'parameters.password');
+        if (action === 'create' && username !== undefined) {
```
jagadish-kb: This check is not done in case of update & this is currently buggy, because for an update operation user can pass bosh name and it would skip this validation and endup creating a new deployment on a new bosh (potentially). Need to further fine tune these checks. May be you can do the following instead:
@line 178 - within update case switch block, you can check if username/password params are passed and if they are then throw exception 'BadRequest'
@line 188 - Move existing code from 192 - 204

lib/bosh/BoshDirectorClient.js
```diff
@@ -304,25 +304,40 @@ class BoshDirectorClient extends HttpClient {
...
-        body: _.isObject(manifest) ? yaml.safeDump(manifest) : manifest
-      }, 302, config)
-      .tap(() => {
-        logger.info(`Cached ${deploymentName} at director: ${config.name} ${config.url}`);
-        this.cache[deploymentName] = config;
+    const boshDirectorName = _.get(opts, 'parameters.bosh_director_name');
+    return Promise.try(() => {
+        if (boshDirectorName) {
+          logger.info(`Deploying through ${boshDirectorName}`);
+          return this.getConfigByName(boshDirectorName);
```
jagadish-kb: Just to be safe here, put the following  additional checks here:
1. `this.getDirectorConfig(deploymentName)` must null - if null, then return `this.getConfigByName(boshDirectorName)` else throw exception. In no scenario can someone call an update with passing director name.

lib/bosh/BoshDirectorClient.js
```diff
@@ -304,25 +304,40 @@ class BoshDirectorClient extends HttpClient {
...
-        qs: query,
-        body: _.isObject(manifest) ? yaml.safeDump(manifest) : manifest
-      }, 302, config)
-      .tap(() => {
-        logger.info(`Cached ${deploymentName} at director: ${config.name} ${config.url}`);
-        this.cache[deploymentName] = config;
+    const boshDirectorName = _.get(opts, 'parameters.bosh_director_name');
+    return Promise.try(() => {
+        if (boshDirectorName) {
+          logger.info(`Deploying through ${boshDirectorName}`);
```
jagadish-kb: can you change this log level to warn. These are far and few and better to see them at warn level so that they stand out.

lib/fabrik/DirectorManager.js
```diff
@@ -168,6 +169,7 @@ class DirectorManager extends BaseManager {
     const previousValues = _.get(params, 'previous_values');
     const action = _.isPlainObject(previousValues) ? 'update' : 'create';
     const opts = _.omit(params, 'previous_values');
+    args = _.assign(args, _.pick(params, 'parameters'));
```
jagadish-kb: no need to pass the input username/password downstream via args object as its not going to be used. With credentials we must only pass them where required. So remove this set here and at line 192 & 193, directly work with params object instead.

lib/bosh/BoshDirectorClient.js
muralisc: done.

lib/bosh/BoshDirectorClient.js
muralisc: Done

lib/fabrik/DirectorManager.js
muralisc: Done

lib/fabrik/DirectorManager.js
muralisc: Done

lib/bosh/BoshDirectorClient.js
```diff
@@ -304,25 +304,47 @@ class BoshDirectorClient extends HttpClient {
...
+            .getDirectorConfig(deploymentName) // if update
+            .catch(() => _.sample(this.activePrimary)); // if create
+        }
       })
-      .then(res => this.prefixTaskId(deploymentName, res));
+      .then(config => {
+        if (config === undefined || config.length === 0) {
+          throw new errors.NotFound('Did not find any bosh director config which supports creation of deployment');
+        }
+        return config;
```
jagadish-kb: unnecessary promise chain. Please club the below `.then` block with this one and directly return result.

muralisc: done

PR #59, opened 2017-10-03
Feature- Scheduled update of service instance

### LINKED ISSUE #10: Scheduled update window for service instances
Currently, SF supports update of all services instances. But there is no way one can schedule update of their service instances. Ideally, user/operator should be able to define their service instance update window so that productive service instances are not updated at unwanted time.

jagadish-kb:
Closes #10


JobScheduler.js
```diff
@@ -1,84 +1,215 @@
...
+      //Agenda Scheduler for each worker will be configured to processJobQueue once every ${maxWorkers} minutes.
+      //So each worker job is started with a delay of 1 minute, as it creates a constant sliding window of 1 min
+      setTimeout(() => this.addJobWorker(), delay);
+    }
+  }
+
+  initWorker() {
+    this.workerType = `Worker - ${cluster.worker.id} - ${process.pid}`;
+    logger.info(`Starting Service Fabrik Batch Job worker: ${cluster.worker.id} - ${process.pid}  @${new Date()}`);
+    require('./lib/jobs');
```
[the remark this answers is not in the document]
jagadish-kb: This is done to force the loading of jobs classes specifically from the context of worker processes. So there is nothing wrong in this per se. These are not like imports of Java which are to be defined at the top of the class.

config/settings.yml
```diff
@@ -66,7 +66,9 @@ defaults: &defaults
   colorize_log: false
   sys_log_level: info
   enable_circuit_breaker: true
-
+  feature:
+    ServiceInstanceAutoUpdate: true #Switch to turn on / turn off schedule_update feature
```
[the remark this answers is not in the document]
jagadish-kb: It is generally kept in snake case. Initially as well this was snake case, however wanted this job name to match the name of job defined in [constants.js](https://github.com/jagadish-kb/service-fabrik-broker/blob/1f47ff72fa6fe2d8ecbee221eacd4a36048e47e5/lib/constants.js#L52) 
felt its better to violate this in this special case as its better to have the same feature name across yml/js files.

lib/constants.js
```diff
@@ -114,7 +135,9 @@ module.exports = Object.freeze({
   ERR_CODES: {
     UNKNOWN: 'ERR-CODE-UNKNOWN',
     PRE_CONDITION_NOT_MET: 'PRE_CONDITION_NOT_MET',
-    DEPLOYMENT_NAME_DUPED_ACROSS_DIRECTORS: 'DEPLOYMENT_NAME_DUPED_ACROSS_DIRECTORS'
+    DEPLOYMENT_NAME_DUPED_ACROSS_DIRECTORS: 'DEPLOYMENT_NAME_DUPED_ACROSS_DIRECTORS',
+    SF_IN_MAINTENANCE: 101,
```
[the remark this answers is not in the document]
jagadish-kb: This error code is used as process exit code &  [nodejs process exit codes](https://nodejs.org/api/process.html#process_exit_codes) are always numeric.

Nodejs does not support enum data types.

JobScheduler.js
```diff
@@ -1,84 +1,215 @@
...
+      });
+  }
+
+  processUnhandledRejection(reason, p) {
+    logger.error(`Unhandled Rejection in - ${this.workerType} - at:`, p, 'reason:', reason);
+    if (reason && reason instanceof errors.DBUnavailable) {
+      logger.error('DB unavailable. shutting down app');
+      this.shutDown();
+    } else {
+      logger.error('Unhandled Rejection at:', p, 'reason:', reason);
```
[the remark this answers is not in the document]
jagadish-kb: No. Ideally all code must handle the rejections / errors as part of their code only. And we also have a global error handler in middleware.js which takes care of any unhandled error globally and responds back. 
This is put in place for two things:
1. If there are some places (which run in background) in code which are not handling errors (which necessarily need not be fatal) are logged, so that they can be fixed at source and rejections are handled.
2. From a background process like (DBConnectionManager) when we want to throw an exception (very deterministic error) which should result in shutdown, then a shutdown is to be initiated. In future if there is another deterministic error which must result in shutdown then the same error type must be added in the above. 

Shutting down on any unhandled rejection is not required.

lib/fabrik/DirectorInstance.js
```diff
@@ -73,7 +73,9 @@ class DirectorInstance extends BaseInstance {
       .try(() => {
         switch (operation.type) {
         case 'create':
-          return this.createSecurityGroup(operation.space_guid);
+          return this
+            .createSecurityGroup(operation.space_guid)
+            .then(() => operation.state === CONST.OPERATION.SUCCEEDED ? this.scheduleAutoUpdate() : {});
```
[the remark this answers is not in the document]
jagadish-kb: Logic is:
1. Check If ScheduleUpdateFeature enabled
2. If yes, invoke the scheduleUpdate API. 
So DirectorInstance is invoking the scheduleupdate API and there is no dependency on scheduler component per se.  Whether scheduler is running or not is taken care of as part of feature enablement check.

Everything related to setting up schedule is encapsulated within the inner method `scheduleAutoUpdate`.  No need to have this split across.

lib/bosh/BoshDirectorClient.js
```diff
@@ -126,7 +127,7 @@ class BoshDirectorClient extends HttpClient {
 
   getDirectorConfig(deploymentName) {
     return Promise.try(() => {
-      logger.info(`Finding the correct director config for:`, deploymentName);
+      logger.debug(`Finding the correct director config for:`, deploymentName);
```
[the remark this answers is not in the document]
jagadish-kb: Ok. Done.

lib/constants.js
```diff
@@ -40,7 +48,12 @@ module.exports = Object.freeze({
     SCHEDULED_OOB_DEPLOYMENT_BACKUP: 'ScheduledOobDeploymentBackup',
     OPERATION_STATUS_POLLER: 'OperationStatusPoller',
     BLUEPRINT_JOB: 'BluePrintJob',
-    BAKUP_REAPER: 'BackupReaper'
+    BAKUP_REAPER: 'BackupReaper',
```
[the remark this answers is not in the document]
jagadish-kb: This typo was not part of this feature. Nevertheless fixed :-)

JobScheduler.js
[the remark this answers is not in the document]
subhankarc: Ok fine.

JobScheduler.js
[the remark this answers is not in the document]
kamath-prasad: Ok , got it.

config/settings.yml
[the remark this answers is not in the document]
kamath-prasad: Ok.

lib/jobs/ServiceInstanceUpdateJob.js
```diff
@@ -0,0 +1,265 @@
...
+      start_after_hr: afterHr,
+      start_before_hr: beforeHr,
+      start_after_min: afterMin,
+      start_before_min: beforeMin
+    };
+    return utils.getRandomCronForOnceEveryXDays(runOnceEvery, opts);
+  }
+
+  static getLastRunStatus(name) {
+    const sortOn = ['createdAt', 'asc'];
```
kamath-prasad: shouldnt the sort criteria be desc for getting the last run ?

lib/jobs/ServiceInstanceUpdateJob.js
```diff
@@ -0,0 +1,265 @@
...
+        logger.info('LastRun details retrieved - ', lastRunDetails);
+        if (_.get(lastRunDetails, 'totalRecordCount', 0) === 0) {
+          return null;
+        }
+        const lastRunList = lastRunDetails.list;
+        const count = lastRunList.length;
+        let initialJobRunStatus, idx;
+        for (idx = count - 1; idx >= 0; idx--) {
+          if (lastRunList[idx].data.attempt === 1) {
+            initialJobRunStatus = lastRunList[idx];
```
kamath-prasad: since the sort option is asc above initialjobrunstatus will be at index 0 , correct ?

lib/constants.js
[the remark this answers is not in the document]
jagadish-kb: Ok. Comment added.

config/settings.yml
```diff
@@ -120,15 +122,30 @@ defaults: &defaults
...
     run_with_web_process: true #if this flag is either undefined or set to true, then agenda scheduler can run along within web process
     max_workers: 2 # Can throttle the number of job workers to a value lesser than default number of cpus
     max_concurrency: 30
     default_concurrency: 20
+    start_delay: 120000 # Scheduler start must be delayed for DB connections to happen.
     default_lock_lifetime: 180000
+    maintenance_check_interval: 90000 #every (ms) scheduler checks for maitenance mode
     agenda_collection: agendaJobDetails #Collectin name that is to be used by agenda scheduler
-    job_history_retention_in_days: 30
+    maintenance_mode_time_out: 36000000 # (ms) - 60 minutes wait for sf deployment to complete
```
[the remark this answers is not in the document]
jagadish-kb: Its kept configurable in [bosh release](https://github.com/jagadish-kb/service-fabrik-boshrelease/blob/T9A_scheduled_update/jobs/service-fabrik-broker/templates/config/settings.yml.erb#L108) as well as product component. 
If required we can override these values from landscape config.

JobScheduler.js
```diff
@@ -1,84 +1,215 @@
 'use strict';
```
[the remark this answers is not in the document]
jagadish-kb: 1. No - Every feature in the system has some pre-requisites. For scheduled updates the pre-req is that DB must be configured & serviceinstanceautoupdate job must be enabled. [See here](https://github.com/jagadish-kb/service-fabrik-broker/blob/master/lib/utils/index.js#L266). 
Now backup / restore are cloud qualities that are to be implemented & need not be tied to each and every feature that needs DB. Now if backup/restore does not exist/work, then it is a risk per-se, but its not a true dependency. Also in case of schedule update, we dont store any critical information. Just random schedules / run history. So even if DB crashes and we cant recover, its not too critical.

2. Polling is a feature for background jobs. If some one tries operations from UI / Cmd line, then they are provided with response to their request immediately. Don't need any polling for UI / Cmd line operations. The reason why background jobs are stopped from running during maintenance is because, if a particular job run fails then it will not run till its next scheduled time. However if an operation fails when user initiates he can always trigger it after sometime.  Even if system is in maintenance & if user initiates an operation then either broker times out or broker returns back successfully. If its asynch operations bosh takes care of performing that operation and cf keeps pining with last operation and if its docker related its going to be synch response.
So really don't see any reasons to block user from performing operations.

JobScheduler.js
```diff
@@ -1,84 +1,215 @@
...
 
-process.on('unhandledRejection', (reason, p) => {
-  if (reason && reason instanceof errors.DBUnavailable) {
-    logger.error('DB unavailable. shutting down app');
-    notifyShutDown();
-  } else {
-    logger.error('Unhandled Rejection at:', p, 'reason:', reason);
+  pollMaintenanceStatus() {
+    const checkMaintenanceStatus = (resolve, reject) => {
+      return maintenanceManager
```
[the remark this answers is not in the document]
jagadish-kb: We agreed that in fresh landscape maintenance_start / end APIs will not be getting invoked. Once scheduler starts it errors out because it cant connect to DB & it kills itself since it [could not determine maintenance status](https://github.com/jagadish-kb/service-fabrik-broker/blob/master/JobScheduler.js#L191) & monit keeps brining back scheduler process. Once DB is provisioned the next moment, DB ConnectionManager connects and I guess it gets back maintenance status as null and then workers get started. Haven't tested this scenario though, but this is how it should work :-)

lib/jobs/ScheduleManager.js
```diff
@@ -123,7 +160,41 @@ class ScheduleManager {
...
+      criteria.push({
+        createdAt: {
+          $lt: retentionDate
+        }
+      });
+      criteria.push({
+        runOnlyOnce: true
+      });
+      criteria.push({
+        type: '/.*_[0-9]+/'
```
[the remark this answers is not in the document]
jagadish-kb: any completed job type `'/.*_[0-9]+/'` which are older than`retentionDate` of type `runOnlyOnce` should be purged.

lib/jobs/Scheduler.js
```diff
@@ -121,28 +124,44 @@ class Scheduler {
...
-            try {
-              this.define(jobType, JobFabrik.getJob(jobType));
-            } catch (err) {
-              logger.error(`error occurred while registering job defintion : ${jobType}`, err);
-            }
-          });
-        } else {
-          return Promise.reject(new Error('No jobs configured in external app configuration. Scheduler not started'));
-        }
+        _.each(this.jobTypeList, (jobType) => {
```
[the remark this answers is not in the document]
jagadish-kb: `if (this.jobTypeList.length > 0) ` was unnecessary code. `_.each(this.jobTypeList, (jobType)` enters only if array length is > 0. So removed unnecessary code as part of making code coverage 100%. I had two choices:
1. Remove unwanted code
2. Write tests to satisfy else clause for unnecessary code.
Chose the former :-) 
Hence having 100% code coverage not only helps ensuring tests. It also ensures we dont have unnecessary code.

lib/jobs/ServiceInstanceUpdateJob.js
```diff
@@ -0,0 +1,265 @@
...
+            logger.info(`${instanceDetails.instance_name} instance update initiated. Status : `, operationResponse);
+            //Even in case of successful initiation the job is rescheduled so that after the reschedule delay,
+            //when this job comes up it must see itself as updated.
+            //This is to handle any Infra errors that could happen post successful initiation of update. (its a retry mechanism)
+            this
+              .rescheduleUpdateJob(instanceDetails, trackAttempts);
+          });
+      });
+  }
+
```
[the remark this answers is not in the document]
jagadish-kb: Cant reuse this for following reasons:
1. Jobs should not refer to Controller classes. Dependency should be always downwards. Cant have jobs refer to  Controller. However moved the common code to Utils and referred same from Job/Controller.
2. Its just a wrapper around `this.cloudController.findServicePlanByInstanceId` - so already refactored.
3. Not too much common code. Its already reusing same code from directormanager.

So only #1 of three needs refactor at both places and its done.

lib/jobs/ServiceInstanceUpdateJob.js
jagadish-kb: Yes. You are right. Initial logic required asc, but subsequently that was modified but the sort order was never changed. Good catch. Fixed it.

lib/jobs/ServiceInstanceUpdateJob.js
jagadish-kb: changed sort order per above comment. Also we are fetching last 3 run history records we cant always assume idx = 0 (asc) or idx = 2 (desc) to be initial run status, because many a times the update might complete with 1 attempt, in which case at idx=1, initialJobStatus will be found.

config/settings.yml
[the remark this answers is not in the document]
kamath-prasad: ok..

JobScheduler.js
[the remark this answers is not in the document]
kamath-prasad: Ok,thanks for the detailed explanation.

lib/constants.js
```diff
@@ -5,6 +5,8 @@ module.exports = Object.freeze({
   BOSH_POLL_MAX_ATTEMPTS: 3,
   DEPLOYMENT_LOCK_NAME: '_LOCK_',
   SERVICE_FABRIK_PREFIX: 'service-fabrik',
+  NOT_APPLICABLE: 'NA',
+  FINISHED_JOBS_RETENTION_DURATION_DAYS: 14,
```
sauravmndl: is it better to make configurable the value 14?

lib/controllers/ServiceFabrikAdminController.js
```diff
@@ -623,6 +601,75 @@ class ServiceFabrikAdminController extends BaseController {
...
+      .getMaintenaceInfo()
+      .then(body => {
+        if (body === null) {
+          res
+            .status(404)
+            .send({
+              system_in_maintenance: false
+            });
+        } else {
+          _.set(body, 'system_in_maintenance', true);
```
sauravmndl: is it better to put the value 'system_in_maintenance' in constant ?

lib/controllers/ServiceFabrikAdminController.js
```diff
@@ -623,6 +601,75 @@ class ServiceFabrikAdminController extends BaseController {
...
+          return maintenanceManager
+            .updateMaintenace(req.body.progress, req.body.state, req.user)
+            .then(body => res
+              .status(200)
+              .send(body));
+        } else {
+          res
+            .status(403)
+            .send({
+              errorMessage: 'System not in maintenance mode'
```
sauravmndl: Better to define error messages in some constants.

lib/controllers/ServiceFabrikAdminController.js
```diff
@@ -623,6 +601,75 @@ class ServiceFabrikAdminController extends BaseController {
...
+          return maintenanceManager
+            .startMaintenace(req.body, req.user)
+            .then(body => res
+              .status(201)
+              .send(body));
+        } else {
+          res
+            .status(403)
+            .send({
+              errorMessage: 'System already in maintenance mode',
```
sauravmndl: Better to define error messages in some constants.

lib/controllers/ServiceFabrikAdminController.js
```diff
@@ -623,6 +601,75 @@ class ServiceFabrikAdminController extends BaseController {
...
+              .send(body));
+        } else {
+          res
+            .status(403)
+            .send({
+              errorMessage: 'System not in maintenance mode'
+            });
+        }
+      });
+
```
sauravmndl: extra new line required ?

lib/controllers/ServiceFabrikAdminController.js
```diff
@@ -623,6 +601,75 @@ class ServiceFabrikAdminController extends BaseController {
...
+            .send(body);
+        }
+      });
+  }
+
+  getMaintenanceHistory(req, res) {
+    return maintenanceManager
+      .getMaintenaceHistory(req.query.offset, req.query.records, req.query.sortBy, req.query.sortOrder)
+      .then(body => res
+        .status(200)
```
sauravmndl: we can put error codes in constant as those are used multiple times.

lib/controllers/ServiceFabrikApiController.js
```diff
@@ -474,6 +475,63 @@ class ServiceFabrikApiController extends BaseController {
         .send({}));
   }
 
+  scheduleUpdate(req, res) {
+    req.manager.isAutoUpdatePossible();
```
sauravmndl: not relying on the value of the method isAutoUpdatePossible() ?

lib/controllers/ServiceFabrikApiController.js
```diff
@@ -474,6 +475,63 @@ class ServiceFabrikApiController extends BaseController {
...
+    if (_.isEmpty(req.body.repeatInterval)) {
+      throw new BadRequest('repeatInterval is mandatory');
```
sauravmndl: keep some reference from the request (e.g. instance id or / deployment name etc.), otherwise in huge log difficult to debug for multiple occurrence of this type.

lib/controllers/ServiceFabrikApiController.js
```diff
@@ -474,6 +475,63 @@ class ServiceFabrikApiController extends BaseController {
...
+        timeZone: req.body.timeZone
+      }))
+      .then((jobData) => ScheduleManager
+        .schedule(req.params.instance_id,
+          CONST.JOB.SERVICE_INSTANCE_UPDATE,
+          req.body.repeatInterval,
+          jobData,
+          req.user))
+      .then(body => res
+        .status(201)
```
sauravmndl: put status code in constants

lib/controllers/ServiceFabrikApiController.js
```diff
@@ -474,6 +475,63 @@ class ServiceFabrikApiController extends BaseController {
...
+        .send(body));
+  }
+
+  getUpdateSchedule(req, res) {
+    req.manager.isAutoUpdatePossible();
+    return ScheduleManager
+      .getSchedule(req.params.instance_id, CONST.JOB.SERVICE_INSTANCE_UPDATE)
+      .then(scheduleInfo => {
+        const checkUpdateRequired = _.get(req.query, 'check_update_required');
```
sauravmndl: can we put 'check_update_required' in constant?

lib/controllers/ServiceFabrikApiController.js
```diff
@@ -474,6 +475,63 @@ class ServiceFabrikApiController extends BaseController {
...
+        logger.info('check outdated status -', checkUpdateRequired);
```
sauravmndl: keep reference of instance id.

lib/controllers/ServiceFabrikApiController.js
```diff
@@ -474,6 +475,63 @@ class ServiceFabrikApiController extends BaseController {
...
+              scheduleInfo.update_required = result.diff && result.diff.length > 0;
+              scheduleInfo.update_details = result.diff;
+              return scheduleInfo;
+            });
+        } else {
+          return scheduleInfo;
+        }
+      })
+      .then(body => res
+        .status(200)
```
sauravmndl: status code in constants.

lib/jobs/JobFabrik.js
```diff
@@ -1,13 +1,8 @@
...
 
 const assert = require('assert');
-let ScheduleBackupJob;
-let ServiceFabrikBackupJob;
-let ScheduledOobDeploymentBackupJob;
-let OperationStatusPollerJob;
-let BackupReaperJob;
-let bluePrintJob;
 const CONST = require('../constants');
+let ScheduleBackupJob, ScheduledOobDeploymentBackupJob, OperationStatusPollerJob, BackupReaperJob, ServiceInstanceUpdateJob, DbCollectionReaperJob, bluePrintJob;
```
sauravmndl: you can rename bluePrintJob ==> 'BluePrintJob' to be consistent with others .

lib/jobs/ScheduleBackupJob.js
```diff
@@ -18,43 +21,50 @@ const ScheduleManager = require('./ScheduleManager');
...
-    return this
-      .isServiceInstanceDeleted(options.instance_id)
-      .tap(deleteStatus => instanceDeleted = deleteStatus)
-      .then(() => {
-        if (instanceDeleted) {
-          return 'instance_deleted';
+    return Promise.try(() => {
+        job.__started_At = new Date();
+        const jobData = job.attrs.data;
+        logger.info(`-> Starting ScheduleBackupJob -  name: ${job.attrs.data[CONST.JOB_NAME_ATTRIB]} - with options: ${JSON.stringify(jobData)} `);
```
sauravmndl: you can make use of jobData instead of job.attrs.data.

JobScheduler.js
```diff
@@ -1,84 +1,215 @@
...
-    //So each worker job is started with a delay of 1 minute, as it creates a constant sliding window of 1 min
-    setTimeout(addJobWorker, delay);
+
+class JobScheduler {
+  constructor() {
+    this.ready = Promise.try(() => {
+      this.jobWorkers = [];
+      this.workerCount = 0;
+      this.workerType = '';
+      this.serviceFabrikInMaintenance = true;
```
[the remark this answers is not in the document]
sauravmndl: +1

JobScheduler.js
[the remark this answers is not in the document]
kamath-prasad: Ok , got it.

lib/jobs/ScheduleManager.js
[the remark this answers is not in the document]
kamath-prasad: ok.

lib/jobs/Scheduler.js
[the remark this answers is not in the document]
kamath-prasad: 👍

lib/jobs/ServiceInstanceUpdateJob.js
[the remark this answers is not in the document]
kamath-prasad: Yes ,I meant moving to utils class . Thanks !

lib/fabrik/DirectorInstance.js
```diff
@@ -318,6 +320,32 @@ class DirectorInstance extends BaseInstance {
...
+    return utils
+      .retry(tries => {
+        logger.info(`+-> ${CONST.ORDINALS[tries]} attempt to schedule auto update for : ${this.guid}`);
+        if (utils.isFeatureEnabled(CONST.FEATURE.SCHEDULED_UPDATE)) {
+          return this
+            .serviceFabrikClient
+            .scheduleUpdate(options)
+            .catch(err => {
+              logger.error(`Error occurred while setting up auto update for : ${this.guid}`, err);
+              throw err;
```
sauravmndl: Will this declare service instance creation as failed? By this time all operations related to instance creation is done e.g. entry in bosh, security group in cf etc.

lib/fabrik/DirectorManager.js
```diff
@@ -113,6 +118,20 @@ class DirectorManager extends BaseManager {
       .tap(networkSegmentIndex => logger.info(`+-> Aquired network segment index '${networkSegmentIndex}'`));
   }
 
+  findDeploymentNameByInstanceId(guid) {
+    logger.info(`Finding deployment name with instance id : '${guid}'`);
+    return this.getDeploymentNames(false)
+      .then(deploymentNames => {
+        const deploymentName = _.find(deploymentNames, name => _.endsWith(name, guid));
+        if (!deploymentName) {
+          logger.warn('+-> Could not find a matching deployment');
```
sauravmndl: keep a reference of instance guid.

lib/jobs/ScheduleManager.js
```diff
@@ -62,7 +87,7 @@ class ScheduleManager {
         logger.debug(`Job retrieved from agenda for : ${name} - ${jobType}`);
       })
       .then((job) => {
-        if (job === null || (job.constructor === Object && Object.keys(job).length === 0)) {
+        if (job === null) {
           throw new errors.NotFound(`Schedule not found for instance ${name}`);
```
sauravmndl: also put job type in message.

JobScheduler.js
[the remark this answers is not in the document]
jagadish-kb: `this.serviceFabrikInMaintenance` is a flag indicating the if SF is in maintenance
`pollMaintenanceStatus()` is a function, its not a flag & this function keeps polling the status till SF is in maintenance.

lib/constants.js
```diff
@@ -114,7 +135,11 @@ module.exports = Object.freeze({
   ERR_CODES: {
     UNKNOWN: 'ERR-CODE-UNKNOWN',
     PRE_CONDITION_NOT_MET: 'PRE_CONDITION_NOT_MET',
-    DEPLOYMENT_NAME_DUPED_ACROSS_DIRECTORS: 'DEPLOYMENT_NAME_DUPED_ACROSS_DIRECTORS'
+    DEPLOYMENT_NAME_DUPED_ACROSS_DIRECTORS: 'DEPLOYMENT_NAME_DUPED_ACROSS_DIRECTORS',
+    SF_IN_MAINTENANCE: 101,
+    INTERNAL_ERROR: 4
+    //Error codes should always be readable strigns. However few error codes (used as process exit codes) must be int.
```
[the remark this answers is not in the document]
jagadish-kb: Ok. Fixing it.

lib/jobs/ScheduleManager.js
```diff
@@ -3,32 +3,48 @@
...
       .then(jobInDB => this.getJobAttrs(jobInDB, agendaJob));
   }
 
+  static schedule(name, jobType, interval, jobData, user) {
+    return Promise.try(() => {
+      let agendaJob = {};
+      if (interval === CONST.SCHEDULE.DAILY) {
+        interval = this.getRandomDailySchedule();
+        logger.info(`Schedule interval input as 'daily'. So setting following daily random schedule - ${interval}`);
+      } else if (interval === CONST.SCHEDULE.RANDOM) {
```
sauravmndl: the repeat interval is implemented by 'ServiceInstanceUpdateJob' only. If for other job, let's say for Backup scheduler if user inputs random as repeat interval will it fail ??

lib/utils/index.js
```diff
@@ -291,4 +330,25 @@ function isServiceFabrikOperation(params) {
 
 function isServiceFabrikOperationFinished(state) {
   return _.includes([CONST.OPERATION.SUCCEEDED, CONST.OPERATION.FAILED, CONST.OPERATION.ABORTED], state);
+}
+
+function isChangesPresentInForbiddenSections(diff) {
```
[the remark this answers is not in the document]
jagadish-kb: its a general programming norm to prefix functions that return back boolean value with 'is' - hence this naming. Anyways will rename it :-)

lib/constants.js
jagadish-kb: Thought about this. These are useless one time job definitions which are no longer valid. No real need to change the configuration of these on the fly. Job runs are anyways retained. There would not be any time a pressing need to change them immediately.

lib/controllers/ServiceFabrikAdminController.js
jagadish-kb: As discussed these are coding standards that exist all across code base. Also these attribs are used at one off place. So no need to really parameterize them.

lib/controllers/ServiceFabrikAdminController.js
jagadish-kb: As discussed HTTP Status codes are used in similar fashion in all controllers and just changing in couple of methods would lead to confusion to external reviewers. Leaving this as-is as agreed.

lib/controllers/ServiceFabrikAdminController.js
jagadish-kb: As discussed currently we do not yet have a defined way of defining all messages in a single central location. Just changing it for this feature would lead it to be a bit confusing. Leaving this as-is as agreed.

lib/controllers/ServiceFabrikAdminController.js
```diff
@@ -623,6 +601,75 @@ class ServiceFabrikAdminController extends BaseController {
...
+      });
+  }
+
+  updateMaintenance(req, res) {
+    return maintenanceManager
+      .getMaintenaceInfo()
+      .then(body => {
+        if (body !== null) {
+          return maintenanceManager
+            .updateMaintenace(req.body.progress, req.body.state, req.user)
```
sauravmndl: The name req.body.progress ==> req.body.progressInfo/maintenanceInfo  would have been better. Or 
Instead of updateMaintenace(req.body.progress, req.body.state, req.user)
 --> updateMaintenace(req.body, req.user) would look cleaner if body doesn't contain other info which should not be passed to updateMaintenace() method.

lib/controllers/ServiceFabrikAdminController.js
jagadish-kb: removed

lib/controllers/ServiceFabrikApiController.js
jagadish-kb: The method `isAutoUpdatePossible` must throw error in case its [not supported](https://github.com/jagadish-kb/service-fabrik-broker/blob/master/lib/fabrik/DockerManager.js#L22). No need to explicitly check response from method.

lib/maintenance/MaintenanceManager.js
```diff
@@ -0,0 +1,102 @@
...
+        if (state !== undefined) {
+          if (_.chain(CONST.OPERATION).valuesIn().indexOf(state).value() === -1) {
+            throw new errors.BadRequest(`Maintenance state can be only one of these values : ${_.valuesIn(CONST.OPERATION)}`);
+          }
+          if (utils.isServiceFabrikOperationFinished(state)) {
+            model.completedAt = new Date();
+          }
+          model.state = state;
+        }
+        return Repository.save(CONST.DB_MODEL.MAINTENANCE_DETAIL, model, user);
```
sauravmndl: You can put a logger before this saying updating with state, so that it completes the circle of start and update of maintenance.

lib/controllers/ServiceFabrikApiController.js
jagadish-kb: Not required. Exception that gets thrown from here gets logged by EventLogInterceptor which logs all the details related to current request. So we will not miss the context.

lib/controllers/ServiceFabrikApiController.js
jagadish-kb: Same as [above](https://github.com/SAP/service-fabrik-broker/pull/59#discussion_r143665557)

lib/utils/index.js
```diff
@@ -278,11 +290,38 @@ function taskIdRegExp() {
...
 function getRandomInt(min, max) {
   min = Math.ceil(min);
   max = Math.floor(max);
-  return Math.floor(Math.random() * (max - min)) + min;
+  const factor = max - min === 1 ? 2 : (max - min);
+  //If we want a random of just 2 numbers then the factor must be 2, else it will always return back the lesser of two number always.
+  return Math.floor(Math.random() * (factor)) + min;
+}
+
+function getRandomCronForOnceEveryXDays(days, options) {
```
sauravmndl: any library available for these kinds of operations?

lib/controllers/ServiceFabrikApiController.js
jagadish-kb: same as [above](https://github.com/SAP/service-fabrik-broker/pull/59#discussion_r143665281)

lib/controllers/ServiceFabrikApiController.js
jagadish-kb: Ok. Done.

lib/controllers/ServiceFabrikApiController.js
jagadish-kb: Same as [above](https://github.com/SAP/service-fabrik-broker/pull/59#discussion_r143665557)

lib/jobs/JobFabrik.js
jagadish-kb: Done

lib/jobs/ScheduleBackupJob.js
jagadish-kb: Even though theres nothing wrong in the above, changing it :-)

lib/fabrik/DirectorInstance.js
jagadish-kb: Good catch. Fixing this by logging the error after all retries fail. Also this is done in tap so it will not change the outcome of create operation.

lib/fabrik/DirectorManager.js
jagadish-kb: Done

lib/jobs/ScheduleManager.js
jagadish-kb: Done

lib/jobs/ScheduleManager.js
jagadish-kb: Please look at [BaseJob](https://github.com/jagadish-kb/service-fabrik-broker/blob/master/lib/jobs/BaseJob.js#L88). Yes, end user will see an error.

lib/controllers/ServiceFabrikAdminController.js
jagadish-kb: 1. Naming is always subjective. Will leave it as-is. [Also please look at method signature of MaintenanceManager.updateMaintenance - this makes it clear that progress is actually progress info.]
2. All the params passed to this method are needed. [JSHint would have errored otherwise]

sauravmndl: "2. All the params passed to this method are needed. [JSHint would have errored otherwise]"
For this I was suggesting updateMaintenace(req.body, req.user) just like the startMaintenance() method.
Yes of course the method signature is correct in current code.

lib/maintenance/MaintenanceManager.js
jagadish-kb: completing the circle :-)

lib/utils/index.js
jagadish-kb: Dont know :-)

test/acceptance/service-fabrik-api.instances.director.spec.js
```diff
@@ -1432,6 +1423,134 @@ describe('service-fabrik-api', function () {
...
+            .then(res => {
+              expect(res).to.have.status(503);
+              expect(res.body.message).to.eql(`${CONST.FEATURE.SCHEDULED_UPDATE} feature not enabled`);
+              config.mongodb.url = mongourl;
+              config.mongodb.provision = mongoprovision;
+              mocks.verify();
+            });
+        });
+
+        it('should return 400 - Badrequest on skipping mandatory params', function () {
```
sauravmndl: Badrequest camel case.

test/acceptance/service-fabrik-api.instances.director.spec.js
```diff
@@ -1432,6 +1423,134 @@ describe('service-fabrik-api', function () {
...
+            .send({})
+            .catch(err => err.response)
+            .then(res => {
+              expect(res).to.have.status(400);
+              expect(res.body).to.eql({});
+              mocks.verify();
+            });
+        });
+
+        it('should return 201 OK', function () {
```
sauravmndl: in description you can put the use case.

jagadish-kb: describe section mentions the use case. This is one of the conditions in that use case. [This means successfully able to schedule update]

PR #60, opened 2017-10-04
Update create check

muralisc:
The existing check does not check for username undefined.
More accurate check is to check for the username scope when the
bosh_director_name is provided.


muralisc:
https://github.com/SAP/service-fabrik-broker/commit/48a7d1f46338e29a553bfb7a5a4a09b9d064a982#commitcomment-24718411

Issue #61, opened 2017-10-05
Wrong quota calculation for similar plan names


subhankarc:
**Description**

Plans with name like v1.0-xsmall and v1.0small are treated as same SKU. Also, v1.0-dev-large and v1.0-large are treated as same SKU

**Steps to reproduce the issue:**
1. Keep quota of v1.0-xsmall as 1 and v1.0-small as 1
2. Create one v1.0-xsmall instance.
3. Now try to create one v1.0-small, it gives quota exceeded.

**If you are running service fabrik locally than Output of `node -v 
for node version and grep version package.json for service fabrik version`:**

```
(paste your output here)
node version: v7.1.0
service fabrik: 3.0.1
```

**Additional environment details (AWS, VirtualBox, OpenStack, etc.):**

kamath-prasad:
Issue fixed and delivered.

PR #62, opened 2017-10-05
Closes issue 61 : Wrong Quota calculation

subhankarc:
Closes bug #61 

1. Updated the retrival of SKU name calculation so that substrings are not considered.
2. Also added additional check for manager name.
3. Updated the unit tests and added new ones to validate the added checks.
4. Added new small plan in the settings.yml, and changed the plan names.

### DIFF
--- a/lib/quota/QuotaManager.js
+++ b/lib/quota/QuotaManager.js
@@ -31,17 +31,8 @@ class QuotaManager {
               let serviceName = _.find(catalog.plans, ['id', planId]).service.name;
               logger.debug(`Plan Name is ${planName}`);
               logger.debug(`Service Name is ${serviceName}`);
-              const skuName = _.join(_.drop(planName.split('-')), '-'); // assumption here is that service pla names are versioned, and the format is like <version>-<deployment-type>-<tshirt-size> and skuName will be only <deployment-type>-<tshirt-size>
-              logger.debug(`SKUName is ${skuName}`);
-              const planIdsWithSameSKU = [];
-              const service = _.find(catalog.services, ['name', serviceName]);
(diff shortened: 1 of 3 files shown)


lib/quota/QuotaManager.js
```diff
@@ -65,6 +56,27 @@ class QuotaManager {
     });
   }
 
+  getAllPlanIdsWithSameSKU(planName, serviceName, serviceCatalog) {
+    return Promise.try(() => {
+      const planManagerName = _.find(catalog.plans, ['name', planName]).manager.name;
+      const splittedPlanNameWithoutVersionInfo = _.drop(planName.split('-')); // assumption here is that service plan names are versioned, and the format is like <version>-{...}-<tshirt-size>
+      splittedPlanNameWithoutVersionInfo.unshift(''); // and skuName will be only -{...}-<tshirt-size>   
+      const skuName = _.join(splittedPlanNameWithoutVersionInfo, '-');
```
jagadish-kb: Above is correct, nothing wrong but hard to read and decipher. Instead can you try something like the below
```
const skuName = _.chain('planName')
.split('-')
.filter((idx, i) => i !== 0)
.join('-')
.value()
```
chains are used for scenarios like above where you are performing a list of operations in sequence.
Alternatively you can just get strip everything after first delimiter '-' as thats what you are getting
```
const firstIdx = planName.indexOf('-');
const skuName = planName.substring(planName.indexOf('-', firstIdx));
```

subhankarc: Adjusted the code accordingly, using the second method, as it is easier to read.

PR #64, opened 2017-10-10
Modified the code to avoid logging additional event to riemann

kamath-prasad:
Modified the code to avoid logging additional event to riemann when guids are not present and also switched off event logging for get backup related events


kamath-prasad:
@ketakig12 : Fixed the UT's and added new ones.

PR #67, opened 2017-10-11
Fix/issue lock deployment error message

### LINKED ISSUE #66: In case of backup already running error message does not have complete details
Currently if a backup is already running if another backup on the instance is triggered again, then the returned error message is just says 'Deployment ${deploymentName} __LOCKED__'. This information is insufficien.

Fix:  Return back the original message thrown from the broker which tells why the lock is put & by whom. 
Example message as thrown from broker:
'Deployment service-fabrik-0315-b9bf180e-1a67-48b6-9cad-32bd2e936849 __Locked__ by admin at Wed Oct 11 2017 04:09:38 GMT+0000 (UTC) for on-demand_backup'

jagadish-kb:
1. Closes #66 
2. JobScheduler will only start if there is no maintenance or if last maintenance was successful. If last maintenance had failed, then scheduler will not start till it is fixed.

### DIFF
--- a/lib/errors.js
+++ b/lib/errors.js
@@ -168,12 +168,14 @@ class UnprocessableEntity extends HttpClientError {
 exports.UnprocessableEntity = UnprocessableEntity;
 
 class DeploymentAlreadyLocked extends UnprocessableEntity {
-  constructor(deploymentName, lockInfo) {
-    let lockBy = '';
+  constructor(deploymentName, lockInfo, lockMessage) {
+    let message = `Deployment ${deploymentName} ${CONST.OPERATION_TYPE.LOCK}`;
     if (lockInfo) {
-      lockBy = `by ${lockInfo.username} at ${lockInfo.createdAt} for ${lockInfo.lockForOperation}`;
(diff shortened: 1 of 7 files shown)


lib/errors.js
```diff
@@ -168,12 +168,14 @@ class UnprocessableEntity extends HttpClientError {
 exports.UnprocessableEntity = UnprocessableEntity;
 
 class DeploymentAlreadyLocked extends UnprocessableEntity {
-  constructor(deploymentName, lockInfo) {
-    let lockBy = '';
+  constructor(deploymentName, lockInfo, lockMessage) {
+    let message = `Deployment ${deploymentName} ${CONST.OPERATION_TYPE.LOCK}`;
     if (lockInfo) {
```
sauravmndl: not using lockMessage in case of lockInfo is present? Is lockMessage undefined in this case ?

jagadish-kb: Yes. The order of precendence is:
1. LockInfo
2. LockMessage
If LockInfo is present then message is built from it. If its not present then whatever lock message is provided is used. If both are undefined then default message is set. (Line 172)

PR #71, opened 2017-10-16
Fix/issue lock deployment error message

### LINKED ISSUE #70: Backup completion message not shown at service instance level.
Backup/restore operations are performed via cf-update operation on the service instance and the last operation message reflected the same. However as part of redesign of backup & restore, this status message now does not show the backup completion message, rather shows a message that deployment is unlocked. This message is a generic message for unlock operation and should be provided only when specific message is not available. however in case of backup, a specific message should be provided similar to how it was done prior to redesign.

jagadish-kb:
Closes #70

### DIFF
--- a/lib/fabrik/FabrikStatusPoller.js
+++ b/lib/fabrik/FabrikStatusPoller.js
@@ -94,9 +94,13 @@ class FabrikStatusPoller {
                   }
                   if (instanceInfo.operationFinished) {
                     logger.info(`Backup complete for guid : ${instanceInfo.backup_guid} -  deployment : ${instanceInfo.deployment} `);
+
                     const unlockOperation = new ServiceFabrikOperation('unlock', {
                       instance_id: instanceInfo.instance_guid,
-                      isOperationSync: true
+                      isOperationSync: true,
+                      arguments: {
+                        description: _.get(operationResponse, 'description')
+                      }
                     });
                     return unlockOperation
                       .invoke()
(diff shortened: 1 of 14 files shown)


lib/fabrik/FabrikStatusPoller.js
```diff
@@ -96,7 +96,10 @@ class FabrikStatusPoller {
                     logger.info(`Backup complete for guid : ${instanceInfo.backup_guid} -  deployment : ${instanceInfo.deployment} `);
                     const unlockOperation = new ServiceFabrikOperation('unlock', {
                       instance_id: instanceInfo.instance_guid,
-                      isOperationSync: true
+                      isOperationSync: true,
+                      arguments: {
+                        description: _.get(operationResponse, 'description')
```
kamath-prasad: 1. As I understand the description would be returned by : 

https://github.com/SAP/service-fabrik-broker/blob/master/lib/fabrik/DirectorManager.js#L456-L475

Is this correct ? If yes , then the message values do not match with the one's below in the test correct ?

2. Also the test contains description like "Backup Operation in-progress" , "Backup operation abort in-progress" : will we ever have these statuses given that the description is passed to CF only in the update call in unlock which gets executed only if backup is either succeeded or failed ?

jagadish-kb: Tests are at Unit level and not an end to end scenario. DirectorManager is anyway stubbed and whatever the stubbed classes returns FabrikStatus poller returns the same. As such fabrikstatus poller really does not care what is the description. Whatever it gets it just passes same back. So no need to replicate exact same message in mock unit tests.

PR #78, opened 2017-10-18
Closes issue #74: Availability set configuration for Azure landscapes

### LINKED ISSUE #74: Availability set configuration for Azure landscapes
Currently service-fabrik-broker reads the `availability-zone` configuration from its manifest for all IaaS and sets it in the cloud-properties section of the spawned instance.
However, for Azure landscapes, `availability-zone` doesn't make a sense. Since, for Azure, the term used is `Availability Set`. This config has to be set by the services' manifest and the broker must read the proper config (availability-set) and set it in the cloud-properties section of the spawned instance.

pritishmishra:
Description: 
1. Modified the resource pools filtering to handle the new inputs of availability set from the manifest, 
2. Made this part iaas specific and driven from config, 
3. Added/modified UTs

Fixes #74

### DIFF
--- a/config/templates/blueprint-manifest.yml.ejs
+++ b/config/templates/blueprint-manifest.yml.ejs
@@ -1,4 +1,4 @@
-<%= spec.header.select({disk_pools: [p('disk_pool')], resource_pools: [p('resource_pool')]}) %>
+<%= spec.header.select({disk_pools: [p('disk_pool')], resource_pools: [{name: p('resource_pool')}]}) %>
 
 update:
   canaries: 0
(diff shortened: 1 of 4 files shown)


pritishmishra:
Implemented the review comments, please check.

subhankarc:
Thanks @jagadish-kb for mentioning it.

@pritishmishra Can you confirm if we need to change in the example ejs files?
 
https://github.com/SAP/service-fabrik-broker/blob/master/config/templates/

jagadish-kb:
Approving this PR, but should not be merged, unless changes in product component/sample manifests are done. Both should be merged at the same time.

pritishmishra:
Yes, @subhankarc, you are correct. This change was breaking for bosh-lite. I will make the relevant changes to the example manifest and broker code and push it.

pritishmishra:
Relevant changes for manifest ejs files have been made. Kindly review and merge.

PR #81, opened 2017-10-24
Initializing initialJobRunStatus to first jobrun status

jagadish-kb:
When backups are running in background the number of retry attempts are not restricted to 3. However when lastrunstatus is checked only last 3 records from job history are fetched & the initial job run status was picked up based on the attempt count. In case of backup runs, we initial job run might not be present in the last 3 run records & hence this scenario could lead to an error while fetching the last run status.
The fix to this issue is to initialise the initialJobRunStatus to the last entry from the fetched set of records. By doing so, we ensure this is always defined.

### DIFF
--- a/lib/jobs/ServiceInstanceUpdateJob.js
+++ b/lib/jobs/ServiceInstanceUpdateJob.js
@@ -204,14 +204,14 @@ class ServiceInstanceUpdateJob extends BaseJob {
         }
         const lastRunList = lastRunDetails.list;
         const count = lastRunList.length;
-        let initialJobRunStatus, idx;
-        for (idx = 0; idx < count; idx++) {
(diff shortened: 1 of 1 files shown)


lib/jobs/ServiceInstanceUpdateJob.js
```diff
@@ -204,14 +204,14 @@ class ServiceInstanceUpdateJob extends BaseJob {
         }
         const lastRunList = lastRunDetails.list;
         const count = lastRunList.length;
-        let initialJobRunStatus, idx;
-        for (idx = 0; idx < count; idx++) {
+        let initialJobRunStatus = lastRunList[count - 1];
```
sauravmndl: For the very first time what is the content  of 'lastRunList' ?

PR #82, opened 2017-10-25
Fix error while fetching schedule update last run status

jagadish-kb:
was logging the entire results from DB which the logger could not handle and let to recursive calls. Fixed it by picking & logging only the require attribs.

### DIFF
--- a/lib/jobs/ServiceInstanceUpdateJob.js
+++ b/lib/jobs/ServiceInstanceUpdateJob.js
@@ -204,14 +204,14 @@ class ServiceInstanceUpdateJob extends BaseJob {
         }
         const lastRunList = lastRunDetails.list;
         const count = lastRunList.length;
-        let initialJobRunStatus, idx;
-        for (idx = 0; idx < count; idx++) {
+        let initialJobRunStatus = lastRunList[count - 1];
+        for (let idx = 0; idx < count; idx++) {
           if (lastRunList[idx].data.attempt === 1) {
             initialJobRunStatus = lastRunList[idx];
             break;
           }
         }
-        logger.debug('Initial run status - ', initialJobRunStatus);
+        logger.info('Initial run status - ', initialJobRunStatus);
         const diffBeforeUpdate = _.get(initialJobRunStatus, 'response.diff', null) || _.get(initialJobRunStatus, 'response.jobStatus.diff', null);
         const diffAfterUpdate = count === 1 ? 'TBD' : (_.get(lastRunList[0], 'response.diff') ||
           _.get(lastRunList[0], 'response.jobStatus.diff', null));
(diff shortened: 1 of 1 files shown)


lib/jobs/ServiceInstanceUpdateJob.js
```diff
@@ -198,10 +198,11 @@ class ServiceInstanceUpdateJob extends BaseJob {
           //Fetch last 3 run history records and figure out the status of the update operation.
         })
       .then(lastRunDetails => {
-        logger.info('LastRun details retrieved - ', lastRunDetails);
+        logger.info('Total last runs # ', _.get(lastRunDetails, 'totalRecordCount', 0));
```
sauravmndl: Specification of last run details of job SI update would be better.

lib/jobs/ServiceInstanceUpdateJob.js
```diff
@@ -211,7 +212,8 @@ class ServiceInstanceUpdateJob extends BaseJob {
             break;
           }
         }
-        logger.info('Initial run status - ', initialJobRunStatus);
+        logger.info('Initial run status - ', _.pick(initialJobRunStatus, 'data', 'response', 'statusCode', 'statusMessage', 'interval'));
+        logger.info('Last run status - ', _.pick(lastRunList[0], 'data', 'response', 'statusCode', 'statusMessage', 'interval'));
```
sauravmndl: In above two I think not mandatory to add SI update job  as data already contains the info.

lib/jobs/ServiceInstanceUpdateJob.js
jagadish-kb: This was causing problems earlier. `lastRunDetails` array is mongo db objects of lastrundetails & they are huge. They not only contain the data representation but it also prints various other attribs of mongo wrapper classes, which are really not required. What really interests us is the first run response and last run response which are logged appropriately.

lib/jobs/ServiceInstanceUpdateJob.js
jagadish-kb: Did not understand the above comment. What in the above logs are you referring as SI Update Job?

lib/jobs/ServiceInstanceUpdateJob.js
sauravmndl: I meant in logger the job name: ServiceInstanceUpdateJob and instance id would be better. As it will keep a reference the id against the count. No other details.

lib/jobs/ServiceInstanceUpdateJob.js
sauravmndl: Ignore. I meant the above logger is perfect :)

PR #83, opened 2017-10-26
Fix error while fetching schedule update last run status

jagadish-kb:
was logging the entire results from DB which the logger could not handle and let to recursive calls. Fixed it by picking & logging only the require attribs.

PS: Same changes are already merged in master.


vinaybheri:
Tested in dev with this code change in master and it worked fine.
Hence merging this PR.

PR #86, opened 2017-10-27
Configured reschedule delay and max_attempt at job level

jagadish-kb:
Currently if an updaate job fails because of  a backup job running or
vice versa, the failed job is rescheduled after a configured
reschedule_delay which is at app level. While this works perfectly,
this will make testing these in an automated way difficult as the
validation must wait to longer periods since the default value
of this attrib is 20 minutes. Hence to overcome this, this param
along with max_attempts can be configured at job level when the
job is scheduled (both backup/update). Please note these overrides
are to be done only for purposes of testing and should not be
(mis)used for any other scenarios.

### DIFF
--- a/lib/controllers/ServiceFabrikApiController.js
+++ b/lib/controllers/ServiceFabrikApiController.js
@@ -490,12 +490,15 @@ class ServiceFabrikApiController extends BaseController {
           .then(instance => _.set(req, 'entity.name', instance.entity.name));
       })
       .then(() => req.manager.findDeploymentNameByInstanceId(req.params.instance_id))
-      .then(deploymentName => ({
-        instance_id: req.params.instance_id,
-        instance_name: req.entity.name,
-        deployment_name: deploymentName,
-        timeZone: req.body.timeZone
-      }))
+      .then(deploymentName => _
+        .chain({
+          instance_id: req.params.instance_id,
+          instance_name: req.entity.name,
+          deployment_name: deploymentName
+        })
+        .assign(req.body)
+        .value()
+      )
       .then((jobData) => ScheduleManager
         .schedule(req.params.instance_id,
           CONST.JOB.SERVICE_INSTANCE_UPDATE,
--- a/lib/controllers/ServiceFabrikApiController.js
+++ b/lib/controllers/ServiceFabrikApiController.js
@@ -496,7 +496,7 @@ class ServiceFabrikApiController extends BaseController {
           instance_name: req.entity.name,
           deployment_name: deploymentName
         })
-        .assign(req.body)
+        .assign(_.omit(req.body, 'repeatInterval'))
         .value()
       )
       .then((jobData) => ScheduleManager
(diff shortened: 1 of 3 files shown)


lib/controllers/ServiceFabrikApiController.js
```diff
@@ -490,12 +490,15 @@ class ServiceFabrikApiController extends BaseController {
           .then(instance => _.set(req, 'entity.name', instance.entity.name));
       })
       .then(() => req.manager.findDeploymentNameByInstanceId(req.params.instance_id))
-      .then(deploymentName => ({
-        instance_id: req.params.instance_id,
-        instance_name: req.entity.name,
-        deployment_name: deploymentName,
-        timeZone: req.body.timeZone
```
kamath-prasad: why is timeZone removed ? We dont use it ?

lib/jobs/ScheduleBackupJob.js
```diff
@@ -139,15 +139,15 @@ class ScheduleBackupJob extends BaseJob {
     return Promise.try(() => {
       const jobData = _.cloneDeep(jobOptions);
       jobData.attempt = jobData.attempt + 1;
-      if (jobData.attempt > config.scheduler.jobs.scheduled_backup.max_attempts) {
+      if (jobData.attempt > _.get(jobData, 'max_attempts', config.scheduler.jobs.scheduled_backup.max_attempts)) {
         logger.error(`Scheduled backup for instance  ${jobData.instance_id} has exceeded max configured attempts : ${config.scheduler.jobs.scheduled_backup.max_attempts} - ${jobData.attempt}}. Initial attempt was done @: ${jobData.firstAttemptAt}.`);
```
kamath-prasad: can you change the log statement also to print the value correctly for  for max_attempts and reschedule_delay ?

lib/jobs/ServiceInstanceUpdateJob.js
```diff
@@ -150,7 +150,7 @@ class ServiceInstanceUpdateJob extends BaseJob {
       const jobData = _.cloneDeep(instanceDetails);
       jobData.attempt = jobData.attempt + 1;
       if (trackAttempts) {
-        if (jobData.attempt > config.scheduler.jobs.service_instance_update.max_attempts) {
+        if (jobData.attempt > _.get(instanceDetails, 'max_attempts', config.scheduler.jobs.service_instance_update.max_attempts)) {
           logger.error(`Auto udpate for instance ${instanceDetails.instance_id}  has exceeded max configured attempts : ${config.scheduler.jobs.service_instance_update.max_attempts}}`);
```
kamath-prasad: same as above , print correct value in log for max_attempts and reschedule_delay.

lib/controllers/ServiceFabrikApiController.js
jagadish-kb: timezone is part of res.body which along with any other attribs coming in request is assisgned as part of job data. Look at inserted code segment.

lib/jobs/ScheduleBackupJob.js
jagadish-kb: Done!

lib/controllers/ServiceFabrikApiController.js
kamath-prasad: Ok , got it .

PR #88, opened 2017-10-30
Modified the code avoid logging riemann events if the input request is bad

kamath-prasad:
Closes https://github.com/SAP/service-fabrik-broker/issues/87

### DIFF
--- a/lib/constants.js
+++ b/lib/constants.js
@@ -37,8 +37,8 @@ module.exports = Object.freeze({
   },
   HTTP_STATUS_CODE: {
     CONFLICT: 409,
-    UNPROCESSABLE_ENTITY: 422 ,
-    BAD_REQUEST : 400
+    UNPROCESSABLE_ENTITY: 422,
+    BAD_REQUEST: 400
   },
   JOB_NAME_ATTRIB: '_n_a_m_e_',
   JOB: {
--- a/lib/constants.js
+++ b/lib/constants.js
@@ -38,7 +38,7 @@ module.exports = Object.freeze({
   HTTP_STATUS_CODE: {
     CONFLICT: 409,
     UNPROCESSABLE_ENTITY: 422,
-    BAD_REQUEST: 400
+    BAD_REQUEST: '400'
   },
   JOB_NAME_ATTRIB: '_n_a_m_e_',
   JOB: {
--- a/lib/constants.js
+++ b/lib/constants.js
@@ -38,7 +38,7 @@ module.exports = Object.freeze({
   HTTP_STATUS_CODE: {
     CONFLICT: 409,
     UNPROCESSABLE_ENTITY: 422,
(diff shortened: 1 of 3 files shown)


lib/constants.js
```diff
@@ -37,7 +37,8 @@ module.exports = Object.freeze({
   },
   HTTP_STATUS_CODE: {
     CONFLICT: 409,
-    UNPROCESSABLE_ENTITY: 422
+    UNPROCESSABLE_ENTITY: 422 ,
+    BAD_REQUEST : 400
```
pritishmishra: A minor edit, can we follow the same pattern for line number 40, 41 as in line number 39.

lib/utils/EventLogRiemannClient.js
```diff
@@ -52,7 +52,7 @@ class EventLogRiemannClient {
 
   handleEvent(message, data) {
     try {
-      if (data.event) {
+      if (data.event && _.get(data, 'event.response.status') != CONST.HTTP_STATUS_CODE.BAD_REQUEST) {
```
pritishmishra: Is it not the same as `data.event.response.status`? 
But may be _.get is better way of writing code??

test/utils.EventLogRiemannClient.spec.js
```diff
@@ -185,7 +185,54 @@ describe('utils', function () {
         expect(secondResponse).to.be.an('object');
         expect(secondResponse).to.eql(expectedSecondResultObject);
       });
+
+      it('should not log any event in case of bad request', function () {
+        const event = {
+          host: 'INLN50932351A',
```
pritishmishra: Can we add a dummy hostname, e.g. hostname, instead of the equipment number? :)

lib/utils/EventLogRiemannClient.js
kamath-prasad: Actually _.get handles all null cases implicitly . Hence no additional handling required by us.

test/utils.EventLogRiemannClient.spec.js
kamath-prasad: Yes , ok :)

lib/utils/EventLogRiemannClient.js
pritishmishra: ok.. understood.

lib/constants.js
```diff
@@ -37,7 +37,8 @@ module.exports = Object.freeze({
   },
   HTTP_STATUS_CODE: {
     CONFLICT: 409,
-    UNPROCESSABLE_ENTITY: 422
+    UNPROCESSABLE_ENTITY: 422,
+    BAD_REQUEST: '400'
```
sauravmndl: is it okay we are putting the 400 status code as string, while others are number ?

kamath-prasad: Changed to int.

PR #90, opened 2017-11-14
Implementing the self-healing property of cache

pritishmishra:


lib/bosh/BoshDirectorClient.js
```diff
@@ -169,6 +183,16 @@ class BoshDirectorClient extends HttpClient {
     return this.request(requestDetails, expectedStatusCode);
   }
 
+  checkAndUpdateCache(deploymentName, config) {
```
jagadish-kb: No need to define this new method. Rather put this implementation into the method [here](https://github.com/SAP/service-fabrik-broker/blob/master/lib/bosh/BoshDirectorClient.js#L141-L153). Current implementation in getDirectorConfig reloads entire cache when there is a miss. Rather it must only update the cache entry for the deployment name which is not found.

lib/bosh/BoshDirectorClient.js
```diff
@@ -156,7 +156,21 @@ class BoshDirectorClient extends HttpClient {
 
   makeRequest(requestDetails, expectedStatusCode, deploymentName) {
     return this.getDirectorConfig(deploymentName)
-      .then(directorConfig => this.makeRequestWithConfig(requestDetails, expectedStatusCode, directorConfig));
+      .then(directorConfig => this.makeRequestWithConfig(requestDetails, expectedStatusCode, directorConfig))
+      .catch(NotFound, (err) => {
+        logger.debug('Going to delete the entry from cache for the deployment:', deploymentName);
+        this.deleteCacheEntry(deploymentName);
+        return Promise
+          .map(config.directors, (directorConfig) => this.checkAndUpdateCache(deploymentName, directorConfig)
```
jagadish-kb: Dont invoke this new method, rather invoke `makeRequest` itself (call self) and pass an additional param named `attempt` with value 1. So code should be like the below 
```
makeRequest(requestDetails, expectedStatusCode, deploymentName, attempt){
...
...
(NotFound, (err) => {
      logger.debug('Going to delete the entry from cache for the deployment:', deploymentName);
      this.deleteCacheEntry(deploymentName);
      if(attempt !== undefined){
            throw err;
       }
       return this.makeRequest(requestDetails, expectedStatusCode, deploymentName, 1);
    });
``` 

The idea here is if deployment is not found, you invoke it again by clearing cache. Also  `getDirectorConfig` method must be enhanced to update cache at deployment level rather than removing all contents of cache and reloading entire cache contents. [Comments provided for the below method should give details on how to fix `getDirectorConfig`]

lib/bosh/BoshDirectorClient.js
```diff
@@ -169,6 +183,16 @@ class BoshDirectorClient extends HttpClient {
     return this.request(requestDetails, expectedStatusCode);
   }
 
+  checkAndUpdateCache(deploymentName, config) {
+    return this
```
jagadish-kb: The looping should not be from caller, but from within here. Loop through each of the director configs and get deployment. Also use `Promise.any` so that you fire get deployment calls in parallel to all directors and whoever returns first will update the cache.

pritishmishra:
Closing this PR. Raised a separate PR.

PR #95, opened 2017-11-15
Add tags block in deployment manifest

amitmalav:
This change adds tags block in manifest of existing as well as new
deployments. If applied, Resources will get tagged with space and org guids at
iaas level. For existing deployments manifest changes are applied after update.
tags:
    space_guid: <space_guid>
    organization_guid: <org_guid>


lib/cf/CloudControllerClient.js
```diff
@@ -227,6 +227,14 @@ class CloudControllerClient extends HttpClient {
   getSpaces(options) {
     return this.createSpaceStream(options).all();
   }
+  getSpaceGuid(instance_guid, options) {
```
jagadish-kb: I would not have this wrapper method but there is nothing wrong with it either. Just that you seem to have a wrapper method in basecontroller as well, so generally wrapping a wrapper method does not seem right. You could have directly done the contents of this method in BaseController. Or even better remove the wrapper functions from BaseController and let the admin controller directly refer this method. Either ways look at how wrapper functions are created. I feel it might be better to remove the wrapper methods in BaseController, so that in future all other modules can then leverage the same.

lib/controllers/BaseController.js
```diff
@@ -109,6 +108,22 @@ class BaseController {
   getPlan(plan_id) {
     return catalog.getPlan(plan_id);
   }
+  getSpaceGuid(deploymentName) {
```
jagadish-kb: Looks like this is unused ?

lib/controllers/BaseController.js
```diff
@@ -109,6 +108,22 @@ class BaseController {
...
+    return this.cloudController.getSpaceGuid(this.getInstanceId(deploymentName));
+  }
+  getOrgGuid(deploymentName) {
```
jagadish-kb: Looks like this is unused

lib/controllers/BaseController.js
```diff
@@ -109,6 +108,22 @@ class BaseController {
...
+    return this.cloudController.getOrgGuid(this.getInstanceId(deploymentName));
+  }
+  createOpts(deploymentName) {
```
jagadish-kb: rename this method to `getOrgAndSpaceGuid` its better than having generic names like this as it helps with readability when one just reads this method from the subclass. Else from subclass he needs to jump into the method definition to know what this createOpts is all about.

lib/controllers/BaseController.js
```diff
@@ -109,6 +108,22 @@ class BaseController {
...
+    return Promise
+      .try(() => this.cloudController.getSpaceGuid(this.getInstanceId(deploymentName))
```
jagadish-kb: No need to have this Promise.try() just return the promise chain.

lib/errors.js
```diff
@@ -208,6 +208,14 @@ class ServiceInstanceNotFound extends NotFound {
 }
 exports.ServiceInstanceNotFound = ServiceInstanceNotFound;
 
+
+class SpaceOrOrgNotFound extends NotFound {
```
jagadish-kb: where is this used? Looks like an unused error. Also even if required, should rather use the error `NotFound` with the required message. Having such a specific NotFound errors not required.

lib/jobs/ServiceInstanceUpdateJob.js
```diff
@@ -118,12 +118,25 @@ class ServiceInstanceUpdateJob extends BaseJob {
       });
   }
 
-  static getOutdatedDiff(deploymentName, plan) {
+  static createOpts(instanceId) {
```
jagadish-kb: If required you can remove the wrapper method `createOpts` and move it to cloudController with the name `getOrgAndSpaceGuid`, so that way you can reuse the same from controller as well as jobs. We move reusable to higher levels of common points based on the places from where it can be consumed. So looking at this methods current usage, to avoid dupe, its best we move this method to cloudController directly.

lib/cf/CloudControllerClient.js
```diff
@@ -227,6 +227,14 @@ class CloudControllerClient extends HttpClient {
...
+    return this.getServiceInstance(instance_guid, options)
+      .then(instance => instance.entity.space_guid);
+  }
+  getOrgGuid(space_guid, options) {
+    return this.getResource(`/v2/spaces/${space_guid}`, options)
+      .then(org => org.entity.organization_guid);
```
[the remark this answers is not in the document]
amitmalav: Agreed, will make corresponding changes.

lib/fabrik/DirectorManager.js
```diff
@@ -322,36 +322,40 @@ class DirectorManager extends BaseManager {
...
-      }));
+      })));
   }
 
   generateManifest(deploymentName, opts) {
     const index = opts.network_index || this.getNetworkSegmentIndex(deploymentName);
     const networks = this.getNetworks(index);
     const allRequiredNetworks = _.union(networks.dynamic, networks.all.filter(net => _.startsWith(net.name, this.networkName)));
-
+    const tags = {
```
subhankarc: Can we put a null check that if the opts is null/empty, it does not create the tag block.

jagadish-kb: +1

lib/bosh/manifest/Header.js
```diff
@@ -14,8 +14,8 @@ class Header {
     this.disk_pools = header.disk_pools;
     this.resource_pools = header.resource_pools;
     this.networks = header.networks;
+    this.tags = header.tags;
```
jagadish-kb: add this line only if tags is defined. This will take care of not having tags block when its undefined.
```
if(header.tags) {
this.tags = header.tags;
}
```

lib/fabrik/DirectorManager.js
jagadish-kb: on second thought, tags should never be null right?

amitmalav: We will have to check if opts is empty (it only has previous_manifest) ie. does not have both org and space guid. In that case we can keep tags as empty.

subhankarc: opts would be null in case of non-cf cross consumption.

But I am ok with taking it up later.

jagadish-kb: @amitmalav if in previous manifest this was empty, it will be updated post pushing this feature. So this will never be empty once this feature goes live.
@subhankarc even in case of non-cf we would have an id equivalent to org/space id? We might not have two ids, but we could still have one ID. so when we normalize it,  the tag might not be empty even for cross consumption.

amitmalav: Yes, Agreed. In current implementation, yes, opts will never be empty. But I think what @subhankarc was speculating, was for generic calls to generateManifest() without space_guid and org_guid info present in opts, then tags block will have space and org guids as null. I think we can take it up later when we will be making changes to support non-cf cross-consumption.

kamath-prasad:
As there are 2 approvals , merging the PR.

PR #98, opened 2017-11-20
 fixing issues with sporadic failures in UT

jagadish-kb:
1. Caching of deployment names in director client lead to some inconsistencies on the state of cache. Now ensured that before each test run the cache state is the same and if required individual tests can override the default cache state.

2. If there is an error loading cache (i.e. on errors when fetching deployments), then bubble up the error than eating it. Bubbling up the error makes it quite evident where the failures are.

3. If bosh config is being duplicated for a deployment name when fetched from cache, it must throw an error even in this case.

4. During creation of deployment an unnecessary call to BOSH was being fired before picking up the bosh config. Avoided this unnecessary call to bosh during create deployment.


lib/bosh/BoshDirectorClient.js
```diff
@@ -31,15 +31,16 @@ class BoshDirectorClient extends HttpClient {
     this.oobDirectorConfigs = BoshDirectorClient.getOobDirectorConfigs();
     this.cache = {};
     this.cacheLoadInProgress = false;
-    this.populateCache();
+    this.isReady = this.populateCache();
```
kamath-prasad: `isReady` is a boolean here ?

populateCache() does not return boolean , correct ?

lib/bosh/BoshDirectorClient.js
```diff
@@ -338,29 +339,24 @@ class BoshDirectorClient extends HttpClient {
...
-          logger.warn(`Deploying through ${boshDirectorName}`);
-          return this
-            .getDirectorConfig(deploymentName)
-            .then(config => {
-              if (config) {
-                throw new BadRequest(`Cannot create an existing deployment ${deploymentName}`);
-              }
-            })
-            .catch(NotFound, () => this.getConfigByName(boshDirectorName));
+        if (action === CONST.OPERATION_TYPE.CREATE) {
```
kamath-prasad: 👍

lib/bosh/BoshDirectorClient.js
jagadish-kb: its a promise.

kamath-prasad: Ok , should we rename the variable then ? 

As I understand now , it contains the reference to cache wrapped in a promise , correct ?

Issue #99, opened 2017-11-21
Broker sends empty response in case of request timeout


pritishmishra:
Currently, the broker sends an empty HTTP Response code if any request sent to the broker times out. The time out for the broker is 120 seconds. If a request exceeds this time, an empty response with no body and no response code is received.
More info here: https://github.com/expressjs/morgan/issues/121
`“This happens when you do not write a response and allow the app to just hang. The reason you get the dashes is because since you didn't write a response, there is no response code, there is no response time, there is no response size, etc. Basically you are seeing all the properties of the response as dashes (i.e. unknown) and the properties of the request populated in that log.”`

This has been observed to cause duplicate create calls from cf, as it expects a proper response from the broker, even if it times out.
More info: https://github.com/openservicebrokerapi/servicebroker/blob/v2.13/spec.md#provisioning

Solution:
We must do something like this from our end in case of timeouts:
https://expressjs.com/en/resources/middleware/timeout.html to explicitly flag HTTP status of timeout.

mowi22:
I also wanted to leave this here for anyone that may have still gotten empty or pending responses from express. If you have not hung up the connection before sending the response. I was making this mistake on a call and it resulted in -- ms -- from morgan. It had nothing to do with Morgan though. Here is a similar SO about it. Hope it helps someone. 
https://stackoverflow.com/questions/19397161/node-express-pending-request

LuisRuizUribeWicho12:
hola, ya hice la instalacion del connect-timeout y me sigue dando el mismo error, lo estoy trabajando con nodejs y postman al igual que con crome, te adjunto el codigo con el que estoy trabajando

const express = require('express');
const app = express();
const http = require('http');
const server = http.createServer(app);
const logger = require('morgan');
const cors = require('cors');


const port = process.env.PORT || 3000;

app.use(logger('dev'));
app.use(express.json);
app.use(express.urlencoded({
    extended: true
}));
app.use(cors());

app.disable('x-powered-by');


app.set('port',port);

server.listen(3000, 'aqui estoy trabando con la ip de mi computadora' || 'localhost', function() {
    console.log('Aplicacion de nodeJS '+ process.pid + ' Iniciada...........');
});

app.get('/', (req, res) => {
    res.send('Ruta raiz del backend');
});     

//ERROR HANDLER

app.use((err, req, res, next) => {
    console.log(err);
    res.status(err.status || 500).send(err.stack);
});

[...]

PR #100, opened 2017-11-21
Closes issue #99: Broker sends empty response in case of request timeout

pritishmishra:
Handling timeout and sending proper HTTP response

### DIFF
--- a/apps.js
+++ b/apps.js
@@ -12,6 +12,8 @@ const config = lib.config;
 const logger = lib.logger;
 const routes = lib.routes;
 const middleware = lib.middleware;
+const connectTimeout = require('connect-timeout');
+const CONST = require('./lib/constants');
 
 // internal app
 const internal = createApp('internal', app => {
@@ -76,6 +78,8 @@ function createApp(type, addRoutes) {
     stream: logger.stream
   }));
   app.use(middleware.requireHttps(cfg));
+  app.use(connectTimeout(CONST.HTTP_TIMEOUT.TIMEOUT_DURATION));
+  app.use(middleware.timeout());
   app.use(express.static(path.join(__dirname, 'public')));
   app.use(bodyParser.urlencoded({
     extended: true
--- a/apps.js
+++ b/apps.js
@@ -13,7 +13,6 @@ const logger = lib.logger;
 const routes = lib.routes;
 const middleware = lib.middleware;
 const connectTimeout = require('connect-timeout');
-const CONST = require('./lib/constants');
 
 // internal app
 const internal = createApp('internal', app => {
@@ -78,8 +77,9 @@ function createApp(type, addRoutes) {
     stream: logger.stream
   }));
   app.use(middleware.requireHttps(cfg));
-  app.use(connectTimeout(CONST.HTTP_TIMEOUT.TIMEOUT_DURATION));
-  app.use(middleware.timeout());
+  app.use(connectTimeout(config.http_timeout, {
+    respond: true
+  }));
   app.use(express.static(path.join(__dirname, 'public')));
   app.use(bodyParser.urlencoded({
     extended: true
--- a/lib/constants.js
+++ b/lib/constants.js
@@ -40,9 +40,6 @@ module.exports = Object.freeze({
     UNPROCESSABLE_ENTITY: 422,
     BAD_REQUEST: 400
   },
-  HTTP_TIMEOUT: {
-    TIMEOUT_DURATION: '120s'
-  },
   JOB_NAME_ATTRIB: '_n_a_m_e_',
   JOB: {
     //Define names of scheduled JOBS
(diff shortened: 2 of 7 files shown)


lib/middleware.js
```diff
@@ -48,6 +49,15 @@ exports.methodNotAllowed = function (allow) {
   };
 };
 
+exports.timeout = function () {
```
kamath-prasad: do we want to add any log statements here ?

lib/constants.js
```diff
@@ -40,6 +40,9 @@ module.exports = Object.freeze({
     UNPROCESSABLE_ENTITY: 422,
     BAD_REQUEST: 400
   },
+  HTTP_TIMEOUT: {
```
jagadish-kb: Rename this to `HTTP_REQUEST` as it then makes it generic and in future you can house other props under it.

lib/middleware.js
```diff
@@ -48,6 +49,15 @@ exports.methodNotAllowed = function (allow) {
...
+  return function (req, res, next) {
+    if (!req.timedout) {
+      return next();
+    }
+    next(new GatewayTimeout('Timeout exceeded for the request'));
```
jagadish-kb: add a log here.

lib/middleware.js
jagadish-kb: +1 for log but location, i have pointed in the below comment. No need to log it for each request only on timeout it can be logged.

jagadish-kb: Agree. We should have some UT & possibly even Acceptance test. Using `sinon.FakeTimers` we can fake the elapsed time and probably catch it in UT and even some acceptance test. Also, did we test the actual bug and did CC stop firing in additional requests? In other words have we verified the actual duplicate deployment scenario.

apps.js
```diff
@@ -76,6 +78,8 @@ function createApp(type, addRoutes) {
     stream: logger.stream
   }));
   app.use(middleware.requireHttps(cfg));
+  app.use(connectTimeout(CONST.HTTP_TIMEOUT.TIMEOUT_DURATION));
```
jagadish-kb: pass second parameter options with `respond` set to true. By setting it, you dont need the next middleware and can move the code that is in middleware.timeout to the global error handler in middleware.

apps.js
```diff
@@ -76,6 +78,8 @@ function createApp(type, addRoutes) {
...
+  app.use(middleware.timeout());
```
jagadish-kb: Not required. Do the timeout required processing in middleware.error.  By default the connect-timeout module returns HTTP Status 503 - which means service is unavailable temporarily (due to load or maintenance). So it is debatable whether we want to add a new HTTP error GatewayTimeout or use the default one from timeout module.

lib/constants.js
pritishmishra: ok..

lib/middleware.js
pritishmishra: Ok.. will work on adding an UT/AT for this..

pritishmishra:
Test results:
1. I had created a dummy API endpoint which does nothing other than return a message after >2 mins (e.g. 5 mins). Upon hitting this endpoint with this change, it gives the proper error of Gateway Timeout error. Without this endpoint, it just shows 'empty response from the server.'
2. I had manipulated the CC call in such a way that it doesn't send a response at all. Upon firing a create call, without this change, the CC call came for 2 times, the 2nd one failed saying such a deployment name exists. With this change, only one call was made which timed out with the Gateway Timeout error.

lib/middleware.js
pritishmishra: For the part of testing if this solves the duplicate deployment scenario, please refer to the Test results comment below. (Point 2)

lib/middleware.js
pritishmishra: sure..

apps.js
pritishmishra: yes.. we can use the default error sent by using `respond` parameter. Need to test once, if it works the same way as Gateway Timeout error. Should work either way.

pritishmishra:
From the queries mentioned above, I believe query 1 can be answered from the test results posted above. (Point 1). Since, in this case, both the frameworks were in place, still we got the error from connect-timeout module. Do we need to be more safe and reduce the timeout by some buffer for connect-timeout, so that we don't accidentally even run into the null response?
Query 2 hasn't been tested. I will test and report back.

subhankarc:
Hi @pritishmishra ,

So CC sends back the same request again if we send empty response ?

Can you please link the issue raise on CC and also update that issue with your findings ?

kamath-prasad:
@subhankarc  : already raised - https://github.com/cloudfoundry/cloud_controller_ng/issues/975

pritishmishra:
Tested successfully the latest changes as per review comments.

pritishmishra:
Adding the config file changes for service-fabrik-boshrelease which should be merged along with this PR. https://github.com/SAP/service-fabrik-boshrelease/pull/25

PR #101, opened 2017-11-22
Add support to skip logging requests with client errors to riemann

kamath-prasad:
Added support in broker which would skip sending events to riemann if the incoming requests to broker failed because of client error .

https://en.wikipedia.org/wiki/List_of_HTTP_status_codes#4xx_Client_errors.

### DIFF
--- a/lib/utils/EventLogRiemannClient.js
+++ b/lib/utils/EventLogRiemannClient.js
@@ -52,7 +52,7 @@ class EventLogRiemannClient {
 
   handleEvent(message, data) {
     try {
-      if (data.event && _.get(data, 'event.response.status') !== CONST.HTTP_STATUS_CODE.BAD_REQUEST) {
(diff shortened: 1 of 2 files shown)


lib/utils/EventLogRiemannClient.js
```diff
@@ -52,7 +52,7 @@ class EventLogRiemannClient {
 
   handleEvent(message, data) {
     try {
-      if (data.event && _.get(data, 'event.response.status') !== CONST.HTTP_STATUS_CODE.BAD_REQUEST) {
+      if (data.event && !(_.get(config, 'riemann.skip_events_with_client_errors', true) && this.isEventWithClientError(_.get(data, 'event.response.status')))) {
```
jagadish-kb: for the sake of defining constants and having a sample. Can you please add this config param `skip_events_with_client_errors` in settings.yml? Also from readability perspective can this be renamed to `riemann.skip_http_status_40x` - client errors does not immediately strike a bell.

kamath-prasad: Yes , agree . Will make the changes.

PR #106, opened 2017-11-28
Add instance specific jobData in scheduleBackup

amitmalav:
Solves issue #108 
Added following data in jobData
 * instance name
 * plan name
 * space name
 * org guid and name
This information will be referred by BackupReport Manager
Changed 'Organisation' to 'Organization' in mocks

### DIFF
--- a/lib/cf/CloudControllerClient.js
+++ b/lib/cf/CloudControllerClient.js
@@ -70,6 +70,10 @@ class CloudControllerClient extends HttpClient {
     return this.getResource(`/v2/service_instances/${guid}`, options);
   }
 
+  getService(guid, options) {
+    return this.getResource(`/v2/services/${guid}`, options);
+  }
+
   updateServiceInstance(instance_id, options) {
     options = options || {};
     const auth = options.auth || {};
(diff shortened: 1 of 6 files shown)


lib/cf/CloudControllerClient.js
```diff
@@ -70,6 +70,10 @@ class CloudControllerClient extends HttpClient {
     return this.getResource(`/v2/service_instances/${guid}`, options);
   }
 
+  getService(guid, options) {
```
jagadish-kb: Method not required. All the service & plan level information for a guid are within service fabrik itself. Refer methods in `catalog.js`

lib/cf/CloudControllerClient.js
```diff
@@ -227,19 +231,54 @@ class CloudControllerClient extends HttpClient {
...
+      .then(body => this.getSpace(body.entity.space_guid)
+        .then(space => ({
           space_guid: body.entity.space_guid,
-          organization_guid: org.entity.organization_guid
+          organization_guid: space.entity.organization_guid
         }))
       );
   }
 
+  getServiceInstanceDetails(instance_guid) {
```
jagadish-kb: Can this method also take additional `options` parameter? This param can have `space_guid` as well provided in input. By the time request reaches all the respective handlers, `space_guid` will be [populated](https://github.com/SAP/service-fabrik-broker/blob/master/lib/controllers/ServiceFabrikApiController.js#L122) in request object. So no need to fetch space guid via a seperate call and it can avoid one additional call to CC. 
Also, no need to get Service & Plan details from CC. It can be retrieved from `catalog.js`.

lib/cf/CloudControllerClient.js
```diff
@@ -227,19 +231,54 @@ class CloudControllerClient extends HttpClient {
...
+    return this.getServiceInstance(instance_guid)
+      .then(instance => {
+        return Promise
+          .all([
+            this.getService(instance.entity.service_guid),
```
jagadish-kb: Replace with `catalog.getService`

lib/cf/CloudControllerClient.js
```diff
@@ -227,19 +231,54 @@ class CloudControllerClient extends HttpClient {
...
+            this.getServicePlan(instance.entity.service_plan_guid),
```
jagadish-kb: replace with `catalog.getPlan`

lib/cf/CloudControllerClient.js
```diff
@@ -227,19 +231,54 @@ class CloudControllerClient extends HttpClient {
...
+            this.getSpace(instance.entity.space_guid)
+          ])
+          .spread((service, plan, space) => {
+            return _
+              .chain({})
+              .set('name', instance.entity.name)
+              .set('label', service.entity.label)
```
jagadish-kb: change this attrib to `service_name` . We have never used `label` anywhere in code base, so it becomes confusing to understand what a label means just looking at job data.

PR #111, opened 2017-12-04
Virtual host manager to allow creation of virtual hosts in dedicated rabbitmq instances

pranjaljain:
Issue: https://github.com/cloudfoundry-incubator/service-fabrik-broker/issues/122

Raising this PR to get initial review comments on the new Virtual Host manager, implemented to allow creation of virtual hosts in dedicated rabbitmq instances.

To create an instance of type virtual_host, the user will first have to create a dedicated rabbitmq instance and then will have to pass the name as a parameter.

We will write unit tests and will follow other requirements listed [here](https://github.com/SAP/service-fabrik-broker/wiki/Process-to-Submit-a-Patch), after implementing the review comments.

A seperate PR will be needed in the service fabrik product component to expose virtual_host plan and add virtual_host.provider in the manifest that provides credentials to access s3/swift rabbitmq container.


config/settings.yml
```diff
@@ -243,6 +243,19 @@ defaults: &defaults
       password: <%= cloudProvider.password %>
       strictSSL: false
       container: <%= cloudProvider.containerPrefix %>-broker
+  virtual_host:
```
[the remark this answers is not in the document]
pranjaljain: Virtual host manager needs to store relationship between virtual host instance id and the deployment name on which it was created. 
They could be stored in any container. We discussed with Jagadish and decided that instead of storing it inside the container used by service fabrik we will use a separate container and hence
separate setting for provider of virtual host will be needed.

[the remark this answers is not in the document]
subhankarc: Please change the container name in the settings then. Also I would imagine the other values of provider would be same. Please try to reuse them in that case instead of copying them again.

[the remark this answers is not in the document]
pranjaljain: We are planning to reuse the provider container speicifc for rabbitmq service that already exists in landscapes. Other services use it for backup metadata purposes but because for RabbitMQ there is no such use case we will reuse it to store virtual host instances metadata.
Credentials for this container are different than the credentials for service-fabrik broker container. 
In settings.yml I will add a separate section for virtual host cloud provider settings and will use it to at virtual_host.provider.
Is it fine?

config/settings.yml
```diff
@@ -579,6 +615,7 @@ development: &development
       - state
       - lifecycle
       - credentials
+      - multi_tenancy
```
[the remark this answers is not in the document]
pranjaljain: In service broker agent we need to add extra APIs to support lifecycle operations of virtual hosts. Collectively these will the APIs to support "multi_tenancy" feature. That's why it is required to add a new feature.

lib/cf/CloudControllerClient.js
```diff
@@ -275,6 +278,27 @@ class CloudControllerClient extends HttpClient {
     return this.getResource(`/v2/organizations/${org_guid}`, options);
   }
 
+  getMultiTenantRabbitMQServiceInstanceIdWithName(rabbitmq_instance_name, org_guid, space_guid, options) {
```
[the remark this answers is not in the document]
pranjaljain: Is it fine if I make the signatures of the functions as?
getMultiTenantServiceInstanceIdWithName(instance_name, org_guid, space_guid, service_name, options)

Then instead of rabbitmq being in the name of the function I will pass it as a parameter from the VirtualHostManager which is anyways very specific to rabbitmq. The method inside CloudControllerClient will become generic because all the references to rabbitmq will be gone.

config/settings.yml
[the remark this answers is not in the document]
subhankarc: Yes, that will be better then.

lib/cf/CloudControllerClient.js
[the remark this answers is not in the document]
subhankarc: Yes that is better. Can you please patch it? Let's see how it looks like.

config/settings.yml
```diff
@@ -51,6 +51,38 @@ const cloudProviders = {
   }
 };
 const cloudProvider = cloudProviders.openstack;
+
+const virtualHostCloudProviders = {
```
jagadish-kb: duplicated configs. See if the config `cloudProviders` can be reused here.

config/settings.yml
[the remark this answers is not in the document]
jagadish-kb: Since VHOSTManager is specific to RabbitMq, is there any RabbitMQ Dedicated service plan which does not support multi_tenancy? Or is it supposed to be supported by all plans? If all plans of RabbitMQ must support it, then should this be required ?

lib/cf/CloudControllerClient.js
```diff
@@ -275,6 +278,27 @@ class CloudControllerClient extends HttpClient {
     return this.getResource(`/v2/organizations/${org_guid}`, options);
   }
 
+  getMultiTenantServiceInstanceIdWithName(instance_name, org_guid, space_guid, service_name, options) {
```
jagadish-kb: can you rename this method to `getServiceInstanceByName` - no need to have the `mutliTenant` added to the method signature as the same can be reused elsewhere in the code where one wants to retrieve instance details by name.

lib/cf/CloudControllerClient.js
```diff
@@ -275,6 +278,27 @@ class CloudControllerClient extends HttpClient {
...
+    return this.getResource(`/v2/spaces/${space_guid}/service_instances?q=name:${instance_name}`, options)
+      .then((serviceInstance) => {
+        if (serviceInstance.resources.length < 1) {
+          logger.debug(`Service instance with name ${instance_name} not found in the space ${space_guid}`);
+          throw new MultiTenantServiceInstanceNameNotFound(instance_name, service_name);
```
jagadish-kb: this error's message reads `Could not find ${service} Service Instance with Name ${name} that supports multi tenancy` - however in this condition, instance with this name itself is not found. So wrong error is being thrown. Why not just throw `errors.NotFound` here?

lib/cf/CloudControllerClient.js
```diff
@@ -275,6 +278,27 @@ class CloudControllerClient extends HttpClient {
     return this.getResource(`/v2/organizations/${org_guid}`, options);
   }
 
+  getMultiTenantServiceInstanceIdWithName(instance_name, org_guid, space_guid, service_name, options) {
+    return this.getResource(`/v2/spaces/${space_guid}/service_instances?q=name:${instance_name}`, options)
+      .then((serviceInstance) => {
+        if (serviceInstance.resources.length < 1) {
+          logger.debug(`Service instance with name ${instance_name} not found in the space ${space_guid}`);
```
jagadish-kb: change log level here to `warn` or `error`

lib/cf/CloudControllerClient.js
```diff
@@ -275,6 +278,27 @@ class CloudControllerClient extends HttpClient {
...
+          throw new MultiTenantServiceInstanceNameNotFound(instance_name, service_name);
+        }
+        var serviceInstanceId = serviceInstance.resources[0].metadata.guid;
```
jagadish-kb: dont use `var` use `const` and if its changing then use `let`

lib/cf/CloudControllerClient.js
jagadish-kb: `org_guid` passed in the signature does not seem to be used in the method.

lib/cf/CloudControllerClient.js
```diff
@@ -275,6 +278,27 @@ class CloudControllerClient extends HttpClient {
...
+        return this.findServicePlanByInstanceId(serviceInstanceId)
+          .then((servicePlan) => {
+            var catalogPlan = catalog.getPlan(servicePlan.entity.unique_id);
+            if (catalogPlan.service.name === service_name && catalogPlan.manager.settings.agent !== undefined && catalogPlan.manager.settings.agent.supported_features.indexOf('multi_tenancy') !== -1) {
+              return Promise.resolve(serviceInstanceId);
```
jagadish-kb: u can just return the value since its already part of promise, no need to `Promise.resolve` the value

lib/cf/CloudControllerClient.js
```diff
@@ -275,6 +278,27 @@ class CloudControllerClient extends HttpClient {
...
+      .then((serviceInstance) => {
+        if (serviceInstance.resources.length < 1) {
+          logger.debug(`Service instance with name ${instance_name} not found in the space ${space_guid}`);
+          throw new MultiTenantServiceInstanceNameNotFound(instance_name, service_name);
+        }
+        var serviceInstanceId = serviceInstance.resources[0].metadata.guid;
+        return this.findServicePlanByInstanceId(serviceInstanceId)
+          .then((servicePlan) => {
+            var catalogPlan = catalog.getPlan(servicePlan.entity.unique_id);
+            if (catalogPlan.service.name === service_name && catalogPlan.manager.settings.agent !== undefined && catalogPlan.manager.settings.agent.supported_features.indexOf('multi_tenancy') !== -1) {
```
jagadish-kb: wouldnt do this logic in `cloudcontroller` class. This is a pure DAO class. So it must be used to just retieve data. Any business logic should move up to manager.

config/settings.yml
pranjaljain: config cloudProviders can be reused if we create a new container with same credentials as of container for service-fabrik-broker.
We were planning to reuse the provider container specific for rabbitmq service that already exists in landscapes. Other services use it for backup metadata purposes but because for RabbitMQ there is no such use case we will reuse it to store virtual host instances metadata.
Credentials for this container are different than the credentials for service-fabrik broker container.
What do you suggest?

lib/cf/CloudControllerClient.js
```diff
@@ -275,6 +278,27 @@ class CloudControllerClient extends HttpClient {
     return this.getResource(`/v2/organizations/${org_guid}`, options);
   }
 
+  getMultiTenantServiceInstanceIdWithName(instance_name, org_guid, space_guid, service_name, options) {
+    return this.getResource(`/v2/spaces/${space_guid}/service_instances?q=name:${instance_name}`, options)
```
jagadish-kb: can you instead use this [cc api](https://apidocs.cloudfoundry.org/280/service_instances/list_all_service_instances.html)  This API is generic enough to list all instances in any org or space. So you can build the query `q` based on what is present in the options. By doing so you would make this DAO method generic and truly reusable for any other usecases as well.

lib/cf/CloudControllerClient.js
```diff
@@ -275,6 +278,27 @@ class CloudControllerClient extends HttpClient {
...
+        }
+        var serviceInstanceId = serviceInstance.resources[0].metadata.guid;
+        return this.findServicePlanByInstanceId(serviceInstanceId)
+          .then((servicePlan) => {
+            var catalogPlan = catalog.getPlan(servicePlan.entity.unique_id);
+            if (catalogPlan.service.name === service_name && catalogPlan.manager.settings.agent !== undefined && catalogPlan.manager.settings.agent.supported_features.indexOf('multi_tenancy') !== -1) {
+              return Promise.resolve(serviceInstanceId);
+            } else {
+              logger.debug(`Service instance with name ${instance_name} was found in the space ${space_guid} but it was not a ${service_name} instance with multi-tenancy feature support.`);
+              throw new MultiTenantServiceInstanceNameNotFound(instance_name, service_name);
```
jagadish-kb: what is the need for this new error type. In the above case you can throw `errors.Forbidden` so semantically it makes more sense to send 403, instead of 404.

config/settings.yml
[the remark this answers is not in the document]
pranjaljain: Currently all the RabbitMQ 3.6 plans will support multi-tenancy. 
With RabbitMQ 3.7 we expect a better support for multi tenancy. 
Although currently we do not provide any plan with RabbitMQ 3.7, in future we will do, and then there is a possibility that some of the existing plans can withdraw support for multi-tenancy. Not fully sure though.

lib/controllers/ServiceBrokerApiController.js
```diff
@@ -71,10 +71,16 @@ class ServiceBrokerApiController extends BaseController {
...
+      /* jshint unused:false */
+      res.status(400).send(`{"err":{${err.message}}}`);
+    }
+
     return Promise
       .try(() => req.instance.create(params))
       .then(done)
-      .catch(ServiceInstanceAlreadyExists, conflict);
+      .catch(ServiceInstanceAlreadyExists, conflict)
+      .catch(BadRequest, badRequest);
```
jagadish-kb: THis is not required. BadRequest is automatically sent with 400 as part of global error handler in middleware.

lib/controllers/ServiceBrokerApiController.js
```diff
@@ -71,10 +71,16 @@ class ServiceBrokerApiController extends BaseController {
       res.status(409).send({});
     }
 
+    function badRequest(err) {
```
jagadish-kb: THis is not required. BadRequest is automatically sent with 400 as part of global error handler in middleware.

lib/errors.js
```diff
@@ -311,4 +311,12 @@ class DBUnavailable extends HttpServerError {
     super(503, 'DB Unavailable', message);
   }
 }
-exports.DBUnavailable = DBUnavailable;
\ No newline at end of file
+exports.DBUnavailable = DBUnavailable;
+
+// Errors specific to virtualhost manager
+class MultiTenantServiceInstanceNameNotFound extends BadRequest {
```
jagadish-kb: Dont think we should use this new error type. It breaks HTTP Semantics.

lib/fabrik/VirtualHostAgent.js
```diff
@@ -0,0 +1,50 @@
+'use strict';
+
+const Agent = require('./Agent');
+
+class VirtualHostAgent extends Agent {
+
```
jagadish-kb: constructor starts immediately. Extra line here.

lib/fabrik/VirtualHostAgent.js
```diff
@@ -0,0 +1,50 @@
...
+    });
+    this.settings = settings;
+  }
+
+  provision(ips, guid) {
+    const body = {
+      parameters: {}
+    };
+    return this
+      .getHost(ips, 'multi-tenancy')
```
jagadish-kb: move `multi-tenancy` to constants.js

lib/fabrik/VirtualHostAgent.js
```diff
@@ -0,0 +1,50 @@
...
+      .then(ip => this.post(ip, `tenants/${guid}/lifecycle/provision`, body, 200));
```
jagadish-kb: can omit `lifecycle/provison` from URL - a HTTP post on `tentants/$guid` - is semantically create.

lib/fabrik/VirtualHostAgent.js
```diff
@@ -0,0 +1,50 @@
...
+  }
+
+  deprovision(ips, guid) {
+    const body = {
+      parameters: {}
+    };
+    return this
+      .getHost(ips, 'multi-tenancy')
+      .then(ip => this.post(ip, `tenants/${guid}/lifecycle/deprovision`, body, 200));
```
jagadish-kb: wrong HTTP semantics. This should be HTTP Delete on the resource `tentants/$guid`

lib/fabrik/VirtualHostAgent.js
```diff
@@ -0,0 +1,50 @@
...
+  }
+
+  createCredentials(ips, guid, parameters) {
+    const body = {
+      parameters: parameters
+    };
+    return this
+      .getHost(ips, 'credentials')
+      .then(ip => this.post(ip, `tenants/${guid}/credentials/create`, body, 200));
```
jagadish-kb: bad practice - HTTP 200 on a resource is create, calling `create` once again on URL isnt right. Can drop the suffix `create` and just have the resource as `tenants/${guid}/credentials/`

lib/fabrik/VirtualHostAgent.js
```diff
@@ -0,0 +1,50 @@
...
+  }
+
+  deleteCredentials(ips, guid, credentials) {
+    const body = {
+      credentials: credentials
+    };
+    return this
+      .getHost(ips, 'credentials')
+      .then(ip => this.post(ip, `tenants/${guid}/credentials/delete`, body, 200));
```
jagadish-kb: wrong HTTP semantics/ bad practice. Use HTTP Delete on resource `tenants/${guid}/credentials/`
suffix `/delete` can be omitted, its redundant.

lib/fabrik/VirtualHostInstance.js
```diff
@@ -0,0 +1,93 @@
+const _ = require('lodash');
+const Promise = require('bluebird');
+const BaseInstance = require('./BaseInstance');
+const bosh = require('../bosh');
+const utils = require('../utils');
+const mapper = require('./VirtualHostRelationMapper');
+const catalog = require('../models').catalog;
+
+class VirtualHostInstance extends BaseInstance {
+
```
jagadish-kb: unnecessary new line. Look at other classes for consistency, remove this.

lib/fabrik/VirtualHostInstance.js
```diff
@@ -0,0 +1,93 @@
...
+    return this.getDeploymentNameForVirtualHostInstanceId(this.guid)
+      .then((deploymentName) => this.deploymentName = deploymentName);
+  }
+
+  create(params) {
+    const rabbitmqInstanceName = params.parameters.dedicated_rabbitmq_instance;
+    return this.cloudController.getMultiTenantServiceInstanceIdWithName(rabbitmqInstanceName, params.organization_guid, params.space_guid, 'rabbitmq')
+      .then((serviceInstanceId) => {
+        return this.director.getDeploymentNameForInstanceId(serviceInstanceId);
+      })
```
jagadish-kb: unnecesary return when there is just one line Can just inline the code like this and have the promise chain. 
```
this.cloudController.getMultiTenantServiceInstanceIdWithName(rabbitmqInstanceName, params.organization_guid, params.space_guid, 'rabbitmq')
 .then(serviceInstanceId => this.director.getDeploymentNameForInstanceId(serviceInstanceId)
 .then(deploymentName => this.manager.createOrUpdateVirtualHost(deploymentName, params, this.guid));
```

lib/fabrik/VirtualHostInstance.js
```diff
@@ -0,0 +1,93 @@
...
+  }
+
+  initialize() {
+    return this.getDeploymentNameForVirtualHostInstanceId(this.guid)
+      .then((deploymentName) => this.deploymentName = deploymentName);
+  }
+
+  create(params) {
+    const rabbitmqInstanceName = params.parameters.dedicated_rabbitmq_instance;
+    return this.cloudController.getMultiTenantServiceInstanceIdWithName(rabbitmqInstanceName, params.organization_guid, params.space_guid, 'rabbitmq')
```
jagadish-kb: Why is the `getMultiTenantServiceInstanceIdWithName` method retrieving service plan & validating the service plan? Anyways you seem to have added the lifecycle 'multi-tenancy'. After fetching instance guid, you can directly invoke the BOSH agent & if no agent supports multi-tenancy it directly throws the error `Feature Not supported`. Most of the times we can avoid unnecessary call to CloudController.

lib/fabrik/VirtualHostInstance.js
```diff
@@ -0,0 +1,93 @@
...
+      .then((serviceInstanceId) => {
+        return this.director.getDeploymentNameForInstanceId(serviceInstanceId);
+      })
+      .then((deploymentName) => {
+        return this.manager
+          .createOrUpdateVirtualHost(deploymentName, params, this.guid);
```
jagadish-kb: can we update a virtual host? if not, then why is the method named `createOrUpdate`. If yes, then dont see an `update` method in this class.

lib/fabrik/VirtualHostInstance.js
```diff
@@ -0,0 +1,93 @@
...
+          service: this.plan.service,
+          instance: instance,
+          parent_plan: parent_instance_plan,
+          parent_instance: parent_instance
+        }));
+    });
+  }
+
+  getDeploymentNameForVirtualHostInstanceId(id) {
+    return this.mapper.getDeploymentName(`virtual_host_instance-${id}`);
```
jagadish-kb: `virtual_host_instance-${` seems unnecessary. The method `getDeploymentName` signature second parameter says `isntance_guid`, however u seem to have prefixed it with some constant. This can be removed and can pass the instance id as is.

lib/fabrik/VirtualHostInstance.js
```diff
@@ -0,0 +1,93 @@
...
+          plan: this.plan,
+          service: this.plan.service,
+          instance: instance,
+          parent_plan: parent_instance_plan,
+          parent_instance: parent_instance
+        }));
+    });
+  }
+
+  getDeploymentNameForVirtualHostInstanceId(id) {
```
jagadish-kb: the method name having `VirtualHost` seems redundant. We are anyways in the code flow/class of virutalhost, so can drop this as its understood and just name it as `getDeploymentNameForInstanceId`

lib/fabrik/VirtualHostInstance.js
```diff
@@ -0,0 +1,93 @@
...
+    this.mapper = mapper.VirtualHostRelationMapper;
+  }
+
+  get async() {
+    return false;
+  }
+
+  initialize() {
+    return this.getDeploymentNameForVirtualHostInstanceId(this.guid)
+      .then((deploymentName) => this.deploymentName = deploymentName);
```
jagadish-kb: can remove paranthesis around `deploymentName` as a convention when there is single param, its avoided in our code base.

lib/fabrik/VirtualHostInstance.js
```diff
@@ -0,0 +1,93 @@
...
+  }
+
+  create(params) {
+    const rabbitmqInstanceName = params.parameters.dedicated_rabbitmq_instance;
+    return this.cloudController.getMultiTenantServiceInstanceIdWithName(rabbitmqInstanceName, params.organization_guid, params.space_guid, 'rabbitmq')
+      .then((serviceInstanceId) => {
+        return this.director.getDeploymentNameForInstanceId(serviceInstanceId);
+      })
+      .then((deploymentName) => {
```
jagadish-kb: same as [above](https://github.com/cloudfoundry-incubator/service-fabrik-broker/pull/111/files#r156889902)

lib/fabrik/VirtualHostInstance.js
```diff
@@ -0,0 +1,93 @@
...
+  }
+
+  unbind(params) {
+    return this
+      .initialize()
+      .then(() => this.manager.deleteBinding(this.deploymentName, this.guid, params.binding_id));
+  }
+
+  getInfo() {
+    return this.initialize().then(() => {
```
jagadish-kb: move .then() to next line - generally done across the code base.

lib/fabrik/VirtualHostInstance.js
```diff
@@ -0,0 +1,93 @@
...
+  create(params) {
+    const rabbitmqInstanceName = params.parameters.dedicated_rabbitmq_instance;
+    return this.cloudController.getMultiTenantServiceInstanceIdWithName(rabbitmqInstanceName, params.organization_guid, params.space_guid, 'rabbitmq')
+      .then((serviceInstanceId) => {
+        return this.director.getDeploymentNameForInstanceId(serviceInstanceId);
+      })
+      .then((deploymentName) => {
+        return this.manager
+          .createOrUpdateVirtualHost(deploymentName, params, this.guid);
+      });
```
jagadish-kb: Have u verified the dashboard URL that is getting generated? And does that dashboard load the contents that you are supposed to be provided with?

lib/fabrik/VirtualHostManager.js
```diff
@@ -0,0 +1,168 @@
...
+const NotImplemented = errors.NotImplemented;
+const ServiceInstanceNotOperational = errors.ServiceInstanceNotOperational;
+const BadRequest = errors.BadRequest;
+const ServiceBindingAlreadyExists = errors.ServiceBindingAlreadyExists;
+const ServiceBindingNotFound = errors.ServiceBindingNotFound;
+const NotFound = errors.NotFound;
+const mapper = require('./VirtualHostRelationMapper');
+
+class VirtualHostManager extends BaseManager {
+
```
jagadish-kb: unnecessary new line

lib/fabrik/VirtualHostManager.js
```diff
@@ -0,0 +1,168 @@
...
+      })
+      .then(manifest => _
+        .chain(manifest.jobs)
+        .map(job => _.map(job.networks, net => net.static_ips))
+        .flattenDeep()
+        .value()
+      );
+  }
+
+  createOrUpdateVirtualHost(deploymentName, params, guid) {
```
jagadish-kb: if we are not updating rename the method to `createVirtualHost`

lib/fabrik/VirtualHostAgent.js
```diff
@@ -0,0 +1,50 @@
...
+class VirtualHostAgent extends Agent {
+
+  constructor(settings) {
+    super({
+      json: true
+    });
+    this.settings = settings;
+  }
+
+  provision(ips, guid) {
```
jagadish-kb: for the sake of consistence across clasess can this be renamed to `createVirtualHost` or `crerateVhost` - the same operation is called differently along the flow from controller to instance -> manager -> agent.

lib/fabrik/VirtualHostManager.js
```diff
@@ -0,0 +1,168 @@
...
+    this.verifyFeatureSupport('multi_tenancy');
+    logger.info(`Creating virtual host for deployment '${deploymentName}'...`);
+    return this.getDeploymentIps(deploymentName)
+      .then(ips => this.agent.provision(ips, guid))
+      .tap(() => this.saveVirtualHostInstanceToDeploymentRelation(deploymentName, guid))
```
jagadish-kb: dont do this in background. this saving the mapping must happen in promise chain with .then(), because if the mapping is not saved and it results in an error, then it must error out the operation. Else, down the line when a restart happens this mapping is lost.

lib/fabrik/VirtualHostManager.js
```diff
@@ -0,0 +1,168 @@
...
+      .catch(err => {
```
jagadish-kb: unnecessary catch and log. This is anyways caught in global error handler.

lib/fabrik/VirtualHostManager.js
```diff
@@ -0,0 +1,168 @@
...
+      });
+  }
+
+  deleteVirtualHost(deploymentName, guid) {
+    this.verifyFeatureSupport('multi_tenancy');
+    logger.info(`deleting virtual host for deployment '${deploymentName}'...`);
+    return this.getDeploymentIps(deploymentName)
+      .then(ips => this.agent.deprovision(ips, guid))
+      .tap(() => this.deleteVirtualHostInstanceToDeploymentRelation(deploymentName, guid))
+      .catch(err => {
```
jagadish-kb: unnecessary catch and log. This is anyways caught in global error handler.

lib/fabrik/VirtualHostManager.js
```diff
@@ -0,0 +1,168 @@
...
+      });
+  }
+
+  createBinding(deploymentName, guid, binding) {
+    this.verifyFeatureSupport('credentials');
+    logger.info(`Creating binding '${binding.id}' for deployment '${deploymentName}', virtual host '${guid}'...`);
+    logger.info('+-> Binding parameters:', binding.parameters);
+    return this.getDeploymentIps(deploymentName)
+      .then(ips => this.agent.createCredentials(ips, guid, binding.parameters))
+      .tap(credentials => this.createBindingProperty(deploymentName, binding.id, _.set(binding, 'credentials', credentials)))
```
jagadish-kb: this should happen with promise chain with .then() - because if binding creds creation fails, then the operation should also be flagged as failed.

lib/fabrik/VirtualHostManager.js
```diff
@@ -0,0 +1,168 @@
...
+      .tap(() => {
+        const bindCreds = _.cloneDeep(binding.credentials);
+        utils.maskSensitiveInfo(bindCreds);
+        logger.info(`+-> Created binding:${JSON.stringify(bindCreds)}`);
+      })
+      .catch(err => {
```
jagadish-kb: unnecessary catch & log

lib/fabrik/VirtualHostManager.js
```diff
@@ -0,0 +1,168 @@
...
+    logger.info(`Deleting binding '${id}' for deployment '${deploymentName}' , virtual host '${guid}'...`);
+    return Promise
+      .all([
+        this.getDeploymentIps(deploymentName),
+        this.getBindingProperty(deploymentName, id)
+      ])
+      .spread((ips, binding) => this.agent.deleteCredentials(ips, guid, binding.credentials))
+      .then(() => this.deleteBindingProperty(deploymentName, id))
+      .tap(() => logger.info('+-> Deleted service binding'))
+      .catch(err => {
```
jagadish-kb: unnecessary catch throw

lib/fabrik/VirtualHostManager.js
```diff
@@ -0,0 +1,168 @@
...
+        logger.error('+-> Failed to delete binding');
+        logger.error(err);
+        throw err;
+      });
+  }
+
+  saveVirtualHostInstanceToDeploymentRelation(deploymentName, id) {
+    return this.mapper
+      .createVirtualHostRelation(deploymentName, `virtual_host_instance-${id}`)
```
jagadish-kb: can drop this prefix `virtual_host_instance-$`

lib/fabrik/VirtualHostManager.js
jagadish-kb: what guid is here? can you name it `instanceId` or `instanceGuid` to signify what is in it.

lib/fabrik/VirtualHostManager.js
```diff
@@ -0,0 +1,168 @@
...
+      .then(() => this.deleteBindingProperty(deploymentName, id))
+      .tap(() => logger.info('+-> Deleted service binding'))
+      .catch(err => {
+        logger.error('+-> Failed to delete binding');
+        logger.error(err);
+        throw err;
+      });
+  }
+
+  saveVirtualHostInstanceToDeploymentRelation(deploymentName, id) {
```
jagadish-kb: what is `id` in this signature. Can your method signatures use some consistency on these ids which are used in params?

lib/fabrik/VirtualHostAgent.js
pranjaljain: For create credentials, I tried keeping URL format similar to what is listed here, https://cloudfoundry-incubator.github.io/service-fabrik-broker/api/agent_v1.1.html. Same is used https://github.com/cloudfoundry-incubator/service-fabrik-broker/blob/master/lib/fabrik/Agent.js#L129. 
Should I change the format?

lib/fabrik/VirtualHostAgent.js
pranjaljain: For delete credentials, I tried keeping URL format similar to what is listed here, https://cloudfoundry-incubator.github.io/service-fabrik-broker/api/agent_v1.1.html. Same is used https://github.com/cloudfoundry-incubator/service-fabrik-broker/blob/master/lib/fabrik/Agent.js#L138
Should I change the format?

lib/fabrik/VirtualHostInstance.js
pranjaljain: getMultiTenantServiceInstanceIdWithName method retrieving service plan & validating the service plan just to make sure:
(1) Service instance name passed to create virtual host is a 'rabbitmq' service instance only.
(2) rabbitmq instance supports multi-tenancy & it is not a dev plan service instance.

Without this check the calls with reach broker agents of instances other then rabbitmq also if wrong instance name is provided.
Hope this is fine?

jagadish-kb: Without this check the calls with reach broker agents of instances other then rabbitmq also if wrong instance name is provided.
-- Agreed, but it avoids the extra call to cloud controller. Either ways there is anyways an additional network call in case of wrong instance names - in your case ur validating against CC & i am suggesting let that validation be part of getting the request executed by the agents. The benefit of the later (along with validation) ensures additional call to CC is not required.

lib/fabrik/VirtualHostManager.js
```diff
@@ -0,0 +1,168 @@
...
+      .getDeploymentManifest(deploymentName)
+      .tap(() => logger.info('+-> Fetched deployment manifest'))
+      .catch(err => {
+        logger.error('+-> Failed to fetch deployment manifest');
+        logger.error(err);
+        throw err;
+      });
+  }
+
+  getDeploymentIps(deploymentName) {
```
jagadish-kb: code duplication with `DirectorManager` - better to move this method `BoshDirectorClient` and use the same from both here & `DirectorManager`.

lib/fabrik/VirtualHostManager.js
```diff
@@ -0,0 +1,168 @@
...
+    this.mapper = mapper.VirtualHostRelationMapper;
+  }
+
+  verifyFeatureSupport(feature) {
+    if (!_.includes(this.agent.features, feature)) {
+      throw new NotImplemented(`Feature '${feature}' not supported`);
+    }
+  }
+
+  getDeploymentManifest(deploymentName) {
```
jagadish-kb: Once the [below](https://github.com/cloudfoundry-incubator/service-fabrik-broker/pull/111/files#r157011348) comment is incorporated, this method can be removed.

lib/fabrik/VirtualHostManager.js
```diff
@@ -0,0 +1,168 @@
...
+  saveVirtualHostInstanceToDeploymentRelation(deploymentName, id) {
+    return this.mapper
+      .createVirtualHostRelation(deploymentName, `virtual_host_instance-${id}`)
+      .catchThrow(BadRequest, new BadRequest(`Virtual host instance with ID ${id} already exists on deployment ${deploymentName}`));
+  }
+
+  deleteVirtualHostInstanceToDeploymentRelation(deploymentName, id) {
+    return this.mapper
+      .deleteVirtualHostRelation(`virtual_host_instance-${id}`)
+      .catchThrow(BadRequest, new BadRequest(`Virtual host instance with ID ${id} does not exists on deployment ${deploymentName}`));
```
jagadish-kb: Delete operations must be idempotent. If a delete request for an already deleted request comes in, must return back with HTTP STATUS 410 - GONE & should not throw an error.

jagadish-kb: From here you could rather throw `ServiceBindingNotFound` and in the controller you can treat this exception with HTTP Status 410 - look [here for ref](https://github.com/cloudfoundry-incubator/service-fabrik-broker/blob/master/lib/controllers/ServiceBrokerApiController.js#L211)

lib/fabrik/VirtualHostManager.js
```diff
@@ -0,0 +1,168 @@
...
+  }
+
+  getLockProperty(deploymentName) {
```
jagadish-kb: why is this method created? where is it used?

lib/fabrik/VirtualHostManager.js
```diff
@@ -0,0 +1,168 @@
...
+  deleteBindingProperty(deploymentName, id) {
+    return this.director
+      .deleteDeploymentProperty(deploymentName, `binding-${id}`);
+  }
+
+  static get instanceConstructor() {
+    return VirtualHostInstance;
+  }
+
+  static load(plan) {
```
jagadish-kb: code is the same as that in `DirectorManager` - can move this code to `BaseManager`. `DockerManager` anyways has a different implementation and that override can remain.

lib/fabrik/VirtualHostRelationMapper.js
```diff
@@ -0,0 +1,66 @@
...
+    logger.info(`Fetching deployment name for virtual_host instance id : '${instanceId}'`);
+    if (this.cache[instanceId] === undefined) {
+      logger.info(`Cache miss for deployment name for virtual_host instance id ${instanceId}, Will load it from store.`);
+      return this.loadCacheforInstance(instanceId);
+    }
+    return Promise.resolve(this.cache[instanceId]);
+  }
+
+  loadCacheforInstance(instanceId) {
+    const data = {
```
jagadish-kb: input params to be named as `options`.

lib/fabrik/VirtualHostRelationMapper.js
```diff
@@ -0,0 +1,66 @@
...
+    return this.virtualHostStore.putFile(data);
+  }
+
+  getDeploymentName(instanceId) {
+    logger.info(`Fetching deployment name for virtual_host instance id : '${instanceId}'`);
+    if (this.cache[instanceId] === undefined) {
+      logger.info(`Cache miss for deployment name for virtual_host instance id ${instanceId}, Will load it from store.`);
+      return this.loadCacheforInstance(instanceId);
+    }
+    return Promise.resolve(this.cache[instanceId]);
```
jagadish-kb: nothing wrong with this, but as a convention in such scenarios the method is wrapped with Promise.try & a normal return is performed. Look at other code references and for keeping it consistent, please change it.

lib/fabrik/VirtualHostRelationMapper.js
```diff
@@ -0,0 +1,66 @@
...
+    return this.virtualHostStore.getVirtualHostFile(data)
+      .then((metadata) => {
+        this.cache[metadata.instance_guid] = metadata.deployment_name;
+        return Promise.resolve(metadata.deployment_name);
+      });
+  }
+
+  deleteVirtualHostRelation(instanceId) {
+    logger.info(`Deleting deployment name relation for virtual_host instance id : '${instanceId}'`);
+    const data = {
```
jagadish-kb: same as [above](https://github.com/cloudfoundry-incubator/service-fabrik-broker/pull/111/files#r157015693)

lib/fabrik/VirtualHostRelationMapper.js
```diff
@@ -0,0 +1,66 @@
...
+      'instance_guid': instanceId
+    };
+    return this.virtualHostStore.removeFile(data)
+      .then(() => {
```
jagadish-kb: unnecessary `{` & `return` when having single statement returns like below both can be ignored.

jagadish-kb: also if a file is not present, ensure controller does not throw exception. Deletes must be idempotent & should return 410 - Gone.

lib/iaas/VirtualHostStore.js
```diff
@@ -0,0 +1,79 @@
+'use strict';
+
+const _ = require('lodash');
+const path = require('path');
+
+class VirtualHostStore {
+
```
jagadish-kb: unnecessary new line

lib/fabrik/VirtualHostAgent.js
jagadish-kb: The ones highlighted are wrong implementation. Must not follow bad samples in existing code, when writing new code :-)
Yes please change it to the way recommended.

lib/fabrik/VirtualHostAgent.js
jagadish-kb: same as [above](https://github.com/cloudfoundry-incubator/service-fabrik-broker/pull/111/files#r157019650)

config/settings.yml
jagadish-kb: Ok you can keep seperate config since its different from broker.

lib/fabrik/VirtualHostInstance.js
```diff
@@ -0,0 +1,102 @@
...
+  }
+
+  initialize() {
+    return this.getDeploymentNameForInstanceId(this.guid)
+      .then(deploymentName => this.deploymentName = deploymentName);
+  }
+
+  create(params) {
+    const rabbitmqInstanceName = params.parameters.dedicated_rabbitmq_instance;
+    return this.cloudController.getServiceInstances({
```
jagadish-kb: can you move this piece of code between Line 30 - 40, into cloud controller client into a new method named getInstanceByName(name, space_guid)? The rules of forming query & syntax to query CF should be within cloud controller client class and not spill outside of it.  This method if successful should return  `serviceInstances[0]` else `throw new NotFound(``Service instance with name ${instanceName} not found``);`

lib/fabrik/VirtualHostManager.js
```diff
@@ -0,0 +1,89 @@
...
+    this.agent = new VirtualHostAgent(this.settings.agent);
+    this.mapper = mapper.VirtualHostRelationMapper;
+  }
+
+  createVirtualHost(deploymentName, params, instanceId) {
+    logger.info(`Creating virtual host for deployment '${deploymentName}'...`);
+    return this.director.getDeploymentIps(deploymentName)
+      .then(ips => this.agent.createVirtualHost(ips, instanceId))
+      .then(() => this.mapper.createVirtualHostRelation(deploymentName, instanceId))
+      .tap(() => logger.info('+-> Created virtual host'));
```
jagadish-kb: Modify the above logger to include '${deploymentName}' - not having this context makes this logger difficult when we would have many such logs which will all be same.

lib/fabrik/VirtualHostManager.js
```diff
@@ -0,0 +1,89 @@
...
+  }
+
+  deleteVirtualHost(deploymentName, instanceId) {
+    logger.info(`Deleting virtual host for deployment '${deploymentName}'...`);
+    return this.director.getDeploymentIps(deploymentName)
+      .then(ips => this.agent.deleteVirtualHost(ips, instanceId))
+      .then(() => this.mapper.deleteVirtualHostRelation(instanceId))
+      .tap(() => logger.info('+-> Deleted virtual host'))
```
jagadish-kb: same as [above](https://github.com/cloudfoundry-incubator/service-fabrik-broker/pull/111/files#r157671566)

lib/fabrik/VirtualHostManager.js
```diff
@@ -0,0 +1,89 @@
...
+      .then(ips => this.agent.createVirtualHost(ips, instanceId))
+      .then(() => this.mapper.createVirtualHostRelation(deploymentName, instanceId))
+      .tap(() => logger.info('+-> Created virtual host'));
+  }
+
+  deleteVirtualHost(deploymentName, instanceId) {
+    logger.info(`Deleting virtual host for deployment '${deploymentName}'...`);
+    return this.director.getDeploymentIps(deploymentName)
+      .then(ips => this.agent.deleteVirtualHost(ips, instanceId))
+      .then(() => this.mapper.deleteVirtualHostRelation(instanceId))
```
jagadish-kb: Delete involves deleting the vhost and deleting the mapping info on container. There can be a scenario where delete of vhost may succeed but delete vhost mapping might fail, which would result in an error and instance deletion will fail. 
now on retry, in the above flow delete virtual host would most probably throw 404 & we wont even attempt to delete vhost relation mapping. When delete vhost says 404, we must handle it and still still go ahead and try to delete this vhost relation.
This can be done in couple of ways, one of the below is my suggestion:
```
return this.director.getDeploymentIps(deploymentName)
           .then(ips => Promise.all([this.agent.deleteVirtualHost(ips, instanceId),
                                                     this.mapper.deleteVirtualHostRelation(instanceId)]);
```
Now doing so ensures both of these happen every time. Ofcse, if deployment itself is deleted `this.director.getDeploymentIps` would throw 404, but i guess there would be some checks not to delete deployment when there are vhosts on it right?

lib/fabrik/VirtualHostManager.js
```diff
@@ -0,0 +1,89 @@
...
+      .tap(() => logger.info('+-> Deleted virtual host'))
+      .catchThrow(NotFound, new ServiceInstanceNotFound(instanceId));
+  }
+
+  createBinding(deploymentName, instanceId, binding) {
+    logger.info(`Creating binding '${binding.id}' for deployment '${deploymentName}', virtual host '${instanceId}'...`);
```
jagadish-kb: merge both the log stmts into one. Else at times with many requests flowing in, these two might not be identified together.

lib/fabrik/VirtualHostManager.js
pranjaljain: I will incorporate above changes.

For deletion of deployment:
I implemented in agent /deprovision call while delete deployment not to allow deletion if some virtual hosts already exists.
But, It also has an issue that in DirectorInstance, deleteSecurityGroup happens in initialize() without even waiting for /deprovision called in manager.deleteDeployment to respond. So even if /deprovision fails, service-instance becomes non operational becuase of deletion of security-group.
Moving security group deletion from https://github.com/cloudfoundry-incubator/service-fabrik-broker/blob/master/lib/fabrik/DirectorInstance.js#L56 to https://github.com/cloudfoundry-incubator/service-fabrik-broker/blob/master/lib/fabrik/DirectorInstance.js#L170 should solve it.
What do you suggest?

pranjaljain: One other way is:
return this.director.getDeploymentIps(deploymentName)
      .then(ips => this.agent.deleteVirtualHost(ips, instanceId))
      .then(() => this.mapper.deleteVirtualHostRelation(instanceId))
      .tap(() => logger.info(`+-> Deleted virtual host for deployment '${deploymentName}'`))
      .catch(NotFound, err => {
        /* jshint unused:false */
        this.mapper.deleteVirtualHostRelation(instanceId);
        throw new ServiceInstanceNotFound(instanceId);
      });

In case of any NotFound issue it will perform clean up task of removing entry from the provider.

test/mocks/virtualHostAgent.js
```diff
@@ -0,0 +1,47 @@
+'use strict';
+
+const nock = require('nock');
+const credentials = {
+  host: 'example.org:31415',
+  uri: 'mongodb://5e91c9e75a0aec0a4d8e9c3523887576:12b6ca434be2dd8716d1009eb280e1c5@10.244.10.160:27017,10.11.252.21:27017,10.11.252.22:27017/8354aac89a3781b8?replicaSet=282477f09a2704e7e1ab1396f7721acc'
```
jagadish-kb: i know it does matter what you put in here as mock data, but it helps to have data which reflect the actual ones. Havent seen one though, but does it have a mongodb uri format for rabbit? if not, can you put in some mock which looks like a real one.

test/fabrik.VirtualHostAgent.spec.js
```diff
@@ -0,0 +1,196 @@
...
+        protocol: protocol,
+        port: port
+      }, opts));
+    }
+
+    function createUrl(pathname) {
+      return `${protocol}://${ip}:${port}/v${version}/${pathname}`;
+    }
+
+    beforeEach(function () {
```
jagadish-kb: you dont need to stub beforeEach and restore afterEach. You can have stubs at `before()` & `after()` at test case level.

test/fabrik.VirtualHostAgent.spec.js
```diff
@@ -0,0 +1,196 @@
...
+
+      after(function () {
+        _.unset(body, 'parameters');
+      });
+
+      it('returns a JSON object', function () {
+        return agent
+          .createVirtualHost(ips, instanceId)
+          .then(() => {
+            expect(requestStub).to.have.been.calledTwice;
```
jagadish-kb: it would also be good to add assertion to `expect` the first call is GET & second call is POST.

test/fabrik.VirtualHostAgent.spec.js
```diff
@@ -0,0 +1,196 @@
...
+        pathname = `tenants/${instanceId}`;
+        expectedStatus = 204;
+        _.set(body, 'parameters', {});
+      });
+
+      it('returns a JSON object', function () {
+        return agent
+          .deleteVirtualHost(ips, instanceId)
+          .then(() => {
+            expect(requestStub).to.have.been.calledTwice;
```
jagadish-kb: same as [above](https://github.com/cloudfoundry-incubator/service-fabrik-broker/pull/111/files#r157937712)

test/fabrik.VirtualHostAgent.spec.js
```diff
@@ -0,0 +1,196 @@
...
+
+      after(function () {
+        _.unset(body, 'credentials');
+      });
+
+      it('returns a JSON object', function () {
+        return agent
+          .deleteCredentials(ips, instanceId, credentials)
+          .then(() => {
+            expect(requestStub).to.have.been.calledTwice;
```
jagadish-kb: same as [above](https://github.com/cloudfoundry-incubator/service-fabrik-broker/pull/111/files#r157937712)

test/acceptance/service-broker-api.instances.virtualHost.spec.js
```diff
@@ -0,0 +1,180 @@
...
+            })
+            .catch(res => {
+              expect(res).to.have.status(404);
+              mocks.verify();
+            });
+        });
+      });
+
+      describe('#deprovision', function () {
+        it('returns 200 OK', function () {
```
jagadish-kb: can you add a use case for deprovision when instance is not found? It must return 410 Gone when instance or virtual host not found.

test/acceptance/dashboard.virtualHost.spec.js
```diff
@@ -0,0 +1,92 @@
...
+        virtualHostStore.cloudProvider.getContainer()
+      ]);
+    });
+
+    afterEach(function () {
+      mocks.reset();
+    });
+
+    describe('/manage/instances/:service_id/:plan_id/:instance_id', function () {
+      this.slow(1500);
```
jagadish-kb: why is this requried? We generally dont have such overrides. Any reason these tests might take longer?

test/acceptance/dashboard.virtualHost.spec.js
```diff
@@ -0,0 +1,92 @@
...
+          })
+          .then(location => agent
+            .get(location.path)
+            .set('Accept', 'application/json')
+            .set('X-Forwarded-Proto', 'https')
+          )
+          .then(res => {
+            expect(res.body.userId).to.equal('me');
+            expect(res.body.instance.metadata.guid).to.equal(instance_id);
+            mocks.verify();
```
jagadish-kb: you are also adding parent instance info as part of dasbhboard, it would make sense to assert and check if that is present in response.

lib/fabrik/VirtualHostManager.js
```diff
@@ -0,0 +1,96 @@
...
+    let instanceDeleted = true;
+    return this.director.getDeploymentIps(deploymentName)
+      .tap(() => instanceDeleted = false)
+      .then(ips => Promise.all([this.agent.deleteVirtualHost(ips, instanceId),
+        this.mapper.deleteVirtualHostRelation(instanceId)
+      ]))
+      .tap(() => logger.info(`+-> Deleted virtual host for deployment '${deploymentName}'`))
+      .catch(NotFound, () => {
+        if (instanceDeleted) {
+          this.mapper.deleteVirtualHostRelation(instanceId);
```
jagadish-kb: make this `return this.mapper.deleteVirtualHostRelation(instanceId);`

lib/iaas/VirtualHostStore.js
```diff
@@ -0,0 +1,67 @@
...
+  removeFile(data) {
+    const filename = this.filename.stringify(data);
+    return this.cloudProvider.remove(filename);
+  }
+
+}
+
+class Filename {
+
```
jagadish-kb: newline

test/acceptance/dashboard.virtualHost.spec.js
pranjaljain: Not fully sure of times taken by all the mocks, but it is also used at docker dashboard unit tests https://github.com/cloudfoundry-incubator/service-fabrik-broker/blob/master/test/acceptance/dashboard.docker.spec.js#L35,
also the CI build failed because of default timeout.

jagadish-kb: quite interesting. It would be worth investigating some other time. For now then this is ok.

config/settings.yml
```diff
@@ -556,6 +617,88 @@ defaults: &defaults
...
+    dashboard_client:
+      id: 'service-fabrik-rabbitmq-client1'
+      secret: 'service-fabrik-rabbitmq-secret1'
+      redirect_uri: <%= redirect_uri %>
+    plan_updateable: true
+    plans:
+    ###########################
+    # DEDICATED SERVICE PLANS #
+    ###########################
+    - &rabbitmq_plan1
```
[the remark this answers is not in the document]
pranjaljain: As discussed, added following comment:
# An agent running in the deployment should implement Service Fabrik Agent APIs listed here https://cloudfoundry-incubator.github.io/service-fabrik-broker/api/agent_v1.1.html

test/acceptance/service-broker-api.instances.virtualHost.spec.js
pranjaljain: added this use case.

test/fabrik.VirtualHostAgent.spec.js
pranjaljain: An assertions are added that chacks that the first call is GET and second is POST.

test/fabrik.VirtualHostAgent.spec.js
pranjaljain: An assertions are added that chacks that the first call is GET and second is DELETE.

test/fabrik.VirtualHostAgent.spec.js
pranjaljain: An assertions are added that chacks that the first call is GET and second is DELETE.

config/templates/rabbitmq-manifest.yml.ejs
```diff
@@ -0,0 +1,150 @@
+<%= spec.header.select({disk_pools: [p('disk_pool')], resource_pools: [{name: p('resource_pool')}]}) %>
```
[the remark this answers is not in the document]
pranjaljain: As discussed, added following comment: An agent running in the deployment should implement Service Fabrik Agent APIs listed here https://cloudfoundry-incubator.github.io/service-fabrik-broker/api/agent_v1.1.html

lib/fabrik/VirtualHostManager.js
pranjaljain: If 'return' is added then it will not throw serviceInstanceNotFound and hence 410 gone, in case parent instance was already deleted.

Issue #117, opened 2017-12-06
Operation Status poller should cancel itself when deployment NotFound


sauravmndl:
**Description**

StatusPoller job should be triggered, log error and cancel itself when deployment is not found. But When it is getting triggered then, deployment is already deleted but it is not cancelling itself. Also it is scheduled for every 2 mins (configurable). so next time onward it is getting triggered unnecessarily. 

**If you are running service fabrik locally than Output of `node -v 
for node version and grep version package.json for service fabrik version`:**

```
(paste your output here)
node version: v6.9.1
service fabrik:  3.0.1
```

**Additional environment details (AWS, VirtualBox, OpenStack, etc.):**
Openstack

Issue #119, opened 2017-12-06
OOB backup metadata should store snapshotId if present


sauravmndl:
**Description**
For snapshot based oob backup the metadata should store snapshotId as well. 

**If you are running service fabrik locally than Output of `node -v 
for node version and grep version package.json for service fabrik version`:**

```
(paste your output here)
node version: v6.9.1
service fabrik: 3.0.1
```

**Additional environment details (AWS, VirtualBox, OpenStack, etc.):**

Issue #118, opened 2017-12-06
Show all OOB backups for a deployment also showing huge logs. 


sauravmndl:
**Description**
GET all oob backups api (/deployments/:name/backup) for a deployment should not print log.

**If you are running service fabrik locally than Output of `node -v 
for node version and grep version package.json for service fabrik version`:**

```
(paste your output here)
node version: v6.9.1
service fabrik: 3.0.1
```

**Additional environment details (AWS, VirtualBox, OpenStack, etc.):**

Issue #11, opened 2017-08-17
BOSH based service provisioning using BOSH 2.0 features like cloud-config.


subhankarc:
Service Fabrik currently uses BOSH1 way of provisioning services, which reserves 8 IPs for a service instance and calculate the network manually. With, BOSH2.0, SF can leverage cloud config feature and dynamic IP management of BOSH so that IPs are allocated on the fly and network calculation code is removed. In that case, services can be scaled up even if they need more than 8 IPs.

PR #123, opened 2017-12-06
Fixes issues #117 #118 #119 OOB Statuspoller fix (#11)

sauravmndl:
1. StatusPoller should cancel itself when NotFound error occurred form
  Status check API
2. Added UT to cover above scenario
3. Added UTs to cover resotre status check scenario
4. Patching metadata file of OOB backup with snapshotId when
   required
5. Discarding log from GET all backups of an OOB deployment.

### DIFF
--- a/lib/fabrik/OobBackupManager.js
+++ b/lib/fabrik/OobBackupManager.js
@@ -106,7 +106,8 @@ class OobBackupManager {
                     root_folder: CONST.FABRIK_OUT_OF_BAND_DEPLOYMENTS.ROOT_FOLDER_NAME
                   }, {
                     state: lastOperation.state,
-                    logs: logs
+                    logs: logs,
+                    snapshotId: lastOperation.snapshotId
                   })
                 );
             }
--- a/lib/jobs/OperationStatusPollerJob.js
+++ b/lib/jobs/OperationStatusPollerJob.js
@@ -8,6 +8,7 @@ const moment = require('moment');
 const CONST = require('../constants');
 const errors = require('../errors');
 const ScheduleManager = require('./ScheduleManager');
+const backupStore = require('../iaas').backupStoreForOob;
 const Promise = require('bluebird');
 
 class OperationStatusPollerJob extends BaseJob {
@@ -25,14 +26,26 @@ class OperationStatusPollerJob extends BaseJob {
       const msg = `Operation status poller cannot be initiated as the required mandatory params 
       (deployment_name | type | operation | operation_response.backup_guid) is empty : ${JSON.stringify(options)}`;
       logger.error(msg);
-      this.runFailed(new errors.BadRequest(msg), {}, job, done);
-      return;
+      return this.runFailed(new errors.BadRequest(msg), {}, job, done);
+    } else if (_.get(options, 'operation') !== 'backup' && _.get(options, 'operation') !== 'restore') {
+      const msg = `Operation pollinng not supported for operation - ${options.operation}`;
+      logger.error(msg);
+      const err = {
+        statusCode: `ERR_${options.operation.toUpperCase()}_NOT_SUPPORTED`,
+        statusMessage: msg
+      };
+      return this.runFailed(err, {}, job, done);
+    } else {
+      return this.checkOperationCompletionStatus(options.operation_response, job)
+        .then(operationStatusResponse => this.runSucceeded(operationStatusResponse, job, done))
+        .catch(err => {
+          logger.error(`Error occurred while running operation ${options.operation} status poller for deployment ${_.get(options, 'deployment_name')}.`, err);
+          return this.runFailed(err, {}, job, done);
+        });
     }
-
-    return this.checkOperationCompletionStatus(options.operation_response, job, done);
   }
 
-  static checkOperationCompletionStatus(operationResp, job, done) {
+  static checkOperationCompletionStatus(operationResp, job) {
     function isFinished(state) {
       return _.includes(['succeeded', 'failed', 'aborted'], state);
     }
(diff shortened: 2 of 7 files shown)


lib/fabrik/OobBackupManager.js
```diff
@@ -106,7 +106,8 @@ class OobBackupManager {
                     root_folder: CONST.FABRIK_OUT_OF_BAND_DEPLOYMENTS.ROOT_FOLDER_NAME
                   }, {
                     state: lastOperation.state,
-                    logs: logs
+                    logs: logs,
+                    snapshotId: lastOperation.snapshotId
```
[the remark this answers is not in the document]
jagadish-kb: guess it would still work when deleting because snapshotId would be coming as undefined, but still it would be a good idea to add this attrib only when the snapshotId is present.

lib/jobs/OperationStatusPollerJob.js
```diff
@@ -41,63 +54,71 @@ class OperationStatusPollerJob extends BaseJob {
...
+        } else if (operationName === 'restore') {
+          return this
+            .getBrokerClient()
+            .getDeploymentRestoreStatus(deploymentName, operationResp.token, boshDirectorName);
+        } else {
+          throw new errors.BadRequest(`Operation ${operationName} not supported by status poller.`);
+        }
+      })
+      .then(operationStatusResponse => {
+        if (isFinished(operationStatusResponse.state)) {
```
jagadish-kb: reuse method `utils/index.js-isServiceFabrikOperationFinished`

lib/jobs/OperationStatusPollerJob.js
```diff
@@ -41,63 +54,71 @@ class OperationStatusPollerJob extends BaseJob {
...
+          return ScheduleManager.cancelSchedule(`${deploymentName}_${operationName}_${backupGuid}`, CONST.JOB.OPERATION_STATUS_POLLER)
+            .then(() => {
+              logger.info(`Deployment ${deploymentName} ${operationName} for backup guid ${backupGuid} completed -`, operationStatusResponse);
+              return operationStatusResponse;
```
jagadish-kb: it would be a good idea to set the attrib `jobCancelled:true` in this case & can set this default value to `false`

lib/jobs/OperationStatusPollerJob.js
```diff
@@ -41,63 +54,71 @@ class OperationStatusPollerJob extends BaseJob {
...
-                  const err = {
-                    statusCode: `ERR_DEPLOYMENT_${operationName.toUpperCase()}_${operationStatusResponse.state.toUpperCase()}`,
-                    statusMessage: msg
-                  };
-                  this.runFailed(err, operationStatusResponse, job, done);
-                }
+                const msg = `Deployment ${deploymentName} ${operationName} with backup guid ${backupGuid} exceeding timeout time 
+                      ${config.backup.backup_restore_status_poller_timeout / 1000 / 60} (mins). Stopping status check`;
+                logger.error(msg);
+                return operationStatusResponse;
```
jagadish-kb: operationResponse must indicate the same status that the operation is being timed out. This is not being set in the same. Also, when the poller is timing out, it must abort the backup that might still be running in background and might finish eventually. Look at FabrikStatusPoller timeout [implementation](https://github.com/cloudfoundry-incubator/service-fabrik-broker/blob/master/lib/fabrik/FabrikStatusPoller.js#L81).

lib/jobs/OperationStatusPollerJob.js
```diff
@@ -41,63 +54,71 @@ class OperationStatusPollerJob extends BaseJob {
...
-                .then(() => {
-                  const msg = `Deployment ${deploymentName} ${operationName} with backup guid ${backupGuid} exceeding timeout time 
-                      ${config.backup.backup_restore_status_poller_timeout / 1000 / 60} (mins). Stopping status check`;
-                  logger.error(msg);
-                  const err = {
-                    statusCode: 'ERR_BACKUP_TIME_OUT',
-                    statusMessage: msg
-                  };
-                  return this.runFailed(err, operationStatusResponse, job, done);
+            return operationStatusResponse;
```
jagadish-kb: here set `job_cancelled:false` or it can set by default.

lib/jobs/OperationStatusPollerJob.js
jagadish-kb: you must invoke abort method on the agents on the deployment & also, update backup status to aborted in meta data. If you can update the sate of metadata to aborted irrespective of the response from the agents on the box, then you can hit the same API even when deployment is not found rather than patching the status as aborted seperately. So same abort implementation can be invoked from here as well as NotFound flow.

lib/fabrik/OobBackupManager.js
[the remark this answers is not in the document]
sauravmndl: We are doing JSON.sringify() which discards parameter having value undefined.  
Here: https://github.com/cloudfoundry-incubator/service-fabrik-broker/blob/master/lib/iaas/CloudProviderClient.js#L147 

Also we are doing same here
https://github.com/cloudfoundry-incubator/service-fabrik-broker/blob/master/lib/fabrik/DirectorManager.js#L687

In metadata snapshotId doesn't come when it is undefined.

lib/jobs/OperationStatusPollerJob.js
sauravmndl: done

lib/jobs/OperationStatusPollerJob.js
sauravmndl: Are you proposing to bring in abort also in OOBBackupManager ? This was not in plan.

lib/jobs/OperationStatusPollerJob.js
sauravmndl: done

lib/jobs/OperationStatusPollerJob.js
sauravmndl: done

lib/fabrik/OobBackupManager.js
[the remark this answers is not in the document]
sauravmndl: Created a ticket https://github.com/cloudfoundry-incubator/service-fabrik-broker/issues/124

Issue #125, opened 2017-12-08
Need to increase timeout duration for broker


pritishmishra:
For certain operations like DELETE on azure iaas, the call times out. This is because the broker follows a standard timeout duration of 120 seconds. This is also the default timeout duration set by the node.js framework.
https://stackoverflow.com/questions/23874954/is-there-a-default-timeout-in-nodejs-for-http-request
Fix:
We need to over-ride the timeout of nodejs framework and set it to a higher value.
We also need to increase the timeout duration in our timeout handler to fix this.
http://derpturkey.com/increase-request-timeout-for-express-4/

PR #126, opened 2017-12-08
Closes issue #125: Increasing timeout for service-fabrik-broker

pritishmishra:


pritishmishra:
Yes. @jagadish-kb Your question was correct. I tested this out. And CF waits only for 180 seconds. So, I am reverting my changes and raising it again by setting timeout to 175 seconds (a 5 sec buffer to be sure). I tested this and we timeout before CF for a long-running request and with the response of 503, as expected.

PR #129, opened 2017-12-12
Added addons as a non-foridden section in manifest

kamath-prasad:
Modified the code consider "addons" as a non-forbidden section while we check for outdated deployments .

### DIFF
--- a/lib/utils/index.js
+++ b/lib/utils/index.js
@@ -345,7 +345,8 @@ function hasChangesInForbiddenSections(diff) {
       'resource_pools',
       'networks',
       'properties',
-      'tags'
+      'tags',
+      'addons'
     ])
     .value();
   if (!_.isEmpty(forbiddenSections) && !_.includes(forbiddenSections, 'director_uuid')) {


kamath-prasad:
@jagadish-kb : 

Tested the following :

- Without mentioning explictly addons module in services manifest , but with addons module in boshruntime-config applicable for non service-fabrik service's deployments :

0. Existing deployment - a diff on BOSH , shows addons in the diff.
1. Existing deployment - after this fix , it didnt show addons as forbidden section change and did the update.
2. Existing deployment after update in #1 - subsequent diff on BOSH director using /diff API of BOSH , didnt return addons as a diff section which means after updating the deployment in #1 , director understands that addons need not be applied to this deployment.


- With mentioning explicitly addon's module in new version of service manifest (blueprint) for BOSH which does not have addons in runtime config :

0. Existing deployment without addons module in service-manifest - a diff on BOSH , shows new addon's in the manifest as the difference.
1. Existing deployment without addons module in service-manifest - after this fix , it didnt show addons as forbidden section change and did the update.
2. Existing deployment after update in #1 - subsequent diff on BOSH director using /diff API of BOSH , didnt return addons as a diff section which means after updating the deployment in #1 , director applied addons in the manifest of the deployment. ( the same was confirmed after downloading the manifest)


- With mentioning explicitly addon's module in new version of service manifest for BOSH which has addons in runtime config :

0. Existing deployment without addons module in service-manifest but created on the BOSH with addons runtime config - a diff on BOSH , didnt show new addon's in the manifest as the difference.
1. Existing deployment without addons module in service-manifest - after this fix , it didnt show addons as forbidden section change and did the update.
2. Existing deployment after update in #1 - subsequent diff on BOSH director using /diff API of BOSH , didnt return addons as a diff section which means after updating the deployment in #1 , director applied addons in the manifest of the deployment.

PR #130, opened 2017-12-13
Modified the code to skip only specific http status response codes from logging to riemann

kamath-prasad:
Modified the code to skip only specific http status response codes from logging to riemann instead of skipping all http 4xx response code as this leads to missing of some events related to forbidden update error.

Corresponding change in bosh release - https://github.com/cloudfoundry-incubator/service-fabrik-boshrelease/pull/34


jagadish-kb:
Apart from those two additional files, rest of changes looks good.

kamath-prasad:
@jagadish-kb : Removed the 2 un-necessary files.

PR #131, opened 2017-12-14
Adhering to OSBAPI specification

subhankarc:
**Adhering to OSBAPI specification**

Getting rid of dependency on space_guid and organization_guid in the create and update API.
Relying on the context parameter.
Changes done accordingly on the BackupStore implementation.

Extension API signatures still accepts space_guid and not changed. However, they are internally trannformed in
tenant_guid while using.

Test cases modified as well.


lib/controllers/ServiceFabrikAdminController.js
```diff
@@ -290,6 +290,7 @@ class ServiceFabrikAdminController extends BaseController {
       .assign(req.body)
       .set('user', req.user)
       .value();
+    options.tenant_guid = req.body.space_guid;
```
jagadish-kb: options would still have `space_guid` as its creation involves assigning attribs from `req.body` . Could be a good idea to change Line 290 to `.assign(_.omit(req.body, 'space_guid'))`. Doing so ensures that any downstream code relying on this will fail, else it might go unnoticed.

lib/controllers/ServiceFabrikApiController.js
```diff
@@ -406,6 +408,7 @@ class ServiceFabrikApiController extends BaseController {
       .pick('backup_guid')
       .assign(req.query)
       .value();
+    options.tenant_guid = req.query.space_guid;
```
jagadish-kb: same comment as [above](https://github.com/cloudfoundry-incubator/service-fabrik-broker/pull/131/files#r157120901)

lib/fabrik/DirectorInstance.js
```diff
@@ -73,9 +76,12 @@ class DirectorInstance extends BaseInstance {
       .try(() => {
         switch (operation.type) {
         case 'create':
-          return this
-            .createSecurityGroup(operation.space_guid)
-            .tap(() => operation.state === CONST.OPERATION.SUCCEEDED ? this.scheduleAutoUpdate() : {});
+          if (operation.context.platform === 'cloudfoundry') {
```
jagadish-kb: Instead of hardcoding platform specific processing with directorInstance like below, i propose the following. 
We would have class called `BaseCloudManager` which defines a basic methods / hooks like `preInstanceProvision`, `postInstanceProvision`, `preInstanceDelete`, `postInstanceDelete` etc... and then we can specific implementations of this baseclass/interface for CF & K8s as `CfCloudManager` & `K8sCloudManager`, which will have these platform specific implementations baked into them. 
Now when DirectorInstance is getting instantiated, it would be injected with the cloud specific manager into it & the `DirectorInstance` just invokes methods like `cloudManager.postInstanceProvision` or `cloudManager.postInstanceDelete` etc.. & so on. 
This keeps our code segregated, pluggable and modular even in our current state & this approach should not consume too much effort as well as the existing hardcoded implementations goes into specific `CloudManager` classes.

lib/fabrik/DirectorInstance.js
```diff
@@ -50,7 +50,10 @@ class DirectorInstance extends BaseInstance {
         this.networkSegmentIndex = networkSegmentIndex;
       })
       .tap(() => {
-        if (operation.type === 'delete') {
+        //TODO: This still has coupling with space guid as it calls ensureSpaceGuid. 
+        //As part of SF 2.0, we can go to ETCD and fetch the tenant_guid which will 
+        //ensure we dont call CF again. For now, we add a check for the feature validation.
+        if (operation.type === 'delete' && _.includes(this.manager.agent.features, 'backup')) {
```
jagadish-kb: This check seems incorrect. We want to delete security groups for all instances irrespective of whether they support backup or not. May be this check should go elsewhere. (maybe within esnureSpaceGuid)

lib/fabrik/DirectorInstance.js
```diff
@@ -73,9 +76,12 @@ class DirectorInstance extends BaseInstance {
       .try(() => {
         switch (operation.type) {
         case 'create':
-          return this
-            .createSecurityGroup(operation.space_guid)
-            .tap(() => operation.state === CONST.OPERATION.SUCCEEDED ? this.scheduleAutoUpdate() : {});
+          if (operation.context.platform === 'cloudfoundry') {
+            return this
+              .createSecurityGroup(operation.context.space_guid)
+              .tap(() => operation.state === CONST.OPERATION.SUCCEEDED ? this.scheduleAutoUpdate() : {});
```
jagadish-kb: K8s instances dont need to be auto-updated?

lib/fabrik/DirectorInstance.js
```diff
@@ -73,9 +76,12 @@ class DirectorInstance extends BaseInstance {
...
+          }
+          break;
         case 'update':
           return this.ensureSecurityGroupExists();
```
jagadish-kb: for update the platform specific check seems missing.

lib/fabrik/DirectorManager.js
```diff
@@ -566,17 +566,19 @@ class DirectorManager extends BaseManager {
         secret: undefined,
         agent_ip: undefined,
         started_at: backupStartedAt,
-        finished_at: null
+        finished_at: null,
+        tenant_guid: this.getTenantGuid(opts.context)
```
jagadish-kb: as part of this implementation then are we dropping `org_guid`? i was thinking in case of CF the tenant_guid would be `orgGuid_spaceGuid`. Is it a good idea to drop org guids?

lib/middleware.js
```diff
@@ -132,8 +132,11 @@ exports.checkQuota = function () {
     if (utils.isServiceFabrikOperation(req.body)) {
       logger.debug('[Quota]: ServiceFabrikOperation : calling next handler..');
       next();
+    } else if (req.body.context.platform !== 'cloudfoundry') {
```
jagadish-kb: Same as [above](https://github.com/cloudfoundry-incubator/service-fabrik-broker/pull/131/files#r157121565)

test/acceptance/service-fabrik-api.instances.director.spec.js
```diff
@@ -951,7 +971,7 @@ describe('service-fabrik-api', function () {
               mocks.verify();
             });
         });
-
+        /////
```
jagadish-kb: unnecessary line

test/acceptance/service-fabrik-api.instances.director.spec.js
```diff
@@ -875,7 +892,7 @@ describe('service-fabrik-api', function () {
               mocks.verify();
             });
         });
-
+        ///
```
jagadish-kb: Did this thing come by mistake?

test/acceptance/service-fabrik-api.instances.director.spec.js
```diff
@@ -522,7 +539,7 @@ describe('service-fabrik-api', function () {
...
```
jagadish-kb: unnecessary line

lib/controllers/ServiceFabrikAdminController.js
subhankarc: Agree...will make this change.

lib/controllers/ServiceFabrikApiController.js
subhankarc: 👍

lib/fabrik/DirectorInstance.js
subhankarc: Hi @jagadish-kb 

With SF 2.0 design, security group and other environment specific things will be executed as part of different controllers. Hence, I think different cloud specific Manager may not be necessary. In that case, all this code will do is trigger scheduleAutoUpdate.

I kept the code schedule autoupdate inside the if clause as at this moment K8S will not productively consume it. May be, I'll add this comment in the code. With SF2.0 it will be taken care of.

Does this sound good ?

lib/fabrik/DirectorInstance.js
subhankarc: Oh yes! Thanks. I am not adding it in ensureSpaceGuid, rather breaking promise.all and putting it in two chains.

Only problem is it will execute sequentially. put two codepath one having promise.all and one having only the deletesecgroup?

Dont want to do big change as with SF 2.0, ensurespaceguid would not be needed. Let me know what you suggest.

lib/fabrik/DirectorInstance.js
subhankarc: As stated above, kept it only for temporary purpose. With SF2.0, only scheduleautoupdate will be done here.

lib/fabrik/DirectorInstance.js
subhankarc: Missed it. Adding the if before the switch.

lib/fabrik/DirectorManager.js
subhankarc: Org_guid wasn't part of the tenant ID in case of backup metadata file. I kept it similar, as otherwise there will be migration effort. technically, spaceguid is unique for each space so IMHO, it serves the same purpose. Do you see a problem in this?

lib/middleware.js
subhankarc: This is the only place where in middleware we have used space_guid and org_guid. As stated above, other checks will go off with SF2.0. Hence I thought we can keep this check.

test/acceptance/service-fabrik-api.instances.director.spec.js
subhankarc: not sure how so many of them came. Removing them.

test/acceptance/service-fabrik-api.instances.director.spec.js
subhankarc: I guess so. Removing them.

test/acceptance/service-fabrik-api.instances.director.spec.js
subhankarc: Removing them.

lib/fabrik/DirectorInstance.js
jagadish-kb: If we are adding any checks like the ones that we have done in the above, am suggesting we refactor such code in to a separate file. We don't even have to define/implement all the above suggested methods, whatever is applicable forn now we just move it to this separate implementation class (plaformmanager subclass).
Also, whatever u refactor from this and put in this plaformmanager classes can directly then go into ur respective controllers when the SF 2.0 code comes in to place, we dont have refactor DirectorInstance code again later. The part of the code that is common (DirectorInstance) will still get executed under one controller and whatever is part of these platform specific cloud manager can go into individual controllers. 
Whatever ur doing here could be thrown away or if u write it with a bit more planning aid to our next steps. Also make the current code look clean till the 2.0 implementation comes into effect (which certainly can take about min 2 months ?).

subhankarc: As discussed with @jagadish-kb , we are not immediately puttin ghte platform checks in this PR. Security groups can be better handled as part of SF2.0 and hence we handle them there. For now, as CF is the only consumer, we dont check for the platform now.

lib/fabrik/DirectorInstance.js
```diff
@@ -50,32 +50,48 @@ class DirectorInstance extends BaseInstance {
...
             .all([
               this.deleteSecurityGroup(),
-              this.ensureSpaceGuid()
-              .then(space_guid => this.manager.deleteRestoreFile(space_guid, this.guid))
-              .catch(err => {
-                logger.error(`Failed to delete restore file of instance '${this.guid}'`);
-                logger.error(err);
-                throw err;
-              })
+              this.deleteRestoreFile()
```
subhankarc: @jagadish-kb Could you please see if this looks good ? Here we only check of backup feature for deletion of restore file in a separate method which does the feature check.

lib/fabrik/BaseManager.js
```diff
@@ -55,6 +55,14 @@ class BaseManager {
     );
   }
 
+  getTenantGuid(context) {
+    if (context.platform === 'cloudfoundry') {
+      return context.space_guid;
+    } else if (context.platform === 'kubernetes') {
```
sauravmndl: You can put 'cloudfoundry' and  'kubernetes' in constants as might be reused in other place of code base.

kamath-prasad: 👍

lib/middleware.js
```diff
@@ -147,6 +147,7 @@ exports.checkQuota = function () {
             logger.error(`[QUOTA] Quota is not sufficient for this request: org '${req.body.organization_guid}', service '${req.instance.service.name}', plan '${req.instance.plan.name}'`);
             next(new Forbidden(`Quota is not sufficient for this request`));
           } else {
+            console.log('Hello');
```
kamath-prasad: need to remove this line.

lib/routes/cf/v2.js
```diff
@@ -13,9 +13,9 @@ const instanceRouter = express.Router({
 
 /* Service Broker API Router */
 router.use(middleware.basicAuth(config.username, config.password));
-if (_.includes(['production', 'test'], process.env.NODE_ENV)) {
-  router.use(controller.handler('apiVersion'));
-}
+//if (_.includes(['production', 'test'], process.env.NODE_ENV)) {
```
kamath-prasad: why is this commented ?

subhankarc: The apiVersion check was skipped unless it is a production/test system. That fails the osb checker tools. Hence, I have enabled it.

lib/middleware.js
subhankarc: Ah. Sorry for that. :) I'll remove it.

lib/fabrik/BaseManager.js
subhankarc: Done

lib/fabrik/BasePlatformManager.js
```diff
@@ -0,0 +1,40 @@
...
+
+const errors = require('../errors');
+const NotImplementedBySubclass = errors.NotImplementedBySubclass;
+
+class BasePlatformManager {
+  constructor(guid, manager, context) {
+    this.guid = guid;
+    this.manager = manager;
+    this.context = context;
+    this.platform = context.platform;
```
jagadish-kb: seems redundant. Rather you can define a getter method and return the same.

lib/fabrik/CfPlatformManager.js
```diff
@@ -0,0 +1,104 @@
...
+  }
+
+  postInstanceDeleteOperations(options) {
+    /* jshint unused:false */
+    throw new NotImplemented('postInstanceDeleteOperations');
+  }
+
+  createSecurityGroup(space_guid, name, rules) {
+    logger.info(`Creating security group '${name}' with rules ...`, rules);
+    return utils
```
jagadish-kb: No point in having this class if still DockerInstance and DirectorInstance classes having code constructs which are CF specific. One can rename the method to `buildIpRules` but its still CF specific constructs that are being returned back & they will break for K8s. Its better to avoid this class all together if it cant be refactored the right way.

lib/fabrik/DirectorInstance.js
```diff
@@ -249,7 +291,7 @@ class DirectorInstance extends BaseInstance {
       .then(() => this.manager.deleteBinding(this.deploymentName, params.binding_id));
   }
 
-  buildSecurityGroupRules() {
+  buildIpRules() {
```
jagadish-kb: This should move out of DirectorInstance to PlatformManager class if we have such an implementation.

lib/fabrik/DirectorInstance.js
```diff
@@ -50,37 +67,62 @@ class DirectorInstance extends BaseInstance {
         this.networkSegmentIndex = networkSegmentIndex;
       })
       .tap(() => {
+        //TODO: This still has coupling with space guid as it calls ensureTenantId. 
+        //As part of SF 2.0, we can go to ETCD and fetch the tenant_id which will 
+        //ensure we dont call CF again. For now, we add a check for the feature validation.
```
jagadish-kb: Not sure if this commend is valid. Code is moved to platformManager. So is this comment still valid?

lib/fabrik/DirectorInstance.js
```diff
@@ -50,37 +67,62 @@ class DirectorInstance extends BaseInstance {
...
     return Promise
       .try(() => {
-        switch (operation.type) {
-        case 'create':
-          return this
-            .createSecurityGroup(operation.space_guid)
-            .tap(() => operation.state === CONST.OPERATION.SUCCEEDED ? this.scheduleAutoUpdate() : {});
-        case 'update':
-          return this.ensureSecurityGroupExists();
+        if (operation.context.platform === CONST.CONSUMING_ENVIRONMENT.CF) {
```
jagadish-kb: The intention of having platformmanager was to avoid such checks within Instance managers. If such checks cannot be avoided then better not to have platform manager, as makes code a bit difficult to read thru, because one would think and ask what would happen to K8S as part of post creation as the code omits it out.

lib/fabrik/DirectorInstance.js
```diff
@@ -5,21 +5,38 @@ const Promise = require('bluebird');
...
 
 class DirectorInstance extends BaseInstance {
-  constructor(guid, manager) {
+  constructor(guid, manager, platformManager) {
     super(guid, manager);
     this.networkSegmentIndex = undefined;
+    this.platformManager = !platformManager ? undefined : platformManager;
+  }
+
+  assignPlatformManager() {
```
jagadish-kb: code smell. Why is this `assignPlatformManager` being called from outside? This class's constructor is also passed with platformmanager as a param. Code flow of assigning platform manager seems convoluted.

Rather i would expect the context to be always present & set in req object. If context is absent, you fetch it either from Deployment or Docker environment and ensure its present. Once its existence is guaranteed in flow, just pass it as a param to Instance's constructor method. The respective instances can themselves call `BasePlatformManager.getInstance` and get an instance. 

This way of assigning platformmanager either via constructor or via this explicit method seems redundant. In all calls of POST/UPDATE context will always be present so even in such scenarios unnecessarily calls will be made to BOSH which could've been avoided.

lib/fabrik/DirectorManager.js
```diff
@@ -194,9 +194,7 @@ class DirectorManager extends BaseManager {
             .getDeploymentManifest(deploymentName)
             .then(manifest => _
               .assign(opts, {
-                previous_manifest: manifest,
-                organization_guid: _.get(previousValues, 'organization_id'),
-                space_guid: _.get(previousValues, 'space_id')
+                previous_manifest: manifest
```
jagadish-kb: You seem to have removed orgid & space id.  If you are removing org id and space id, always ensure context is passed so that consumers (service manifests) can pick the same from context. You are passing it as part of tags, but explicitly passing in context & calling it as context makes sense from readability perspective of services manifest, as it does not seem right that one picks up org id from tags. This way it remains consistent across all consumers. Everyone lookups context object to get what they want, rather than looking up tags in some place and context in other.

lib/fabrik/DockerInstance.js
```diff
@@ -7,6 +7,9 @@ const errors = require('../errors');
 const utils = require('../utils');
 const docker = require('../docker');
 const BaseInstance = require('./BaseInstance');
+const CONST = require('../constants');
+const CfPlatformManager = require('./CfPlatformManager');
```
jagadish-kb: DockerInstance referring both platformManagers? It breaks the purpose of having platform managers.

lib/fabrik/Fabrik.js
```diff
@@ -31,12 +34,17 @@ class Fabrik {
...
 
-  static createInstance(instance_id, service_id, plan_id) {
+  static createInstance(instance_id, service_id, plan_id, context) {
     const plan = catalog.getPlan(plan_id);
     assert.strictEqual(service_id, plan.service.id);
     return this
       .createManager(plan)
-      .then(manager => manager.createInstance(instance_id));
+      .then(manager => {
+        const platformManager = !context ? undefined : (context.platform === CONST.CONSUMING_ENVIRONMENT.CF ?
```
jagadish-kb: I would probably move this to BasePlatformManager and ask it to give an instance from its static method.

lib/fabrik/DirectorInstance.js
subhankarc: This method should be part of BaseController IMO. The code 

```
return this.inspectContainer()
 .then(() => JSON.parse(this.getEnvironment().context))
```
and

```
this.manager.director.getDeploymentNameForInstanceId(this.guid)
      .then(deploymentName => this.manager.director.getDeploymentProperty(deploymentName, 'platform-context'))

```

can be part of specific instance methods which can be called from there.

lib/fabrik/DirectorInstance.js
```diff
@@ -50,37 +67,62 @@ class DirectorInstance extends BaseInstance {
         this.networkSegmentIndex = networkSegmentIndex;
       })
       .tap(() => {
+        //TODO: This still has coupling with space guid as it calls ensureTenantId.
```
subhankarc: Please remove this comment. It is obsolete now.

lib/controllers/ServiceFabrikApiController.js
```diff
@@ -111,19 +111,32 @@ class ServiceFabrikApiController extends BaseController {
...
+        const platform = (req.body ? (req.body.context ? req.body.context.platform : undefined) : undefined) ||
+          req.query.platform;
+        if (platform !== undefined) {
+          _.set(req, 'entity.platform', platform);
+        }
+        /*Following statement for backward compatibility*/
+        const tenant_id = req.body.space_guid || req.query.space_guid ||
+          (platform ? (req.query.tenant_id || req.body.context.space_guid || req.body.context.namespace) : undefined);
+        if (tenant_id) {
+          if (!BaseController.uuidPattern.test(tenant_id) &&
```
subhankarc: I think this check is for the fact that cloudfoundry would have space_guid in GUID format and K8S will have it as name of the namespace, not guid. But in that case, if the tenant_id is from CF, we dont need to mtch it with k8s and vice versa.

lib/reports/BackupReportManager.js
```diff
@@ -134,7 +134,7 @@ class BackupReportManager {
         'data.instance_name': 1,
         'data.plan_id': 1,
         'data.service_plan_name': 1,
-        'data.space_guid': 1,
+        'data.tenant_id': 1,
```
amitmalav: Can you please change the following to `tenant_id`
https://github.infra.hana.ondemand.com/cloudfoundry/service-fabrik-product-component/blob/master/scripts/reports/generate-backup-report#L118
and
https://github.infra.hana.ondemand.com/cloudfoundry/service-fabrik-product-component/blob/master/scripts/reports/generate-backup-report#L101

lib/controllers/ServiceFabrikApiController.js
```diff
@@ -111,19 +111,32 @@ class ServiceFabrikApiController extends BaseController {
...
-          this.validateUuid(space_guid, 'Space GUID');
-          return space_guid;
+        /* Following statement to address cross consumption scenario*/
+        const platform = (req.body ? (req.body.context ? req.body.context.platform : undefined) : undefined) ||
+          req.query.platform;
+        if (platform !== undefined) {
+          _.set(req, 'entity.platform', platform);
+        }
+        /*Following statement for backward compatibility*/
+        const tenant_id = req.body.space_guid || req.query.space_guid ||
```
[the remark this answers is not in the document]
sauravmndl: Will put in _.get
Need suggestion for other place. In whole code base, in other places these kinds of param fetch is written in above way. how about them? 
And,  `unnecessarily these lookups will happen` --> exactly for this I put the context based code at the last, as we are not changing in the code of integration tests (where we call these APIs), will still be sending space_guid/ (most of the cases not sent at all), not the 'context'. This method needs to be refactored again when K8S consumption comes in place :)

lib/controllers/ServiceFabrikApiController.js
```diff
@@ -111,19 +111,32 @@ class ServiceFabrikApiController extends BaseController {
...
+          (platform ? (req.query.tenant_id || req.body.context.space_guid || req.body.context.namespace) : undefined);
+        if (tenant_id) {
+          if (!BaseController.uuidPattern.test(tenant_id) &&
+            !BaseController.k8sNamespacePattern.test(tenant_id)) {
```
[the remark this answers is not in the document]
sauravmndl: Agree. The lower piece of code should be moved to platform manager. I have put the code here for now. As when K8S consumption becomes active, we need to do some amount of refactoring that time as well. Planned for the concrete changes together that time. Is that okay?

lib/controllers/BaseController.js
```diff
@@ -99,6 +101,16 @@ class BaseController {
       .throw(new ContinueWithNext());
   }
 
+  validateRequest(req, res) {
+    if (req.instance.async && ((!req.query) || (!req.query.accepts_incomplete) ||
+        req.query.accepts_incomplete !== 'true')) {
+      res.status(422).send({
```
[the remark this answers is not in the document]
sauravmndl: sure. would use value from constant.

lib/controllers/BaseController.js
```diff
@@ -99,6 +101,16 @@ class BaseController {
       .throw(new ContinueWithNext());
   }
 
+  validateRequest(req, res) {
+    if (req.instance.async && ((!req.query) || (!req.query.accepts_incomplete) ||
```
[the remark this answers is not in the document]
sauravmndl: done

lib/controllers/ServiceFabrikApiController.js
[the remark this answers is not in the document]
sauravmndl: done.

lib/fabrik/CfPlatformManager.js
sauravmndl: 'buildIpRules' --> should be refactored and it should only send the params to build the rule (in cf) or network policy (in K8S). Want to avoid further segregation of docker and director related code platform manager. Yes, agree the buildIpRules method now send the rule (fits only for K8S) but will send ips and  ports in future. As part of dynamic ip feature we will take that up (I remember we have decided this.)

lib/fabrik/DirectorInstance.js
sauravmndl: Please comment on https://github.com/cloudfoundry-incubator/service-fabrik-broker/pull/131#discussion_r167478083

lib/fabrik/DirectorInstance.js
sauravmndl: Removed

lib/fabrik/DirectorInstance.js
sauravmndl: sorry. forgot to remove this. This is not at all required :P

lib/fabrik/DirectorInstance.js
sauravmndl: 'In all calls of POST/UPDATE context will always be present so even in such scenarios unnecessarily calls will be made to BOSH which could've been avoided.' ==> whenever context is present in request body, not extra call is made to BOSH or Docker. Platform manager is built directly based on the value of context.platform.

sauravmndl: 'If context is absent, you fetch it either from Deployment or Docker environment and ensure its present. Once its existence is guaranteed in flow, just pass it as a param to Instance's constructor method.' ==> To determine the docker environment we need to instantiate DocketInstance itself first. So, that's why defined the method to fetch context details.

lib/fabrik/DockerInstance.js
sauravmndl: It is to instantiate the classes but this.platformManager having the actual value.

lib/fabrik/Fabrik.js
sauravmndl: Let's decide on resolving this https://github.com/cloudfoundry-incubator/service-fabrik-broker/pull/131#discussion_r167481063 first. Then we can decide about instantiation of platform manager.

lib/fabrik/CfPlatformManager.js
```diff
@@ -0,0 +1,104 @@
...
+
+  postInstanceProvisionOperations(options) {
+    return this.createSecurityGroup(this.space_guid, options.ruleId, options.ipRules);
+  }
+
+  preInstanceDeleteOperations(options) {
+    return this.deleteSecurityGroup(options.ruleId);
+  }
+
+  postInstanceDeleteOperations(options) {
```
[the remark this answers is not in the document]
sauravmndl: Agree. would remove them.

lib/fabrik/CfPlatformManager.js
```diff
@@ -0,0 +1,104 @@
...
+const ordinals = ['First', 'Second', 'Third', 'Fourth', 'Fifth', 'Sixth', 'Seventh', 'Eighth', 'Ninth', 'Tenth'];
+
+class CfPlatformManager extends BasePlatformManager {
+  constructor(guid, manager, context) {
+    super(guid, manager, context);
+    this.space_guid = context.space_guid;
+    this.cloudController = cloudController;
+  }
+
+  preInstanceProvisionOperations(options) {
```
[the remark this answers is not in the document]
sauravmndl: Agree. would remove them. We can keep adding later.

lib/fabrik/BasePlatformManager.js
```diff
@@ -0,0 +1,40 @@
...
+  }
+
+  preInstanceProvisionOperations(options) {
+    /* jshint unused:false */
+    throw new NotImplementedBySubclass('preInstanceProvisionOperations');
+  }
+
+  postInstanceProvisionOperations(options) {
```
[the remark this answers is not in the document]
sauravmndl: Yes. But for all Base* classes we have defined like this.

lib/fabrik/DirectorInstance.js
```diff
@@ -50,37 +67,62 @@ class DirectorInstance extends BaseInstance {
...
             ]);
         }
       });
   }
 
+  deleteRestoreFile() {
+    Promise.try(() => {
+      if (_.includes(this.manager.agent.features, 'backup')) {
+        return this.platformManager.ensureTenantId()
+          .then(space_guid => this.manager.deleteRestoreFile(space_guid, this.guid))
```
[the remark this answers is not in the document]
sauravmndl: Sure sure. Would add it back. Don't know how got removed. My bad!!

lib/fabrik/DirectorInstance.js
```diff
@@ -50,37 +67,62 @@ class DirectorInstance extends BaseInstance {
...
-        case 'create':
-          return this
-            .createSecurityGroup(operation.space_guid)
-            .tap(() => operation.state === CONST.OPERATION.SUCCEEDED ? this.scheduleAutoUpdate() : {});
-        case 'update':
-          return this.ensureSecurityGroupExists();
+        if (operation.context.platform === CONST.CONSUMING_ENVIRONMENT.CF) {
+          switch (operation.type) {
+          case 'create':
+            //With SF 2.0 design, security group and other environment specific things
```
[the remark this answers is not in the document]
sauravmndl: Done

lib/controllers/BaseController.js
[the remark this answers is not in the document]
sauravmndl: done

lib/fabrik/DirectorInstance.js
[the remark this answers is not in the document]
sauravmndl: yes. Removed.

lib/fabrik/DirectorInstance.js
```diff
@@ -5,21 +5,38 @@ const Promise = require('bluebird');
 const _ = require('lodash');
 const yaml = require('js-yaml');
 const BaseInstance = require('./BaseInstance');
+const CfPlatformManager = require('./CfPlatformManager');
```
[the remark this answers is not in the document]
sauravmndl: This is to decide which platformManager needs to set. But I might refactor instantiation of PlatformManager further :). see https://github.com/cloudfoundry-incubator/service-fabrik-broker/pull/131#discussion_r167481063 and below comments, for more details.

lib/fabrik/DirectorInstance.js
sauravmndl: sure

lib/fabrik/DirectorInstance.js
sauravmndl: PlatformManager instantiation might be changed.

lib/reports/BackupReportManager.js
sauravmndl: done

lib/fabrik/DirectorInstance.js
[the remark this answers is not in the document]
sauravmndl: if (_.includes(this.manager.agent.features, 'backup')) { ====> is the check?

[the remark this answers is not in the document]
subhankarc: Yes!

lib/fabrik/CfPlatformManager.js
jagadish-kb: This still does not seem right :-) lemme try something hands on later.

lib/fabrik/CfPlatformManager.js
```diff
@@ -0,0 +1,103 @@
...
+      .then(securityGroup => securityGroup.metadata.guid)
+      .tap(guid => logger.info(`+-> Created security group with guid '${guid}'`))
+      .catch(err => {
+        logger.error(`+-> Failed to create security group ${name}`);
+        logger.error(err);
+        throw new SecurityGroupNotCreated(name);
+      });
+  }
+
+  ensureSecurityGroupExists(space_guid, name) {
```
jagadish-kb: platformanager has space_guid already set, so you want to revisit this method signature?

lib/fabrik/DockerInstance.js
```diff
@@ -48,17 +48,26 @@ class DockerInstance extends BaseInstance {
...
+      context: params.context
+    };
     return this
       .buildContainerOptions(parameters, exposedPorts, options)
       .then(opts => this.createAndStartContainer(opts, true))
       .catchThrow(DockerError.Conflict, new ServiceInstanceAlreadyExists(this.guid))
       .then(() => this.ensureContainerIsRunning(true))
-      .then(() => this.createSecurityGroup(space_guid));
+      .then(() => this.platformManager.postInstanceProvisionOperations({
+        ruleId: this.securityGroupName,
```
jagadish-kb: atleast the securitygroup name can be directly within platform manager right? Why should this class care about this name? Move all code related to security group name to platform manager

lib/fabrik/DockerInstance.js
```diff
@@ -82,7 +91,9 @@ class DockerInstance extends BaseInstance {
   delete(params) {
     /* jshint unused:false */
     return this
-      .deleteSecurityGroup()
+      .platformManager.preInstanceDeleteOperations({
+        ruleId: this.securityGroupName
```
jagadish-kb: security group name can be within platform manager. no need to pass this here.

lib/fabrik/Fabrik.js
```diff
@@ -31,12 +34,29 @@ class Fabrik {
...
-  static createInstance(instance_id, service_id, plan_id) {
+  static createInstance(instance_id, service_id, plan_id, context) {
     const plan = catalog.getPlan(plan_id);
     assert.strictEqual(service_id, plan.service.id);
     return this
       .createManager(plan)
-      .then(manager => manager.createInstance(instance_id));
+      .then(manager => {
+        const instance = manager.createInstance(instance_id);
+        return Promise
```
jagadish-kb: something still does not seem right in this block. Will need to try this hands on later in the day tomorrow.

lib/fabrik/DirectorInstance.js
```diff
@@ -53,34 +57,52 @@ class DirectorInstance extends BaseInstance {
...
   finalize(operation) {
     const action = _.capitalize(operation.type);
     return Promise
       .try(() => {
         switch (operation.type) {
         case 'create':
-          return this
-            .createSecurityGroup(operation.space_guid)
+          return this.platformManager.postInstanceProvisionOperations({
+              ruleId: this.securityGroupName,
```
jagadish-kb: rule id - security group name , can be within platform manager.

lib/constants.js
```diff
@@ -167,6 +167,11 @@ module.exports = Object.freeze({
       DESC: -1
     }
   },
+
+  CONSUMING_ENVIRONMENT: {
```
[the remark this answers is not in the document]
sauravmndl: This  name is I haven't decided. Inherited. Would change to PLATFORM.

lib/fabrik/DirectorInstance.js
```diff
@@ -12,16 +12,20 @@ const jwt = require('../jwt');
...
 class DirectorInstance extends BaseInstance {
-  constructor(guid, manager) {
-    super(guid, manager);
+  constructor(guid, manager, platformManager) {
+    super(guid, manager, platformManager);
     this.networkSegmentIndex = undefined;
   }
 
+  get platformContext() {
+    return this.manager.director.getDeploymentNameForInstanceId(this.guid)
```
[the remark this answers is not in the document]
sauravmndl: getDeploymentName need 'networkSegmentIndex' which is not populated when platformContext() method is called!!

lib/fabrik/DirectorInstance.js
```diff
@@ -53,34 +57,52 @@ class DirectorInstance extends BaseInstance {
...
         switch (operation.type) {
         case 'create':
-          return this
-            .createSecurityGroup(operation.space_guid)
+          return this.platformManager.postInstanceProvisionOperations({
+              ruleId: this.securityGroupName,
+              ipRules: this.buildIpRules()
+            })
+            .then(() => this.manager.director
+              .createDeploymentProperty(this.deploymentName, 'platform-context', JSON.stringify(operation.context))
```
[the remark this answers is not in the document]
sauravmndl: .Thanks !! moved up before creating security groups.

lib/fabrik/DockerInstance.js
sauravmndl: Removed.

lib/fabrik/DockerInstance.js
sauravmndl: Removed

lib/fabrik/Fabrik.js
sauravmndl: sure. But we don't want to definitely fork the platform manager to have docker and director related stuff further unless it is absolute necessary.

lib/fabrik/DirectorInstance.js
sauravmndl: removed

lib/controllers/BaseController.js
```diff
@@ -99,6 +101,16 @@ class BaseController {
       .throw(new ContinueWithNext());
   }
 
+  validateRequest(req, res) {
+    if (req.instance.async && ((!req.query) ||
```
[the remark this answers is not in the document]
sauravmndl: done

lib/fabrik/DockerInstance.js
sauravmndl: Thought of platformManager to be generic and methods would be independent. As someone might need to create a security group of any name, in that case let caller decide which name it should pass. Platform manager should be unaware of that, it will just create. similarly for delete and ensure security group methods also.

lib/fabrik/Fabrik.js
jagadish-kb: If you argue that there should not be further bifurcation based on docker/director then instead of these new constructs of platform manager, we could have directly injected `CloudControllerClient` or `K8SControllerClient` (when its added) into directorInstance.
The idea of having these platform managers are they help in abstracting any platform specific operations within themselves. In the process if they need to communicate with BOSH/DOCKER they should be able to communicate & build the required data.

sauravmndl: Then BOSH/DOCKER specific operations are done both in DirectorInstance/DockerInstance and 
CfPlatformManager/K8sPlatformManager. Then I would suggest let's re-look at design of these service layer class segregation. In that case we should not call the new classes as '*PlatformManager', something else. Anyway DockerInstance-->DokcerManager methods placing are not not in-line with DirectorInstance-->DockerManager. Let's sync up.

lib/fabrik/DirectorManager.js
sauravmndl: Added opts.context in tags, but it is not sufficient for getSchedule update api. 

but whye 
`
-                organization_guid: _.get(previousValues, 'organization_id'),	
-                space_guid: _.get(previousValues, 'space_id'
`required. Needs to be verified.

amitmalav: org_guid and space_guid has to be passed, as generateManifest is being called from diffManifest also.
Or in generateManifest, we can check if `opts.context` is null then assign tags to org_guid and space_guid
https://github.com/cloudfoundry-incubator/service-fabrik-broker/blob/84702b2fc2c2cf86706fcd1c88c612810a41d172/lib/jobs/ServiceInstanceUpdateJob.js#L121

sauravmndl: That part, I would take care.
How about  following?
'
organization_guid: _.get(previousValues, 'organization_id'),	
space_guid: _.get(previousValues, 'space_id'
'

lib/fabrik/BasePlatformManager.js
```diff
@@ -0,0 +1,44 @@
...
+class BasePlatformManager {
+  constructor(guid, context) {
+    this.guid = guid;
+    this.context = context;
+  }
+
+  get platform() {
+    return this.context.platform;
+  }
+  preInstanceProvisionOperations(options) {
```
jagadish-kb: similar to this void method do you think you have missed out `postInstanceDeleteOperations`, `preInstanceUpdateOperations` methods?

lib/fabrik/DirectorInstance.js
```diff
@@ -249,12 +288,21 @@ class DirectorInstance extends BaseInstance {
       .then(() => this.manager.deleteBinding(this.deploymentName, params.binding_id));
   }
 
-  buildSecurityGroupRules() {
+  buildIpRules() {
+
+    function getPorts(first, last) {
```
jagadish-kb: unnecessary method.

lib/fabrik/DirectorInstance.js
```diff
@@ -249,12 +288,21 @@ class DirectorInstance extends BaseInstance {
...
+      return ports;
+    }
+
     return _.map(this.manager.getNetwork(this.networkSegmentIndex), net => {
       return {
         protocol: 'tcp',
-        destination: `${_.first(net.static)}-${_.last(net.static)}`,
-        ports: '1024-65535'
+        ips: net.static,
+        ports: getPorts(1024, 65535)
```
jagadish-kb: instead you could have written `ports: [1024, 65535]`. You are anyways considering first and last. So why populate such a huge array. If your problem is that semantically its not right, then you set attrib `port` - for a single port in docker & `port_range` with value `1024-65535` so that way you pick based on port or port_range and you can avoid this huge ports array.

lib/fabrik/DirectorManager.js
```diff
@@ -545,7 +530,7 @@ class DirectorManager extends BaseManager {
     const backupStartedAt = new Date().toISOString();
     const data = _
       .chain(opts)
-      .pick('service_id', 'plan_id', 'instance_guid', 'organization_guid', 'space_guid', 'username')
+      .pick('service_id', 'plan_id', 'organization_guid', 'space_guid', 'instance_guid', 'username')
```
jagadish-kb: Do you want to drop `space_guid` in the above?

lib/fabrik/DirectorInstance.js
sauravmndl: will set [1024, 65535]

lib/fabrik/BasePlatformManager.js
sauravmndl: Not doing any operation in postInstanceDeleteOperations, preInstanceUpdateOperations.

lib/fabrik/DockerInstance.js
```diff
@@ -48,17 +50,54 @@ class DockerInstance extends BaseInstance {
     return false;
   }
 
+  get platformContext() {
+    return this.inspectContainer()
+      .then(() => {
+        if (this.getEnvironment().context) {
+          return JSON.parse(this.getEnvironment().context);
+        } else {
+          throw new NotFound(`platform-context not found in docker environemt for instance ${this.guid}`);
```
jagadish-kb: isnt it simple to just resolve the default value in else, rather than throwing `not found` and then resolving default value.
Also the err clause of NOT FOUND. Are you sure you want to do that, because thats the error from inspectContainer Call which might return back with container not found and for that you are returning default context value, which to me seems wrong. So you should not have any error handler but just send default value in else clause.

lib/fabrik/DirectorInstance.js
sauravmndl: will remove as setting [1024, 65535] in ports.

lib/fabrik/Fabrik.js
```diff
@@ -31,12 +32,27 @@ class Fabrik {
...
+          .try(() => {
```
jagadish-kb: can covert all of this try section to the below oneline
```
return Promise.try(() => context ? context : instance.platformContext)
....
```

lib/fabrik/DockerInstance.js
sauravmndl: In order to send default value we need to handle not found, as the 'context'  is not present for older containers.  So if we don't handle error in this case it will fail the operations where platform doesn't send context e.g. unbind and delete.

And in case where container not found itself then (even though context is set to default) in 'removeContainer' method itself it will thrwo error (https://github.com/cloudfoundry-incubator/service-fabrik-broker/blob/master/lib/fabrik/DockerInstance.js#L234).

lib/fabrik/Fabrik.js
sauravmndl: sure. would do that.

lib/fabrik/Fabrik.js
```diff
@@ -31,12 +32,27 @@ class Fabrik {
...
+            if (!context) {
+              return instance.platformContext;
+            } else {
+              return context;
+            }
+          })
+          .then(context => {
+            const platformManager = BasePlatformManager.getInstance(instance_id, context);
+            instance.assignPlatformManager(platformManager);
```
jagadish-kb: Just from beautification the same block of code can look like this
```
const instance = manager.createInstance(instance_id);
return Promise.try(() => context ? context : instance.platformContext)
            .then(context => instance.assignPlatformManager(BasePlatformManager.getInstance(instance_id, context))
            .return(instance);
```

lib/fabrik/VirtualHostInstance.js
```diff
@@ -5,10 +5,11 @@ const bosh = require('../bosh');
 const utils = require('../utils');
 const mapper = require('./VirtualHostRelationMapper');
 const catalog = require('../models').catalog;
+const CONST = require('../constants');
 
 class VirtualHostInstance extends BaseInstance {
-  constructor(guid, manager) {
-    super(guid, manager);
+  constructor(guid, manager, platformManager) {
```
jagadish-kb: constructor still has platform manager?

lib/fabrik/VirtualHostInstance.js
```diff
@@ -22,6 +23,12 @@ class VirtualHostInstance extends BaseInstance {
       .then(deploymentName => this.deploymentName = deploymentName);
   }
 
+  get platformContext() {
+    return Promise.resolve({
```
jagadish-kb: wrong. Should get the parent instance id and from there can get the deployment name and from the deployment can figure out the context. THis is incorrect.

lib/fabrik/DirectorManager.js
sauravmndl: Removed

lib/fabrik/VirtualHostInstance.js
sauravmndl: changed

lib/fabrik/VirtualHostInstance.js
sauravmndl: done

lib/fabrik/Fabrik.js
sauravmndl: Done. Okay with return part. But Breaking in two lines adds more ease in readability. that's why put. But made changes as per suggestion.

lib/fabrik/VirtualHostInstance.js
sauravmndl: The virtual host related change we haven't done as it has cloud controller specific operations and we have decided not to do changes. That's why was returning default value. Now I remember.

sauravmndl: Actually returning no ops is ideal.

lib/controllers/BaseController.js
```diff
@@ -99,6 +103,15 @@ class BaseController {
       .throw(new ContinueWithNext());
   }
 
+  validateRequest(req, res) {
```
subhankarc: @sauravmndl As discussed earlier, can you also add the check for mandatory space_guid and organization_guid in here?
Something like this.

```
    if (!_.get(req.body, 'space_guid') || !_.get(req.body, 'organization_guid')) {
      res.status(CONST.HTTP_STATUS_CODE.BAD_REQUEST).send({
        error: 'BadRequest',
        description: 'This request is missing mandatory organization guid and space guid.'
      });
    }
```

lib/fabrik/BasePlatformManager.js
subhankarc: Let's keep methods only which are currently used. We can add later if anything is needed.

lib/controllers/BaseController.js
sauravmndl: Done

lib/fabrik/DockerInstance.js
```diff
@@ -48,17 +50,49 @@ class DockerInstance extends BaseInstance {
     return false;
   }
 
+  get platformContext() {
+    return this.inspectContainer()
+      .then(() => {
+        if (this.getEnvironment().context) {
```
jagadish-kb: _.get(this.getEnvironment, 'context')  !== undefined)

PR #133, opened 2017-12-18
Feature/bosh2

pritishmishra:
1. Migrating to bosh v2 schema
2. Introducing 'provides' and 'consumes' for jobs
3. Moving respective properties to respective jobs
4. Providing shared properties for dependent jobs and consuming from these shared properties.
related to https://github.com/cloudfoundry-incubator/service-fabrik-broker/issues/11


lib/bosh/BoshDirectorClient.js
```diff
@@ -432,7 +434,16 @@ class BoshDirectorClient extends HttpClient {
         }
       })
       .then(manifest => {
-        return manifest.properties.agent;
+        if (manifest.instance_groups) {
+          let agentJob = {};
+          _.each(manifest.instance_groups, (instance_group) => {
+            agentJob = _.find(instance_group.jobs, job => job.name === 'broker-agent');
```
kamath-prasad: can you store 'broker-agent' in constants ?

lib/bosh/manifest/Networks.js
```diff
@@ -22,14 +22,15 @@ class Networks {
         const networkSegmentation = network.segmentation || defaultSegmentationOptions;
         const nets = network.subnets.map(subnet => {
           const net = Network.create(_.assign({
-            name: `${network.name}_${subnet.az}`,
+            name: network.name,
```
kamath-prasad: why the values are interchanged ?

lib/bosh/BoshDirectorClient.js
pritishmishra: ok..

lib/bosh/manifest/Networks.js
pritishmishra: that was how we were referring in bosh1 .. i.e. via subnet name, e.g. 3 names for 3 subnets
and in bosh2 we are referring to the network name, and then subsequently to the subnet name.
Here, network name could be 'sf_bosh_services' and subsequent 3 subnets can have 3 subnet names (with the same network name).

config/settings.yml
```diff
@@ -357,6 +358,7 @@ defaults: &defaults
     lock_deployment_max_duration:  86400 # lock ttl for deployments in seconds. Its kept at 24 hrs - 60 * 60 * 24
     infrastructure: &director_infrastructure
       stemcell:
+        alias: service_fabrik_stemcell
```
jagadish-kb: many things in the setting.yml which are IAAS specific can be removed. ex. azs, disk_types, vm_types & compilaton. 
Just keep `networks` as we might need it. But rest all things IAAS specific should be removed.

config/settings.yml
```diff
@@ -821,6 +823,8 @@ boshlite2:
   - <<: *director
     url: https://192.168.50.6:25555
     password: v5o4z3qg3og1gjct0eea
+    primary: true
```
jagadish-kb: better blank the password in the above.

config/settings.yml
jagadish-kb: stemcells are common across landscape. Dont prefix the name as service-fabrik_stemcell. Rather can be something like 'ubunutu-trusty' [just check whatever is used in cf core.]

config/settings.yml
```diff
@@ -837,7 +841,64 @@ boshlite2:
...
+  - <<: *director
+    prefix: service-fabrik-azure-dev1
+    url: https://127.0.0.1:35555
+    password: admin
+    primary: true
+    support_create: true
+    infrastructure:
+      <<: *director_infrastructure
+      stemcell:
+        alias: service_fabrik_stemcell
```
jagadish-kb: stemcell alias need not be sf specific

config/templates/blueprint-manifest.yml.ejs
```diff
@@ -34,21 +30,36 @@ if (!spec.previous_manifest) {
     }
   };
 } else {
-  /* update scenario: preserve previous service instance properties */
-  properties.blueprint = spec.previous_manifest.properties.blueprint;
+  if(!spec.previous_manifest.instance_groups){
+    properties.blueprint = spec.previous_manifest.properties.blueprint;
+  }else{
+    properties.blueprint = spec.previous_manifest.instance_groups[0].jobs[0].properties;
```
jagadish-kb: In your testing didnt you find this was to be named as `spec.previous_manifest.instance_groups[0].jobs[0].properties.blueprint` - ? Just checking

lib/bosh/BoshDirectorClient.js
```diff
@@ -38,9 +39,9 @@ class BoshDirectorClient extends HttpClient {
   clearCache(config) {
```
jagadish-kb: can this method also be renamed to `clearConfigCache`, since the variable is also named thus?

lib/bosh/BoshDirectorClient.js
```diff
@@ -72,7 +73,7 @@ class BoshDirectorClient extends HttpClient {
   }
 
   populateCache() {
-    logger.debug('Loading Bosh DeploymentName cache... current cached deployments:', _.keys(this.cache));
+    logger.debug('Loading Bosh DeploymentName cache... current cached deployments:', _.keys(this.boshConfigCache));
```
jagadish-kb: same as above - rename to `populateConfigCache`

lib/bosh/BoshDirectorClient.js
```diff
@@ -134,20 +135,20 @@ class BoshDirectorClient extends HttpClient {
   }
 
   updateCacheEntry(deployment, config) {
-    const directorConfig = this.cache[deployment];
+    const directorConfig = this.boshConfigCache[deployment];
```
jagadish-kb: same as above - rename to `updateConfigCacheEntry`

lib/bosh/BoshDirectorClient.js
```diff
@@ -432,7 +434,16 @@ class BoshDirectorClient extends HttpClient {
...
-        return manifest.properties.agent;
+        if (manifest.instance_groups) {
+          let agentJob = {};
+          _.each(manifest.instance_groups, (instance_group) => {
+            agentJob = _.find(instance_group.jobs, job => job.name === 'broker-agent');
+            return !agentJob;
+          });
+          return agentJob.properties;
+        } else {
+          return manifest.properties.agent;
```
jagadish-kb: put a comment here that this is to handle backward compatibility and a TODO: to remove this once all instances are migrated. Else down (say 3 months) down the line this becomes confusing.

lib/bosh/manifest/Networks.js
jagadish-kb: @kamath-prasad in V1 each of the subnet themselves were called as network and their names had suffix of `az`. As part of v2, manifest only refer to the network name as defined in cloud config. Now subnet name is nowhere required but just in future if someone wants a subnet name as well he can refer to this attrib and subnet names are suffixed with az

lib/fabrik/DirectorManager.js
```diff
@@ -352,6 +341,7 @@ class DirectorManager extends BaseManager {
       name: deploymentName,
       director_uuid: this.director.uuid,
       releases: this.releases,
+      stemcells: [this.stemcell],
       compilation: BoshDirectorClient.getInfrastructure().compilation,
```
jagadish-kb: remove unnecessary things that are not used in Header.js like compilation, disk_pools, resource_pools

config/settings.yml
pritishmishra: ok.. I will be renaming the alias to 'ubuntu-trusty'

config/templates/blueprint-manifest.yml.ejs
pritishmishra: that was a mistake in the test implementation. I fixed it.
In bosh v2 manifest schema, we don't use the level 'blueprint' and place properties under it, the way we were doing in the bosh v1 schema. I had missed that part in my test mocks.

pritishmishra:
Implemented review comments. Kindly review.

config/templates/blueprint-manifest.yml.ejs
jagadish-kb: ok.

PR #140, opened 2018-01-06
Restore across service plans

muralisc:
With this change Service Fabrik should be able to restore the backup of
one plan across plans of the same instance. Only backups taken from
plans specified in the restore_predecessor section of the current plan
can be restored. If the restore_predecessor section is absent in the
plan update_predecessor is checked.

* Broker script changes for working with bosh-lite: Added `--post302`
  which is necessary to query broker when the url has redirection.
* Added restore_predecessors check for update operation.
* Removed plan_id from backup and restore file name.
* Unit and acceptance tests.

### DIFF
--- a/bin/broker
+++ b/bin/broker
@@ -1,6 +1,6 @@
 #!/bin/bash -eu
 
-broker_host="10.0.2.2:9292"
+broker_host="127.0.0.1:9292"
 instance_guid=""
 plan_id=""
 service_id=""
@@ -11,6 +11,7 @@ auth_header=""
 function init() {
   # service instance name
   local name="${1}"
+  [[ -z $name ]] && echo "Service name absent" && exit
   # instance
   local instance=$(cf curl "/v2/service_instances?q=name:${name}" | jq -r .resources[0])
   instance_guid=$(echo ${instance} | jq -r .metadata.guid)
@@ -33,6 +34,10 @@ function init_without_instance() {
   auth_header=$(cat ~/.cf/config.json | jq -r .AccessToken)
 }
 
+function info() {
+  curl "http://${broker_host}/api/v1/info" -L -k
+}
+
 ########## GENERAL
 
 function service_instance_state() {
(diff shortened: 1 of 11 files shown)


test/fabrik.DirectorManager.spec.js
```diff
@@ -55,6 +58,26 @@ describe('fabrik', function () {
         manager.findNetworkSegmentIndex(used_guid).then(res => expect(res).to.eql(21));
       });
     });
-
+    describe('#isRestorePossible', function () {
+      it('should return false when plan not in restore_predecessors', function () {
+        // restore not possible from small to xsmall
+        manager = new DirectorManager(catalog.getPlan(xsmall_plan_id));
+        manager.update_predecessors = [];
+        killJsHintWarning = expect(manager.isRestorePossible(small_plan_id)).to.be.false;
```
kamath-prasad: what is this variable ?

`killJsHintWarning` ?

lib/controllers/ServiceFabrikApiController.js
```diff
@@ -296,7 +295,7 @@ class ServiceFabrikApiController extends BaseController {
         if (metadata.state !== 'succeeded') {
           throw new UnprocessableEntity(`Can not restore backup '${backup_guid}' due to state '${metadata.state}'`);
         }
-        if (metadata.plan_id !== plan_id) {
+        if (!req.manager.isRestorePossible(metadata.plan_id)) {
           throw new UnprocessableEntity(`Backup '${backup_guid}' to restore from is not applicable to the instance's service plan`);
```
jagadish-kb: sentence in the error does not sound right gramatically. Can you please check.

muralisc: Updated

test/fabrik.DirectorManager.spec.js
muralisc: Jshint expects a assignment or function call in each statement and not an expression.
The expect check for true/false is a expression and not a function call like other expect statements.
So `killJsHintWarning` is used to make the expression into and assignment. 
Also renamed `killJsHintWarning` to `return_value`.

kamath-prasad: Ok , got it .

bin/broker
```diff
@@ -1,6 +1,6 @@
 #!/bin/bash -eu
```
subhankarc: 👍 Thanks for fixing this!

muralisc:
https://github.com/cloudfoundry-incubator/service-fabrik-broker/issues/146

PR #143, opened 2018-01-09
Fixes #142 added code to skip quota check when configured in plan metadata

### LINKED ISSUE #142: Quota check should be disabled when configured in plan
**Description**
Quota check should be disabled when configured in plan

**If you are running service fabrik locally than Output of `node -v 
for node version and grep version package.json for service fabrik version`:**

```

node version: v6.9.1
service fabrik: 3.0.1
```

sauravmndl:
added code to skip quota check when configured in plan metadata to fix #142

### DIFF
--- a/test/quota.QuotaManager.spec.js
+++ b/test/quota.QuotaManager.spec.js
@@ -173,6 +173,7 @@ describe('quota', () => {
       const smallPlanId = 'bc158c9a-7934-401e-94ab-057082a5073e';
       const xsmallPlanId = 'bc158c9a-7934-401e-94ab-057082a5073f';
       const smallPlanGuid = 'c77ff479-ea63-42ea-b6d1-d824c4012f1c';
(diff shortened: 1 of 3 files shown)


test/quota.QuotaManager.spec.js
```diff
@@ -173,6 +173,7 @@ describe('quota', () => {
       const smallPlanId = 'bc158c9a-7934-401e-94ab-057082a5073e';
       const xsmallPlanId = 'bc158c9a-7934-401e-94ab-057082a5073f';
       const smallPlanGuid = 'c77ff479-ea63-42ea-b6d1-d824c4012f1c';
+      const rabbitmqVirtualHostPlanGuid = 'd035f948-5d3a-43d7-9aec-954e134c3e9d';
```
kamath-prasad: better to rename the variable and have a name independent of rabbitmq.

`skipQuotaCheckPlanGuid` ?

also just to understand , this plan is not defined in the test ? like it does not have metadata attribute defined ?

Issue #144, opened 2018-01-10
Replace the values for AWS keys with dummy values


ashishjain14:
In the file https://github.com/cloudfoundry-incubator/service-fabrik-broker/blob/1c89ab53d6be883cdd2213ab53df8a15f00b6ad4/config/settings.yml#L72 we have AWS credentials such as key and keyId. Similarly here https://github.com/cloudfoundry-incubator/service-fabrik-broker/blob/master/config/settings.yml#L898. These values are unusable however let us change these to dummy values to avoid any confusion.

PR #149, opened 2018-01-11
masked the invalid credentials from settings

kamath-prasad:
Closes - https://github.com/cloudfoundry-incubator/service-fabrik-broker/issues/144


muralisc:
what about this
https://github.com/kamath-prasad/service-fabrik-broker/blob/5152bdee8fda56b6dd593b525ec7ea9a5dbadcd5/config/settings.yml#L121

Issue #157, opened 2018-01-20
virtualhost service instance deletion failing


pranjaljain:
**Description**
virtualhost service instance deletion is failing because entry in `deploymentIpsCache` used in `BoshDirectorClient` is not cleaned when a deployment is deleted.

**Steps to reproduce the issue:**
1. Create a rabbitmq service instance of plan v3.6-xsmall.
2. Now create instance of plan virtualhost on it.
3. Delete the parent rabbitmq service instance.
4. Now, delete the instances of plan virtualhost. It will fail.

```
node version: v8.5.0
service fabrik: 3.0.1
```

PR #159, opened 2018-01-20
virtualhost service instance deletion is failing because entry in deploymentIpsCache used in BoshDirectorClient is not cleaned

pranjaljain:
Closes-Bug: https://github.com/cloudfoundry-incubator/service-fabrik-broker/issues/157
Description: virtualhost service instance deletion is failing because entry in deploymentIpsCache used in BoshDirectorClient is not cleaned when a deployment is deleted.
This fix removes the entry from deploymentIpsCache when delete deployment is called, as similarly done in the code when createOrUpdate is called [here](https://github.com/cloudfoundry-incubator/service-fabrik-broker/blob/master/lib/bosh/BoshDirectorClient.js#L350).

### DIFF
--- a/lib/bosh/BoshDirectorClient.js
+++ b/lib/bosh/BoshDirectorClient.js
@@ -383,6 +383,7 @@ class BoshDirectorClient extends HttpClient {
   }
 
   deleteDeployment(deploymentName) {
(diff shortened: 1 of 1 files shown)


pritishmishra:
this part was missed by us.. 
the PR looks fine to me..

Issue #161, opened 2018-01-24
Update README to use custom settings for development


muralisc:
**Description**

Update README.md to use custom settings.yml file for development to prevent unwanted checkins of development changes.

PR #162, opened 2018-01-24
Update readme to use custom settings.yml

muralisc:
Issue: #161

### DIFF
--- a/README.md
+++ b/README.md
@@ -144,7 +144,8 @@ If you need to change the `settings.yml` configuration you should copy the file 
 ```shell
 # env vars you may like to set to different than these default values
 # export NODE_ENV=development ## For bosh2.0, use the environment boshlite2, as the passwords and BOSH IP are different.
-# export SETTINGS_PATH=$(pwd)/config/settings.yml
+# cp $(pwd)/config/settings.yml $(pwd)/config/my-settings.yml
+# export SETTINGS_PATH=$(pwd)/config/my-settings.yml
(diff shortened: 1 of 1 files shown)


muralisc:
@subhankarc 
But this will make updating setting.yml very cumbersome.
https://www.atlassian.com/git/tutorials/gitignore#ignoring-a-previously-committed
go ahead with this ?

jagadish-kb:
@subhankarc @muralisc making settings.yml file as part of .gitignore list only makes the whole proceess cumbersome and also it does not guarantee accidental checkin of sensitive info. My recommendation is that every member of the team religiously follows the existing guideline of using the settings file of the format 'my-settings.yml'. 
The other option which can be done in the long run is to extract all sensitive info into a seperate file called `credentials.yml` and then we can gitignore this file. Properties related to credentials are something that will be only added once in a while and sample attrib names can be added and force commit for this ignore file can be done occassionally.

muralisc:
https://github.com/cloudfoundry-incubator/service-fabrik-broker/pull/167
In addition to above removed the default settings.yml from service-fabrik-broker so the environment variable has to be set for local development.

kamath-prasad:
@subhankarc  :  Can you pls review/approve this PR as well ?

subhankarc:
Hi @jagadish-kb ,

Just to force the developers to use custom settings, also suggested @mualisc to remove the default settings yaml code. 

I think this should make it complete.

Issue #156, opened 2018-01-15
Add support for configurable additional content in service instance dashboard 


amitmalav:
**Description**
In service instance dashboard, Pre defined key-value pairs are present inside dashboard. Currently no support available, If a service owner wants to add some fields specific to the service.
https://github.com/cloudfoundry-incubator/service-fabrik-broker/blob/master/views/mixins/sections.pug#L11

```
node version: v6.9.1
service fabrik: 3.0.1
```

PR #163, opened 2018-01-29
Add dashboard template

amitmalav:
closes: #156 
* Remove dependency on swarm manager
* Add configurable director dashboard

### DIFF
--- a/config/settings.yml
+++ b/config/settings.yml
@@ -494,6 +494,7 @@ defaults: &defaults
           image: 'servicefabrikjenkins/blueprint'
           tag: 'latest'
           memory: '128m'
+          dashboard_template_path: <%= dashboard_template_path('blueprint') %>
           persistent_volumes:
           - name: 'data'
             path: '/data'
@@ -561,6 +562,7 @@ defaults: &defaults
           prefix: 'blueprint'
           agent: *agent_blueprint
           template: <%= base64_template('blueprint') %>
+          dashboard_template_path: <%= dashboard_template_path('blueprint') %>
           releases: &dedicated_releases
           - { name: blueprint, version: 1.0.0, git: "https://github.com/sap/service-fabrik-blueprint-boshrelease.git" }
           context: &context_plan3
--- a/config/settings.yml
+++ b/config/settings.yml
@@ -103,6 +103,7 @@ defaults: &defaults
   colorize_log: false
   sys_log_level: info
   enable_circuit_breaker: true
+  enable_swarm_manager: true
   feature:
     ServiceInstanceAutoUpdate: true #Switch to turn on / turn off schedule_update feature
   # this timeout is set to 175 secs by default because CF has a default timeout of 180 secs
--- a/config/settings.yml
+++ b/config/settings.yml
@@ -494,7 +494,6 @@ defaults: &defaults
           image: 'servicefabrikjenkins/blueprint'
           tag: 'latest'
           memory: '128m'
-          dashboard_template_path: <%= dashboard_template_path('blueprint') %>
           persistent_volumes:
           - name: 'data'
             path: '/data'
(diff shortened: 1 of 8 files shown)


config/templates/blueprint-dashboard.yml.ejs
```diff
@@ -0,0 +1,6 @@
+<%
+const plan_manager = plan.manager.name;
+%>
+
+plan:
+  Manager: <%= plan_manager %>
```
kamath-prasad: why caps ?

amitmalav: This key will get displayed on dashboard as it is. 
https://github.com/cloudfoundry-incubator/service-fabrik-broker/pull/163/files#diff-fda1a509b3b7827775e98e4fcec2a0feR11

kamath-prasad: ok , got it. thanks .

views/mixins/sections.pug
```diff
@@ -1,34 +1,34 @@
...
+mixin showSectionService(service, plan, instance, options)
   section#service.section
     .container
       .row
         .col-lg-12
           h4 Service Information
       .row
         .col-lg-6
-           +renderPropertyListPanel('Service', {
+           +renderPropertyListPanel('Service', _.assign({
```
jagadish-kb: you are passing this additional options parameter, but the mixin defintion `renderPropertyListPanel` isnt changed. So how will this value `options.service` be used? Also, isnt `options` an HTML fragment which is to be just embedded? Looks like you are using it as a JSON.

Also it would help if you can checkin a sample blueprint_dashboard template sample as well so that we can see the exact flow play out here.

config/settings.yml
```diff
@@ -559,7 +560,8 @@ defaults: &defaults
         settings: &settings_plan3
           prefix: 'blueprint'
           agent: *agent_blueprint
-          template: <%= base64_template('blueprint') %>
+          template: <%= base64_template('blueprint', 'manifest') %>
```
jagadish-kb: why confuse with this 'type' parameter. Type then must be enumerated. Why not pass the full file name like 'blueprint-manifest' or 'blueprint-dashboard'. Also this is to be included conditionally, as not all services will have dashboard template. Need to raise PR even on service-fabrik-product component with sample for dashboard.

amitmalav: PR on sfpc is raised. There also i have added an example for blueprint dashboard.

views/mixins/sections.pug
amitmalav: `options` is a JSON object. In `renderPropertyListPanel` we pass title of panel and a json object which is then displayed in tabular format.
https://github.com/cloudfoundry-incubator/service-fabrik-broker/blob/master/views/mixins/controls.pug#L54
You can find a sample `blueprint-dashboard.yml.ejs`  file in pr

jagadish-kb: Ok. Got it, but as discussed let us also evaluate and try to put in an option of embedding an HTML snippet as well from service providers.

lib/controllers/DashboardController.js
```diff
@@ -28,6 +29,13 @@ class DashboardController extends BaseController {
         info = _.assign({
           userId: req.session.user_id,
         }, info);
+        let additional_info = {};
+        if (req.instance.plan.manager.settings.dashboard_template) {
+          additional_info = yaml.safeLoad(_.template(Buffer.from(req.instance.plan.manager.settings.dashboard_template, 'base64'))(info));
+        }
+        info = _.assign({
+          options: additional_info
```
jagadish-kb: userId: req.session.user_id -- add this directly here.

jagadish-kb: rename `options` to something more meaningful like `service_custom_attrs` or `custom_attrs`

views/mixins/sections.pug
jagadish-kb: ok. Got it.

views/mixins/controls.pug
```diff
@@ -40,6 +40,8 @@ mixin renderPropertyListValue(value)
       td= _.first(value)
   else if /^https?:\/\//.test(value)
     td: a(href=value)= value
+  else if /^adt:\/\//.test(value)
```
subhankarc: Can we make this check more generic?

anything with `://` can be treated as hyperlink.

adt is more SAP specific.

views/dashboard-director.pug
```diff
@@ -6,5 +6,5 @@ block navbar
 
 block content
   include mixins/sections.pug
-  +showSectionService(service, plan, instance)
+  +showSectionService(service, plan, instance, curstomAttrs)
```
subhankarc: spelling mistake! `curstomAttrs` should be `customAttrs`

PR #164, opened 2018-01-30
Fix/forbidden manifest

pritishmishra:
Modifying the approach of finding forbidden manifest changes, as per the bosh2 manifest changes

Related to https://github.com/cloudfoundry-incubator/service-fabrik-broker/issues/11

### DIFF
--- a/lib/utils/index.js
+++ b/lib/utils/index.js
@@ -341,17 +341,34 @@ function hasChangesInForbiddenSections(diff) {
     .map(line => _.nth(/^([a-z]\w+):/.exec(line), 1))
     .difference([
       'update',
-      'compilation',
       'releases',
-      'resource_pools',
-      'networks',
-      'properties',
       'tags',
       'addons'
     ])
     .value();
+  function isJobRemoved(){
+    var jobsRegex = new RegExp('^  jobs'); // this regex is to find the position of jobs section
+    var jobsLevelRegex = new RegExp('^  [a-z]+'); // this regex is to find the position of next section at the same level as jobs
+    var jobNameRegex = new RegExp('^  [- ] name'); // this regex is to find the position of removal in job name
+    var jobStartIndex = _.findIndex(diff, element => {
+      return jobsRegex.test(element[0]);
+    });
+    var jobEndIndex = _.findIndex(diff, element => {
+      return jobsLevelRegex.test(element[0]);
+    }, jobStartIndex+1);
+    var jobNameRemovedIndex = _.findIndex(diff, element => {
+      return jobNameRegex.test(element[0]) && _.includes(element[1], 'removed');
+    });
+    if (jobNameRemovedIndex > jobStartIndex && (jobEndIndex === -1 || jobNameRemovedIndex < jobEndIndex)){
+      return true;
+    } // this checks if the name removal is present within the jobs section only.
+    return false;
+  }
   if (!_.isEmpty(forbiddenSections) && !_.includes(forbiddenSections, 'director_uuid')) {
-    throw new errors.Forbidden(`Automatic update not possible. Detected changes in forbidden section(s) '${forbiddenSections.join(',')}'`);
+    const forbiddenSectionsDiff = _.filter(diff, element => _.includes(element[0], 'instances') || _.includes(element[0], 'persistent_disk_type'));
+    if (!_.isEmpty(forbiddenSectionsDiff) || isJobRemoved()){
+      throw new errors.Forbidden(`Automatic update not possible. Detected changes in forbidden section(s) '${forbiddenSections.join(',')}'`);
+    }
   }
   return false;
 }
--- a/lib/utils/index.js
+++ b/lib/utils/index.js
@@ -363,7 +363,7 @@ function hasChangesInForbiddenSections(diff) {
     const forbiddenSectionsDiff = _.filter(diff, element => _.includes(element[0], 'instances') || _.includes(element[0], 'persistent_disk_type'));
     const removedJobName = findRemovedJob();
     if (!_.isEmpty(forbiddenSectionsDiff) || removedJobName) {
-      throw new errors.Forbidden(`Automatic update not possible. ${!_.isEmpty(forbiddenSectionsDiff)? 'Detected changes in forbidden sections:' + forbiddenSectionsDiff.join(','): `Job definition removed: ${removedJobName}`}`);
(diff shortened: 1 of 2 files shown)


lib/utils/index.js
```diff
@@ -341,17 +341,37 @@ function hasChangesInForbiddenSections(diff) {
...
     .value();
+
+  function isJobRemoved() {
+    const jobsRegex = new RegExp('^  jobs'); // this regex is to find the position of jobs section
+    const jobsLevelRegex = new RegExp('^  [a-z]+'); // this regex is to find the position of next section at the same level as jobs
+    const jobNameRegex = new RegExp('^  [- ] name'); // this regex is to find the position of removal in job name
+    const jobStartIndex = _.findIndex(diff, element => {
+      return jobsRegex.test(element[0]);
+    });
+    const jobEndIndex = _.findIndex(diff, element => {
```
jagadish-kb: similarly above can be written like below. If the method has only one statement and return, then follow the below concise way.
```
const jobEndIndex = _.findIndex(diff, jobsLevelRegex.test(element[0]), jobStartIndex + 1);
```

lib/utils/index.js
```diff
@@ -341,17 +341,37 @@ function hasChangesInForbiddenSections(diff) {
...
+      return jobsLevelRegex.test(element[0]);
+    }, jobStartIndex + 1);
+
+    const jobDiff = _.slice(diff, jobStartIndex, jobEndIndex);
+    const jobNameRemovedIndex = _.findIndex(jobDiff, element => {
```
jagadish-kb: same as above

lib/utils/index.js
```diff
@@ -341,17 +341,37 @@ function hasChangesInForbiddenSections(diff) {
...
+    });
+    if (jobNameRemovedIndex !== -1) {
+      return true;
+    } // this checks if the name removal is present within the jobs section only.
+    return false;
+  }
   if (!_.isEmpty(forbiddenSections) && !_.includes(forbiddenSections, 'director_uuid')) {
-    throw new errors.Forbidden(`Automatic update not possible. Detected changes in forbidden section(s) '${forbiddenSections.join(',')}'`);
+    const forbiddenSectionsDiff = _.filter(diff, element => _.includes(element[0], 'instances') || _.includes(element[0], 'persistent_disk_type'));
+    if (!_.isEmpty(forbiddenSectionsDiff) || isJobRemoved()) {
```
jagadish-kb: you are not passing any args to the function `isJobRemoved()` how do you pass `diff` to it? Has this passed UT?

test/jobs.ServiceInstanceUpdateJob.spec.js
```diff
@@ -236,10 +236,10 @@ describe('Jobs', function () {
         organization_guid: organization_guid
       });
       const diff = [
-        ['jobs:', null],
-        ['- name: blueprint_z1', null],
-        ['  instances: 2', 'removed'],
-        ['  instances: 1', 'added']
+        ['instance_groups:', null],
```
jagadish-kb: ADD UT for the new section of Code written. Exclusively test with all the changes ex., instance group, persitence disk, job removal.

lib/utils/index.js
```diff
@@ -341,17 +341,37 @@ function hasChangesInForbiddenSections(diff) {
...
+      throw new errors.Forbidden(`Automatic update not possible. Detected changes in forbidden section(s) '${forbiddenSections.join(',')}'`);
```
jagadish-kb: change the error message here to reflect the actual changes. for ex, it should reflect the values from `forbiddenSectionsDiff` and not from `forbiddenSections`

lib/utils/index.js
pritishmishra: `diff` is already a local variable to the function `hasChangesInForbiddenSections` and the function `isJobRemoved()` is present within this function. So, do I require to pass `diff` again?

test/jobs.ServiceInstanceUpdateJob.spec.js
pritishmishra: ok.. will add another test case which tests the job removal. Persistence disk type is exactly the same as instances, the call flow is same too. So, adding a test case for job removal which will test the newly added functions.

lib/utils/index.js
pritishmishra: ok.. I will change the message

lib/utils/index.js
pritishmishra: ok

lib/utils/index.js
pritishmishra: ok

lib/utils/index.js
```diff
@@ -341,17 +341,37 @@ function hasChangesInForbiddenSections(diff) {
...
       'tags',
       'addons'
     ])
     .value();
+
+  function isJobRemoved() {
+    const jobsRegex = new RegExp('^  jobs'); // this regex is to find the position of jobs section
+    const jobsLevelRegex = new RegExp('^  [a-z]+'); // this regex is to find the position of next section at the same level as jobs
+    const jobNameRegex = new RegExp('^  [- ] name'); // this regex is to find the position of removal in job name
+    const jobStartIndex = _.findIndex(diff, element => {
```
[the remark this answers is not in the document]
pritishmishra: ok

lib/utils/index.js
```diff
@@ -334,24 +334,37 @@ function isServiceFabrikOperationFinished(state) {
...
     ])
     .value();
+
   if (!_.isEmpty(forbiddenSections) && !_.includes(forbiddenSections, 'director_uuid')) {
-    throw new errors.Forbidden(`Automatic update not possible. Detected changes in forbidden section(s) '${forbiddenSections.join(',')}'`);
+    const forbiddenSectionsDiff = _.filter(diff, element => _.includes(element[0], 'instances') || _.includes(element[0], 'persistent_disk_type'));
+    const removedJobName = findRemovedJob();
+    if (!_.isEmpty(forbiddenSectionsDiff) || removedJobName) {
+      throw new errors.Forbidden(`Automatic update not possible. ${!_.isEmpty(forbiddenSectionsDiff)? 'Detected changes in forbidden sections:' + forbiddenSectionsDiff.join(','): `Job definition removed: ${removedJobName}`}`);
+    }
```
jagadish-kb: make this ${removedJobName[0]}

PR #167, opened 2018-01-30
Remove default settings file

muralisc:


kamath-prasad:
@muralisc / @subhankarc  : We do not have such settings.yml for boshrelease repo , correct ?

subhankarc:
In bosh release we export it here https://github.com/cloudfoundry-incubator/service-fabrik-boshrelease/blob/master/jobs/service-fabrik-broker/templates/bin/job_properties.sh.erb

So it should be ok. Good to go!

PR #177, opened 2018-02-12
Modified the code to fix the issue while parsing timestamp in BackupStore which is used to list all backups

kamath-prasad:
Fixes https://github.com/cloudfoundry-incubator/service-fabrik-broker/issues/178

### DIFF
--- a/lib/iaas/BackupStore.js
+++ b/lib/iaas/BackupStore.js
@@ -291,7 +291,10 @@ class Filename {
       .zipObject(path.basename(filename).split('.'))
       .assign(data)
       .tap(filenameObject => {
-        if (filenameObject.started_at && this.keys.exclude_file_with_root_value && this.keys.exclude_file_with_root_value !== _.get(filenameObject, this.root)) {
+        if (this.keys.exclude_file_with_root_value !== undefined && this.keys.exclude_file_with_root_value === _.get(filenameObject, this.root)) {
+          return;
+        }
+        if (filenameObject.started_at && this.isValidTimeStamp(filenameObject.started_at)) {
           filenameObject.started_at = this.parseTimestamp(filenameObject.started_at);
         }
       })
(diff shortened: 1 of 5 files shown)


lib/iaas/BackupStore.js
```diff
@@ -291,7 +291,10 @@ class Filename {
       .zipObject(path.basename(filename).split('.'))
       .assign(data)
       .tap(filenameObject => {
-        if (filenameObject.started_at) {
+        if (this.keys.exclude_file_with_root_value !== undefined && this.keys.exclude_file_with_root_value === _.get(filenameObject, this.root)) {
```
jagadish-kb: zipObject would fail if `path.basename(filename).split('.).length !== this.keys.length` so may be this could be used as a condition rather than having this check.

Also, if such a check is done then OOB backups will be excluded from being cleaned by BackupReaper. Backupreaper must clean up both OOB and instances scheduled backups!.

kamath-prasad: when you say , zipObject would fail , you mean some of the key's will have value as `undefined` correct ? It would not error out as seen now . 

`path.basename(filename).split('.).length !== this.keys.length` - this does not guarantee that we will filter out OOB backups right ? Since in the future if the keys in BackupStoreForOOB = keys in BackupStoreForServiceInstance , then this check will be in-affective ?!

> Also, if such a check is done then OOB backups will be excluded from being cleaned by BackupReaper. Backupreaper must clean up both OOB and instances scheduled backups!.

Backup reaper job even now doesnt process OOB backups , correct ? Will check this,

lib/iaas/BackupStore.js
```diff
@@ -319,6 +322,15 @@ class Filename {
     );
   }
 
+  /**
+   * 
+   * Checks if the input timestamp is in the following format : yyyy-mm-ddTHH-mm-ssZ
+   **/
+  isValidTimeStamp(timestamp) {
+    var timeStampRegex = new RegExp(`^[0-9]{4}-(0[1-9]|1[012])-(0[1-9]|[12][0-9]|3[01])T(0[0-9]|1[0-9]|2[0-4])-([0-5][0-9])-([0-5][0-9])Z$`);
```
jagadish-kb: use const not 'var'. rather you could also use it like `/^[0-9]{4}-(0[1-9]|1[012])-(0[1-9]|[12][0-9]|3[01])T(0[0-9]|1[0-9]|2[0-4])-([0-5][0-9])-([0-5][0-9])Z$/.test(timestamp)`

kamath-prasad: Done .

lib/iaas/BackupStore.js
kamath-prasad: Created an issue for handling backup reaper job use case for OOB backups - https://github.com/cloudfoundry-incubator/service-fabrik-broker/issues/190 . Will send it in another PR.

jagadish-kb: ok

PR #179, opened 2018-02-13
Deprecating a Service Plan

aishwaryjoshi94:
- Introduce metadata.state: deprecated in service catalogue to deprecate a plan
- Blocking creation of the deprecated plan
- Other CF operation remain unaffected
- Fixes https://github.com/cloudfoundry-incubator/service-fabrik-broker/issues/183


kamath-prasad:
@aishwaryjoshi94  : Can format the code and submit again , travis validation has failed - https://travis-ci.org/cloudfoundry-incubator/service-fabrik-broker/builds/340885651?utm_source=github_status&utm_medium=notification

lib/middleware.js
```diff
@@ -207,10 +208,26 @@ exports.requireEventLogging = function (appConfig, appType) {
...
 }
 
 function formatContentSecurityPolicyDirective(values, key) {
   return `${key} ${values.join(' ')};`;
-}
\ No newline at end of file
+}
+function checkIfPlanDeprecated(plan_id){
+ const plan_state=_.get(catalog.getPlan(plan_id), 'metadata.state','active');
+ if (plan_state == 'deprecated')
```
kamath-prasad: you can directly do 

`return plan_state == 'deprecated';`

ketakig12: Can you also define enum in constant.js for possible states of the plan and then use it here instead of hardcoded strings?

PR #191, opened 2018-02-22
Fixing Services and Plan organization for boshlite2

subhankarc:
- Refering to the same default services

- Removing services settings from the development and boshlite2


subhankarc:
Sure! Created this https://github.com/cloudfoundry-incubator/service-fabrik-broker/issues/192

PR #198, opened 2018-02-22
Fixes for OSB API Changes

sauravmndl:
1. For non-cf platforms avoiding deleting restore file at the time of deletion of service instance
2. Parsing context from BOSH property
3. UT mock fix


lib/fabrik/DirectorInstance.js
```diff
@@ -76,17 +79,15 @@ class DirectorInstance extends BaseInstance {
...
-        return this.platformManager.ensureTenantId()
-          .then(space_guid => this.manager.deleteRestoreFile(space_guid, this.guid))
-          .catch(err => {
-            logger.error(`Failed to delete restore file of instance '${this.guid}'`);
-            logger.error(err);
-            throw err;
-          });
-      }
-    });
+    if (_.includes(this.manager.agent.features, 'backup')) {
```
kamath-prasad: @sauravmndl : how are we ensuring here that only for CF platform restore files are deleted ? Is it based on the tenant_id check below or based on this if condition ?

sauravmndl: the tenant_id below.

kamath-prasad: ok , got it . Looks fine to me then.

PR #207, opened 2018-02-26
Catalog handling to filter plans/services in case of cross consumption

subhankarc:
* Parameterized the path that gets registered as broker.
* Added settings for supported_platform and osbapi_compliant_name
* Catalogs are shown based on the supported_platform
* Changed implementation for PlatformManagers, instance assignment is removed from it.
* Added new test for K8S catalog
* Fixed Quota test which was changing the catalog
* Removed the unnecessary CC API version check

Also, see https://github.com/openservicebrokerapi/servicebroker/issues/442 and https://github.com/openservicebrokerapi/servicebroker/pull/452

### DIFF
--- a/config/settings.yml
+++ b/config/settings.yml
@@ -432,6 +432,9 @@ defaults: &defaults
       secret: 'service-fabrik-blueprint-secret'
       redirect_uri: <%= redirect_uri %>
     plan_updateable: true
+    supported_platform:
+    - k8s
+    - cf
     plans:
     ###########################
     # CONTAINER SERVICE PLANS #
@@ -565,6 +568,10 @@ defaults: &defaults
         - '1 vCPUs'
         - '1 GB Memory'
         - '1 GB Disk'
+      supported_platform:
+      - k8s
+      - cf
+      osbapi_compliant_name: 'v10-xsmall'
     - &blueprint_plan4
       id: 'bc158c9a-7934-401e-94ab-057082a5073e'
       name: 'v1.0-small'
@@ -594,6 +601,10 @@ defaults: &defaults
         - '1 vCPUs'
         - '1 GB Memory'
         - '2 GB Disk'
+      supported_platform:
+      - k8s
+      - cf
+      osbapi_compliant_name: 'v10-small'
     - &blueprint_plan5
       id: 'd616b00a-5949-4b1c-bc73-0d3c59f3954a'
       name: 'v1.0-large'
@@ -624,6 +635,10 @@ defaults: &defaults
         - '1 vCPUs'
         - '2 GB Memory'
         - '2 GB Disk'
+      supported_platform:
+      - k8s
+      - cf
+      osbapi_compliant_name: 'v10-large'
     - &blueprint_plan6
       id: 'b91d9512-b5c9-4c4a-922a-fa54ae67d235'
       name: 'v1.0-dedicated-large-deprecated'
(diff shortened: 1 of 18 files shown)


lib/fabrik/Fabrik.js
```diff
@@ -41,11 +42,20 @@ class Fabrik {
...
         return Promise
           .try(() => context ? context : instance.platformContext)
-          .then(context => instance.assignPlatformManager(BasePlatformManager.getInstance(instance_id, context)))
+          .then(context => instance.assignPlatformManager(Fabrik.getPlatformManager(context.platform)))
           .return(instance);
       });
   }
 
+  static getPlatformManager(platform) {
+    const PlatformManager = (platform && CONST.PLATFORM_MANAGER[platform]) ? require(`./${CONST.PLATFORM_MANAGER[platform]}`) : undefined;
```
kamath-prasad: attribute name should start with small case here - `platformManager`

config/settings.yml
```diff
@@ -432,6 +432,9 @@ defaults: &defaults
       secret: 'service-fabrik-blueprint-secret'
```
kamath-prasad: maybe its better to have a new plan to have support for k8s and leave the current one's for cf as the plan name in k8s has to be in different format ?!

subhankarc: There will not be any new plan. The same existing BOSH plans will be shown for K8S as well.

lib/fabrik/Fabrik.js
subhankarc: As it is a class reference, kept it as `PlatformManager`

lib/controllers/ServiceBrokerApiController.js
```diff
@@ -34,20 +33,13 @@ class ServiceBrokerApiController extends BaseController {
         } else {
           throw new PreconditionFailed(`At least Broker API version ${minVersion} is required.`);
         }
```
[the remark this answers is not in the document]
jagadish-kb: Post discussion, this is not needed. Keeping this comment for history.

lib/fabrik/BasePlatformManager.js
```diff
@@ -1,15 +1,12 @@
...
-  constructor(guid, context) {
-    this.guid = guid;
-    this.context = context;
+  constructor(platform) {
+    this.platform = platform;
   }
 
-  get platform() {
-    return this.context.platform;
+  getPlatformSpecificCatalog(catalog) {
```
jagadish-kb: Platformmanager is to return things platformSpecifc :-), so may be can we strip this from the method name and just keep it `getCatalog`. Also can return `undefined` from BasePlatformManager as its implementation from specific manager is absent.

lib/constants.js
```diff
@@ -180,8 +180,16 @@ module.exports = Object.freeze({
...
   },
 
+  PLATFORM_ALIAS: {
+    CF: 'cf',
+    K8S: 'k8s'
+  },
+
   PLATFORM_MANAGER: {
-    'cloudfoundry': 'CfPlatformManager'
+    'cloudfoundry': 'CfPlatformManager',
```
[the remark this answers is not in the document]
subhankarc: I have change this and removed the duped ones. Have handled that in the `getPlatformManager` code.

lib/controllers/ServiceBrokerApiController.js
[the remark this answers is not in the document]
subhankarc: sure.

lib/fabrik/BasePlatformManager.js
subhankarc: :) changed it.

PR #213, opened 2018-02-26
Fix the parsing of agent properties for OOB based backups

kamath-prasad:
Modified the code to handle the parsing of agent properties in case the properties are wrapped in an external "agent" attribute.


kamath-prasad:
@jagadish-kb : Have made the changes . Didn't do it originally as I thought some UT's would break :)

PR #218, opened 2018-03-06
GCP B&R Support added

### LINKED ISSUE #217: Support Backup and Restore for GCP
**Story**
As a Service Owner I want to support Backup and Restore functionality for GCP.

**Description**
The Backup functionality introduced for GCP is snapshot based which uses of GCP's incremental snapshots feature. The snapshots taken are encrypted if the underlying disk is encrypted. Restore works with normal disk attach, copy data and disk detach method.
To support this use-case, support for metadata management and delete backup for GCP is added in service-fabrik-broker.

ketakig12:
To add support for GCP B&R via service-fabrik-broker, support for metadata management and delete backup for GCP is added in service-fabrik-broker.
Closes #217

### DIFF
--- a/lib/iaas/BaseCloudClient.js
+++ b/lib/iaas/BaseCloudClient.js
@@ -2,17 +2,18 @@
 
 const _ = require('lodash');
 const errors = require('../errors');
+const CONST = require('../constants');
 const NotImplementedBySubclass = errors.NotImplementedBySubclass;
 
 const CloudProviderError = {
   NotFound: err => {
-    return err.statusCode === 404 || err.failCode === 'Item not found' || err.code === 404 || err.code === 'NotFound';
+    return err.statusCode === CONST.HTTP_STATUS_CODE.NOT_FOUND || err.failCode === 'Item not found' || err.code === CONST.HTTP_STATUS_CODE.NOT_FOUND || err.code === 'NotFound';
   },
   Unauthorized: err => {
-    return err.statusCode === 401 || err.failCode === 'Unauthorized';
+    return err.statusCode === CONST.HTTP_STATUS_CODE.UNAUTHORIZED || err.failCode === 'Unauthorized';
   },
   Forbidden: err => {
-    return err.statusCode === 403 || err.code === 403;
+    return err.statusCode === CONST.HTTP_STATUS_CODE.FORBIDDEN || err.code === CONST.HTTP_STATUS_CODE.FORBIDDEN;
   }
 };
 
--- a/lib/iaas/BaseCloudClient.js
+++ b/lib/iaas/BaseCloudClient.js
@@ -12,7 +12,7 @@ const CloudProviderError = {
     return err.statusCode === 401 || err.failCode === 'Unauthorized';
   },
   Forbidden: err => {
-    return err.statusCode === 403 || err.code === 404;
+    return err.statusCode === 403 || err.code === 403;
   }
 };
 
--- a/lib/iaas/GcpClient.js
+++ b/lib/iaas/GcpClient.js
@@ -11,6 +11,7 @@ const BaseCloudClient = require('./BaseCloudClient');
 const NotFound = errors.NotFound;
 const Unauthorized = errors.Unauthorized;
 const Forbidden = errors.Forbidden;
+const UnprocessableEntity = errors.UnprocessableEntity;
 
 class GcpClient extends BaseCloudClient {
   constructor(settings) {
@@ -59,14 +60,11 @@ class GcpClient extends BaseCloudClient {
       .bucket(container)
       .getFiles(queryOptions)
       .then(results => {
-        const resultFiles = results[0];
-        const files = [];
-        _.each(resultFiles, file => files.push(_
-          .chain(file)
-          .pick('name')
-          .value()
-        ));
-        return files;
+        return _.get(results, 0, []).map(file => {
+          return {
+            name: file.name
+          };
+        });
       })
     );
   }
@@ -150,7 +148,7 @@ class GcpClient extends BaseCloudClient {
         remote: file
       })
       .then((data) => JSON.parse(data))
-      .catchThrow(SyntaxError, new NotFound(`Object '${file}' not found`));
+      .catchThrow(SyntaxError, new UnprocessableEntity(`Object '${file}' data unprocessable`));
   }
 
   deleteSnapshot(snapshotName) {
--- a/lib/iaas/GcpClient.js
+++ b/lib/iaas/GcpClient.js
@@ -60,11 +60,9 @@ class GcpClient extends BaseCloudClient {
       .bucket(container)
       .getFiles(queryOptions)
       .then(results => {
-        return _.get(results, 0, []).map(file => {
-          return {
-            name: file.name
-          };
-        });
+        return _.get(results, 0, []).map(file => ({
+          name: file.name
+        }));
       })
     );
   }
--- a/lib/iaas/GcpClient.js
+++ b/lib/iaas/GcpClient.js
@@ -147,19 +147,40 @@ class GcpClient extends BaseCloudClient {
   }
 
   deleteSnapshot(snapshotName) {
-    return Promise
-      .try(() => {
-        return this.computeClient
-          .snapshot(snapshotName)
-          .delete()
-          .then(retval => logger.info(`Deleted snapshot ${snapshotName}`, retval));
-      })
-      .catchThrow(BaseCloudClient.providerErrorTypes.Unauthorized,
-        new Unauthorized(`Authorization at google compute failed while deleting snapshot ${snapshotName}`))
-      .catchThrow(BaseCloudClient.providerErrorTypes.Forbidden,
-        new Forbidden(`Authentication at google compute failed while deleting snapshot ${snapshotName}`))
-      .catchThrow(BaseCloudClient.providerErrorTypes.NotFound,
-        new NotFound(`Snapshot ${snapshotName} not found while deleting`));
+    return new Promise((resolve, reject) => {
+      function cleanup() {
+        deleteOperation.removeAllListeners();
+      }
+
+      function onerror(err) {
+        cleanup();
+        reject(err);
+      }
+
+      function oncomplete() {
+        cleanup();
+        resolve(logger.info(`Deleted snapshot ${snapshotName}`));
+      }
+
+      function onrunning() {
+        resolve(logger.info(`Deleting snapshot ${snapshotName}`));
+      }
+      var deleteOperation;
+      this.computeClient
+        .snapshot(snapshotName)
+        .delete()
+        .then(data => {
+          // Returns an Operation object that can be used to check the status of the request.
+          // All operations are event emitters. The status of each operation is polled
+          // continuously, starting only after you register a "complete" listener.
+          // https://cloud.google.com/nodejs/docs/reference/compute/0.9.x/Operation#get 
+          deleteOperation = data[0];
+          deleteOperation.on('complete', oncomplete);
+          deleteOperation.on('error', onerror);
+          deleteOperation.on('running', onrunning);
+        })
+        .catch(err => reject(err));
+    });
   }
 
   static validateParams(options) {
--- a/lib/iaas/GcpClient.js
+++ b/lib/iaas/GcpClient.js
@@ -1,11 +1,12 @@
 'use strict';
+const assert = require('assert');
 const _ = require('lodash');
 const Promise = require('bluebird');
 const GcpStorage = require('@google-cloud/storage');
+const GcpCompute = require('@google-cloud/compute');
 const logger = require('../logger');
 const errors = require('../errors');
 const utils = require('../utils');
-const ComputeClient = require('./ComputeClient');
 const BaseCloudClient = require('./BaseCloudClient');
 const NotFound = errors.NotFound;
 const Unauthorized = errors.Unauthorized;
@@ -14,7 +15,7 @@ const Forbidden = errors.Forbidden;
 class GcpClient extends BaseCloudClient {
   constructor(settings) {
     super(settings);
-    this.constructor.validateParams(_.chain(this.settings).value());
+    this.constructor.validateParams(this.settings);
     this.storageClient = this.constructor.createStorageClient(_
       .chain(this.settings)
       .omit('name')
@@ -32,7 +33,6 @@ class GcpClient extends BaseCloudClient {
     if (arguments.length < 1) {
       container = this.containerName;
     }
-
     return this.storageClient
       .bucket(container)
       .get();
@@ -43,28 +43,26 @@ class GcpClient extends BaseCloudClient {
       options = container;
       container = this.containerName;
     }
-    const prefix = options ? options.prefix : null;
     const queryOptions = {};
-    if (prefix) {
-      queryOptions.prefix = prefix;
+    if (options && options.prefix) {
+      queryOptions.prefix = options.prefix;
     }
-    queryOptions.autoPaginate = (options && options.autoPaginate) ? options.autoPaginate : true;
-
-    return Promise.try(() => {
-      return this.storageClient
-        .bucket(container)
-        .getFiles(queryOptions)
-        .then(results => {
-          const resultFiles = results[0];
-          const files = []; 
-          _.each(resultFiles, file => files.push(_       
-            .chain(file)       
-            .pick('name')       
-            .value()      
-          ));   
-          return files;
-        });
-    });
+    queryOptions.autoPaginate = true;
+    return Promise.try(() =>
+      this.storageClient
+      .bucket(container)
+      .getFiles(queryOptions)
+      .then(results => {
+        const resultFiles = results[0];
+        const files = [];
+        _.each(resultFiles, file => files.push(_
+          .chain(file)
+          .pick('name')
+          .value()
+        ));
+        return files;
+      })
+    );
   }
 
   remove(container, file) {
@@ -73,20 +71,16 @@ class GcpClient extends BaseCloudClient {
       container = this.containerName;
     }
     logger.debug(`Deleting file ${file} in container ${container} `);
-    return Promise.try(() => {
-        return this.storageClient
-          .bucket(container)
-          .file(file)
-          .delete();
-      })
-      .catch(BaseCloudClient.providerErrorTypes.Unauthorized, err => {
-        logger.error(err.message);
-        throw new Unauthorized(`Authorization at google cloud storage provider failed while deleting blob ${file} in container ${container}`);
-      })
-      .catch(BaseCloudClient.providerErrorTypes.Forbidden, err => {
-        logger.error(err.message);
-        throw new Forbidden(`Authentication at google cloud storage provider failed while deleting blob ${file} in container ${container}`);
-      })
+    return Promise.try(() =>
+        this.storageClient
+        .bucket(container)
+        .file(file)
+        .delete()
+      )
+      .catchThrow(BaseCloudClient.providerErrorTypes.Unauthorized,
+        new Unauthorized(`Authorization at google cloud storage provider failed while deleting blob ${file} in container ${container}`))
+      .catchThrow(BaseCloudClient.providerErrorTypes.Forbidden,
+        new Forbidden(`Authentication at google cloud storage provider failed while deleting blob ${file} in container ${container}`))
       .catchThrow(BaseCloudClient.providerErrorTypes.NotFound,
         new NotFound(`Object '${file}' not found while deleting in container ${container}`));
   }
@@ -115,12 +109,10 @@ class GcpClient extends BaseCloudClient {
         cleanup();
         resolve(JSON.parse(buffer));
       }
-
-      let uploadedStream = this.storageClient
+      const uploadedStream = this.storageClient
         .bucket(options.container)
         .file(options.remote)
         .createWriteStream();
-
       uploadedStream.on('error', onerror);
       uploadedStream.on('finish', onfinish);
       uploadedStream.end(buffer);
@@ -162,14 +154,10 @@ class GcpClient extends BaseCloudClient {
           .delete()
           .then(retval => logger.info(`Deleted snapshot ${snapshotName}`, retval));
       })
-      .catch(BaseCloudClient.providerErrorTypes.Unauthorized, err => {
-        logger.error(err.message);
-        throw new Unauthorized(`Authorization at google compute failed while deleting snapshot ${snapshotName}`);
-      })
-      .catch(BaseCloudClient.providerErrorTypes.Forbidden, err => {
-        logger.error(err.message);
-        throw new Forbidden(`Authentication at google compute failed while deleting snapshot ${snapshotName}`);
-      })
+      .catchThrow(BaseCloudClient.providerErrorTypes.Unauthorized,
+        new Unauthorized(`Authorization at google compute failed while deleting snapshot ${snapshotName}`))
+      .catchThrow(BaseCloudClient.providerErrorTypes.Forbidden,
+        new Forbidden(`Authentication at google compute failed while deleting snapshot ${snapshotName}`))
       .catchThrow(BaseCloudClient.providerErrorTypes.NotFound,
         new NotFound(`Snapshot ${snapshotName} not found while deleting`));
   }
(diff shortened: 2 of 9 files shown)


lib/iaas/GcpClient.js
```diff
@@ -0,0 +1,200 @@
...
+      .bucket(container)
+      .get();
+  }
+
+  list(container, options) {
+    if (arguments.length < 2) {
+      options = container;
+      container = this.containerName;
+    }
+    const prefix = options ? options.prefix : null;
```
jagadish-kb: if options is defined but prefix is undefined, you would be setting prefix to undefined as against to null as is set when options is undefined. What should be the value of prefix if its missing? is it null or undefined? Fix same default value.

lib/iaas/GcpClient.js
```diff
@@ -0,0 +1,200 @@
...
+    const queryOptions = {};
+    if (prefix) {
+      queryOptions.prefix = prefix;
+    }
+    queryOptions.autoPaginate = (options && options.autoPaginate) ? options.autoPaginate : true;
+
+    return Promise.try(() => {
```
jagadish-kb: if there is a single statement the '{' and 'return' can be avoided. Also have you tried promisifying the functions of the lib ? Then may be you dont have to wrap all methods with a Promise.try().

lib/iaas/GcpClient.js
```diff
@@ -0,0 +1,200 @@
...
+        .then(results => {
+          const resultFiles = results[0];
+          const files = []; 
+          _.each(resultFiles, file => files.push(_       
+            .chain(file)       
+            .pick('name')       
+            .value()      
+          ));   
+          return files;
+        });
```
jagadish-kb: Lines 57 thru 66 can be replaced with this 
```
.then(results => results instanceof Array? results.map((file) => _.get(file, 'name')): []);
```

lib/iaas/GcpClient.js
```diff
@@ -0,0 +1,200 @@
...
+  }
+
+  remove(container, file) {
+    if (arguments.length < 2) {
+      file = container;
+      container = this.containerName;
+    }
+    logger.debug(`Deleting file ${file} in container ${container} `);
+    return Promise.try(() => {
+        return this.storageClient
```
jagadish-kb: same comments as above with regards to explicit return and promisifying lib func.

lib/iaas/GcpClient.js
```diff
@@ -0,0 +1,200 @@
...
+          .bucket(container)
+          .file(file)
+          .delete();
+      })
+      .catch(BaseCloudClient.providerErrorTypes.Unauthorized, err => {
+        logger.error(err.message);
```
jagadish-kb: when using logger.error, always provide the error object so that stack trace gets printed. So you could change the above to 
```
logger.error(`Authorization at google cloud storage provider failed while deleting blob ${file} in container ${container}`, err);
```

lib/iaas/GcpClient.js
```diff
@@ -0,0 +1,200 @@
...
+        throw new Unauthorized(`Authorization at google cloud storage provider failed while deleting blob ${file} in container ${container}`);
+      })
+      .catch(BaseCloudClient.providerErrorTypes.Forbidden, err => {
+        logger.error(err.message);
```
jagadish-kb: same as above

lib/iaas/GcpClient.js
```diff
@@ -0,0 +1,200 @@
...
+        throw new Forbidden(`Authentication at google cloud storage provider failed while deleting blob ${file} in container ${container}`);
+      })
+      .catchThrow(BaseCloudClient.providerErrorTypes.NotFound,
```
jagadish-kb: Any reason why 2 of the 3 in the chain are catch and the other catchThrow? I would keep it consistent. Either ways.

lib/iaas/GcpClient.js
```diff
@@ -0,0 +1,200 @@
...
+      function onerror(err) {
+        cleanup();
+        reject(err);
+      }
+
+      function onfinish() {
+        cleanup();
+        resolve(JSON.parse(buffer));
+      }
+
```
jagadish-kb: unnecessary new lines at 108,113,118 & 123

lib/iaas/GcpClient.js
```diff
@@ -0,0 +1,200 @@
...
+      let uploadedStream = this.storageClient
```
jagadish-kb: use const

lib/iaas/GcpClient.js
```diff
@@ -0,0 +1,200 @@
...
+      })
+      .catch(BaseCloudClient.providerErrorTypes.Unauthorized, err => {
+        logger.error(err.message);
+        throw new Unauthorized(`Authorization at google compute failed while deleting snapshot ${snapshotName}`);
+      })
+      .catch(BaseCloudClient.providerErrorTypes.Forbidden, err => {
+        logger.error(err.message);
+        throw new Forbidden(`Authentication at google compute failed while deleting snapshot ${snapshotName}`);
+      })
+      .catchThrow(BaseCloudClient.providerErrorTypes.NotFound,
```
jagadish-kb: same comments as that provided in `remove` method.

lib/iaas/GcpClient.js
```diff
@@ -0,0 +1,200 @@
...
+        new NotFound(`Snapshot ${snapshotName} not found while deleting`));
+  }
+
+  static validateParams(options) {
+    if (!options) {
+      throw new Error('GcpClient can not be instantiated as backup provider config not found');
+    } else {
+      if (!options.projectId || !options.credentials) {
```
jagadish-kb: Dont you also want to check the validity of `credentials` object?

lib/iaas/GcpClient.js
```diff
@@ -0,0 +1,200 @@
...
+
+  static createStorageClient(options) {
+    return GcpStorage({
+      projectId: options.projectId,
+      credentials: options.credentials
+    });
+  }
+
+  static createComputeClient(settings) {
+    return ComputeClient.createComputeClient(settings).value();
```
jagadish-kb: Somehow this pattern of IAAS specific class taking help of an IAAS agnostic factory doesnt seem right. Why not directly put the GCP Compute creation logic within this function? Factory is used generally when the invoker does not know the flavor. But in this case from a specific flavor we are going to the factory, which seems incorrect. I see that this is following Azure way, but i guess even thats not right.

test/iaas.GcpClient.spec.js
```diff
@@ -0,0 +1,88 @@
+'use strict';
+
+const lib = require('../lib');
+const GcpClient = lib.iaas.GcpClient;
+
+describe('iaas', function () {
+  describe('GcpClient', function () {
+
+    const config = {
```
jagadish-kb: All methods can be UTed via the usage of at-least stubs, if not via nocks (external http)

jagadish-kb:
Also, i see there is security/synk issue? what is it? When i tried to view it, it said i dont have access.

lib/iaas/GcpClient.js
```diff
@@ -0,0 +1,200 @@
...
+const ComputeClient = require('./ComputeClient');
+const BaseCloudClient = require('./BaseCloudClient');
+const NotFound = errors.NotFound;
+const Unauthorized = errors.Unauthorized;
+const Forbidden = errors.Forbidden;
+
+class GcpClient extends BaseCloudClient {
+  constructor(settings) {
+    super(settings);
+    this.constructor.validateParams(_.chain(this.settings).value());
```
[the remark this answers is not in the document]
ketakig12: Yes, should not be required. Will change.

lib/iaas/GcpClient.js
```diff
@@ -0,0 +1,200 @@
...
+    if (arguments.length < 2) {
+      options = container;
+      container = this.containerName;
+    }
+    const prefix = options ? options.prefix : null;
+    const queryOptions = {};
+    if (prefix) {
+      queryOptions.prefix = prefix;
+    }
+    queryOptions.autoPaginate = (options && options.autoPaginate) ? options.autoPaginate : true;
```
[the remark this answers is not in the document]
ketakig12: Yes, that's true. Pagination handling is via nextQuery in GCP as shown at [Bucket#getFiles](https://cloud.google.com/nodejs/docs/reference/storage/1.5.x/Bucket#getFiles). Should I then keep it as default = true ? All the results will be fetched in multiple requests internally by library method.

lib/iaas/GcpClient.js
ketakig12: I do not want to set prefix if null or undefined. Change the query to check both options and options.prefix.

lib/iaas/GcpClient.js
ketakig12: Will remove '{' and return.
The functions are already promisified but it's normal promise and not the bluebird promise. Hence wrapped around try().

lib/iaas/GcpClient.js
ketakig12: - We have to get the filenames from results[0] actually.
- Also, for other IaaS clients, like [AzureClient](https://github.com/cloudfoundry-incubator/service-fabrik-broker/blob/rel-2017.T12b/lib/iaas/AzureClient.js#L61), we also pick lastModified, which I checkd is not getting used. Hence started sending files as an array of objects instead of array of names. I think let's keep this as an array of objects for future needs.

lib/iaas/GcpClient.js
ketakig12: Will remove '{' and return.
The functions are already promisified but it's normal promise and not the bluebird promise, could not use catch with filtered errors and catchthrow, hence wrapped around try().

lib/iaas/GcpClient.js
ketakig12: Since I am throwing error back, will not the stack trace be printed in parent classes? If I also log it here, debugging through logs will be difficult as same errors/stack traces will get printed multiple times at multiple places.

lib/iaas/GcpClient.js
ketakig12: Sure, then I will keep it as catchThrow everywhere in the class.

lib/iaas/GcpClient.js
ketakig12: Replaced with catchThrow().

lib/iaas/GcpClient.js
ketakig12: There are multiple fields in the credentials object like type, client_email, client_id etc. and this is directly imported from a key file generated via GCP console / cli . Hence I do not think there is a need to validate the credentials object further.

lib/iaas/GcpClient.js
ketakig12: Yeah, I have referenced AzureClient for this. We can discuss and I can implement accordingly.

ketakig12:
About the synk issue, I checked with @ashishjain14 , he will get in touch with cf guys and get the vulnerability report.

lib/iaas/GcpClient.js
ketakig12: As discussed, I will use GCP computeclient directly here and not call computeclient abstraction.

lib/iaas/GcpClient.js
ketakig12: As discussed, I will also check for individual parameters in validate.

lib/iaas/GcpClient.js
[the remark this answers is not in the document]
ketakig12: As discussed, keeping it true as a default without providing option.

ketakig12:
@jagadish-kb @kamath-prasad @subhankarc 
Delete snapshot call returns an operation which needs to be polled. Added this polling.

lib/iaas/GcpClient.js
```diff
@@ -0,0 +1,222 @@
...
+const utils = require('../utils');
+const BaseCloudClient = require('./BaseCloudClient');
+const NotFound = errors.NotFound;
+const Unauthorized = errors.Unauthorized;
+const Forbidden = errors.Forbidden;
+
+class GcpClient extends BaseCloudClient {
+  constructor(settings) {
+    super(settings);
+    this.constructor.validateParams(this.settings);
```
sauravmndl: should it be this.constructor.validateParams(this.settings) --> this.constructor.validateParams(_.chain(this.settings).value()) ??

ketakig12: I tested this locally, it works with simple settings param. I have also added unit tests for the same.

lib/iaas/BaseCloudClient.js
```diff
@@ -12,7 +12,7 @@ const CloudProviderError = {
     return err.statusCode === 401 || err.failCode === 'Unauthorized';
   },
   Forbidden: err => {
-    return err.statusCode === 403 || err.code === 404;
+    return err.statusCode === 403 || err.code === 403;
```
sauravmndl: use status code value from constant. We are gradually improving on this in our whole code base. If the PR needs to be merged immediately, then not required to fix it now.

lib/iaas/GcpClient.js
```diff
@@ -0,0 +1,222 @@
...
+          // Returns an Operation object that can be used to check the status of the request.
+          // All operations are event emitters. The status of each operation is polled
+          // continuously, starting only after you register a "complete" listener.
+          // https://cloud.google.com/nodejs/docs/reference/compute/0.9.x/Operation#Operation 
+          deleteOperation = data[0];
+          deleteOperation.on('complete', oncomplete);
+          deleteOperation.on('error', onerror);
+          deleteOperation.on('running', onrunning);
+        })
+        .catch(err => reject(err));
```
kamath-prasad: can you log the error while deleting the snapshot ?

lib/iaas/BaseCloudClient.js
ketakig12: Sure, will do.

lib/iaas/GcpClient.js
ketakig12: Sure, added.

lib/iaas/GcpClient.js
```diff
@@ -0,0 +1,225 @@
...
+      file = container;
+      container = this.containerName;
+    }
+    return this
+      .download({
+        container: container,
+        remote: file
+      })
+      .then((data) => JSON.parse(data))
+      .catchThrow(SyntaxError, new NotFound(`Object '${file}' not found`));
```
jagadish-kb: is syntax error to be treated as 404 not found? or an internal Server Error 500?

lib/iaas/GcpClient.js
jagadish-kb: ok

test/iaas.GcpClient.spec.js
```diff
@@ -0,0 +1,538 @@
...
+  delete: () => {
+    return Promise.reject(deleteSnapshotErrMsg);
+  }
+};
+
+describe('iaas', function () {
+  describe('GcpClient', function () {
+    let sandbox, client;
+    before(function () {
+      sandbox = sinon.sandbox.create();
```
jagadish-kb: Better to create sandbox and restore sandbox where its used. Here at first look it seemed like this was not used.

test/iaas.GcpClient.spec.js
```diff
@@ -0,0 +1,538 @@
...
+        sandbox.stub(utils, 'streamToPromise', utilsStreamToPromiseStub);
+      });
+      after(function () {
+        sandbox.restore();
+      });
+
+      it('container properties should be retrived successfully', function () {
+        return client.getContainer()
+          .then(result => {
+            expect(result[1].name).to.equal(settings.container);
```
jagadish-kb: should the getContainer method return an array or a single object? Since you are fetching a particular container wouldnt it be better that the method returns `result[0]` so that upstream always knows its container details? Also check from other IAAS clients if this method returns an object or an array. So keep it in line.

Also you are referring as `result[1]` & `result[0]` is put as empty object. Why is this?

test/iaas.GcpClient.spec.js
```diff
@@ -0,0 +1,538 @@
...
+    name: 'sample-container',
+    timeCreated: '2017-12-24T10:23:50.348Z',
+    updated: '2017-12-24T10:23:50.348Z',
+    metageneration: '1',
+    location: 'EUROPE-WEST1',
+    storageClass: 'REGIONAL',
+    etag: 'aaa'
+  }
+];
+const getFilesResponse = [
```
jagadish-kb: is this supposed to be an array of array?

test/iaas.GcpClient.spec.js
```diff
@@ -0,0 +1,538 @@
...
+          }
+        ];
+        return client.list(options)
+          .then(result => {
+            expect(result).to.be.an('array');
+            expect(result).to.have.lengthOf(2);
+            expect(result).to.be.eql(expectedResponse);
+          });
+      });
+
```
jagadish-kb: unnecessary new line. generally within a `describe` section all test cases (it) are lined up with no spaces in other test cases.

test/iaas.GcpClient.spec.js
```diff
@@ -0,0 +1,538 @@
...
+            name: 'blob1.txt'
+          },
+          {
+            name: 'blob2.txt'
+          }
+        ];
+        return client.list(options)
+          .then(result => {
+            expect(result).to.be.an('array');
+            expect(result).to.have.lengthOf(2);
```
jagadish-kb: shouldnt this be result[0]?

test/iaas.GcpClient.spec.js
```diff
@@ -0,0 +1,538 @@
...
+            expect(result).to.be.eql(expectedResponse);
+          });
+      });
+
+      it('file/blob deletion should be successful', function () {
+        return client.remove(validBlobName)
+          .then(result => expect(result[0]).to.be.undefined);
```
jagadish-kb: same comment as that for getContainer. Shouldnt the method return return[0] so that such things are not done upstream?

test/iaas.GcpClient.spec.js
```diff
@@ -0,0 +1,538 @@
...
+          .then(() => {
+            if (streamEventHandlers.finish && _.isFunction(streamEventHandlers.finish)) {
+              return streamEventHandlers.finish.call(streamEventHandlers.finish);
+            } else {
+              throw new Error('Event Handlers not registered in delete snapshot');
+            }
+          })
+          .then(() => {
+            return uploadJsonPromise
+              .then(response => expect(response).to.be.eql(jsonContent))
```
jagadish-kb: Does upload method return back the data it uploaded again in response? Just checking?

test/iaas.GcpClient.spec.js
```diff
@@ -0,0 +1,538 @@
...
+      });
+
+      it('file/blob deletion should be successful', function () {
+        return client.remove(validBlobName)
+          .then(result => expect(result[0]).to.be.undefined);
+      });
+
+      it('file/blob deletion should be throw NotFound exception', function () {
+        return client.remove(notFoundBlobName)
+          .catch(err => expect(err).to.be.instanceof(NotFound));
```
jagadish-kb: it would be better to add an explicit `then` in the chain which throws an error like done in LN391. So that for whatsoever reason the method does not throw error then test fails. For now, test would pass.

test/iaas.GcpClient.spec.js
```diff
@@ -0,0 +1,538 @@
...
+      });
+
+      it('file/blob deletion should be throw Forbidden exception', function () {
+        return client.remove('blob4.txt')
+          .catch(err => expect(err).to.be.instanceof(Forbidden));
```
jagadish-kb: same as above

test/iaas.GcpClient.spec.js
```diff
@@ -0,0 +1,538 @@
...
+      });
+
+      it('file/blob upload should be successful', function () {
+        var uploadJsonPromise = client.uploadJson(validBlobName, jsonContent);
+        return Promise.delay(CONNECTION_WAIT_SIMULATED_DELAY)
+          .then(() => {
+            if (streamEventHandlers.finish && _.isFunction(streamEventHandlers.finish)) {
```
jagadish-kb: it would be a good idea to reset the object `streamEventHandlers` to empty afterEach test run. Else it will carry refereces from previous test.

test/iaas.GcpClient.spec.js
```diff
@@ -0,0 +1,538 @@
...
+      });
+      after(function () {
+        sandbox.restore();
+      });
+
+      it('snapshot should be deleted successfully', function () {
+        var deleteSnapshotPromise = client.deleteSnapshot('snapshot-1');
+        return Promise.delay(CONNECTION_WAIT_SIMULATED_DELAY)
+          .then(() => {
+            if (deleteSnapshotEventHandlers.running && _.isFunction(deleteSnapshotEventHandlers.running)) {
```
jagadish-kb: This handler might not be required as quoted above.

lib/iaas/GcpClient.js
```diff
@@ -0,0 +1,225 @@
...
+    if (options && options.prefix) {
+      queryOptions.prefix = options.prefix;
+    }
+    queryOptions.autoPaginate = true;
+    return Promise.try(() =>
+      this.storageClient
+      .bucket(container)
+      .getFiles(queryOptions)
+      .then(results => {
+        const resultFiles = results[0];
```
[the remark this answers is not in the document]
ketakig12: For other IaaS clients, like [AzureClient](https://github.com/cloudfoundry-incubator/service-fabrik-broker/blob/rel-2017.T12b/lib/iaas/AzureClient.js#L61), we also pick lastModified, which I checked is not getting used. Hence started sending files as an array of objects instead of array of names. I think let's keep this as an array of objects for future needs.

lib/iaas/GcpClient.js
```diff
@@ -0,0 +1,225 @@
...
+        .delete()
+        .then(data => {
+          // Returns an Operation object that can be used to check the status of the request.
+          // All operations are event emitters. The status of each operation is polled
+          // continuously, starting only after you register a "complete" listener.
+          // https://cloud.google.com/nodejs/docs/reference/compute/0.9.x/Operation#Operation 
+          deleteOperation = data[0];
+          deleteOperation.on('complete', oncomplete);
+          deleteOperation.on('error', onerror);
+          deleteOperation.on('running', onrunning);
```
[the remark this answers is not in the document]
ketakig12: It will print multiple times. Not required, removing it.

test/iaas.GcpClient.spec.js
ketakig12: Done.

test/iaas.GcpClient.spec.js
ketakig12: getContainer call from GCP returns Bucket object in `result[0]` and metadata in `result[1]` as written in [Bucket#get](https://cloud.google.com/nodejs/docs/reference/storage/1.6.x/Bucket#get). So, instead of returning the array of result, we can return the metadata in `result[1]`

test/iaas.GcpClient.spec.js
ketakig12: Yes, getFiles call returns array in which 0th element is again array of file. Please see [Bucket#getFiles](https://cloud.google.com/nodejs/docs/reference/storage/1.6.x/Bucket#getFiles) and [GetFilesResponse](https://cloud.google.com/nodejs/docs/reference/storage/1.6.x/global#GetFilesResponse)

test/iaas.GcpClient.spec.js
ketakig12: Okay, will remove these extra lines everywhere.

test/iaas.GcpClient.spec.js
ketakig12: The expected response is array of objects, like below, hence not required.
```
expectedResponse = [{
            name: 'blob1.txt'
          },
          {
            name: 'blob2.txt'
          }
        ];
```

test/iaas.GcpClient.spec.js
ketakig12: True, will change.

test/iaas.GcpClient.spec.js
ketakig12: Yes, it returns like below.
```
resolve(JSON.parse(buffer));
```

test/iaas.GcpClient.spec.js
ketakig12: Yes, missed that at some places. Will explicitly add the same at all places.

test/iaas.GcpClient.spec.js
ketakig12: Removed.

test/iaas.GcpClient.spec.js
ketakig12: Sure, added below: 
```
      afterEach(function () {
        streamEventHandlers = {};
      });
```
Also added the same for `deleteSnapshotEventHandlers` in #ComputeOperations.

lib/iaas/GcpClient.js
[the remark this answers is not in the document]
jagadish-kb: The code suggested above by me will still return an array of file names and it does exactly same what is written below from lines 56-63 :-). You can test once if you would like, its just concise and thats how its kept at other places in code base.

lib/iaas/GcpClient.js
```diff
@@ -0,0 +1,225 @@
...
+      }
+
+      function onerror(err) {
+        cleanup();
+        reject(err);
+      }
+
+      function oncomplete() {
+        cleanup();
+        resolve(logger.info(`Deleted snapshot ${snapshotName}`));
```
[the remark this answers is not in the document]
ketakig12: Okay, will move logger statement out.

test/iaas.GcpClient.spec.js
ketakig12: Even AzureClient returns metadata. So, keeping the same behaviour for GCPClient as well.

lib/iaas/GcpClient.js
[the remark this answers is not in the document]
ketakig12: With little modification it returns as expected array of objects.
```
        return _.get(results, 0, []).map(file => {
          return {name: file.name}
        });
```

ketakig12:
@jagadish-kb I have incorporated review comments. Please check and if everything looks fine, please approve.

lib/iaas/GcpClient.js
jagadish-kb: It can just be like this. No need of returns when there is just one line.
```
.then(results => _.get(results, 0, []).map(file => ({name: file.name}));
```

ketakig12: Done, please check and if everything is looks fine, please approve.

PR #220, opened 2018-03-09
Fix/issue deployment ip cache 200

### LINKED ISSUE #200: Deployment Ips Cache contains invalid ips at times
We have introduced DeploymentIpsCache which caches the ips of the vms of a deployment deployed on bosh. This cache is invalidated only at create/update deployment.

However, this cache sometimes contains invalid ips. One such scenario is when a VM goes down for a deployment and other VMs are healthy. (e.g. Ip failover) If a binding request comes for this deployment, the ips for these VMs are cached. But till the other VM comes up, bosh returns one less ip for the deployment (the ip of the VM which has gone down) and we cache these ips for the deployment. 
Later on, when this VM comes up, we still have a cache of ips which doesn't contain this ip.
Thus, any operation we do with these ips is basically erroneous. (e.g. backup call for these ips will not go to one vm. If that ip is actually the master IP (supporting backup feature), then backup call fails)

The cache invalidation finally occurs, as mentioned earlier, during update call. 

An immediate fix could be to remove the cache to avoid such errors. 
A permanent fix could be to have a proper invalidation logic for the cache. This could be a proper TTL (Time To Leave) for the cache entries.
Another solution could be we store the value of number of VMs for a deployment. If any time, we retrieve the ips of the deployment from bosh and we get back lesser ips, we don't store this value in the cache (invalid cache entry).

jagadish-kb:
Fixes #200

### DIFF
--- a/test/mocks/director.js
+++ b/test/mocks/director.js
@@ -83,6 +83,7 @@ exports.verifyDeploymentLockStatus = verifyDeploymentLockStatus;
 exports.acquireLock = acquireLock;
 exports.releaseLock = releaseLock;
 exports.manifest = manifest;
+exports.getDeploymentInstances = getDeploymentInstances;
 
 function uuidByIndex(index) {
   const buffer = new Buffer(1);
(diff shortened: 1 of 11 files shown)


test/mocks/director.js
```diff
@@ -452,4 +453,27 @@ function getDeploymentVms(deploymentName, times, vms, boshDirectorUrlInput, foun
...
+      .reply(404, {
+        'code': 70000,
+        'description': `'Deployment ${deploymentName} doesn\'t exist'`
+      });
+  } else {
+    return nock(boshDirectorUrl)
+      .get(`/deployments/${deploymentName}/instances`)
+      .times(times || 1)
+      .reply(200, vms || [{
+        cid: '081e3263-e066-4a5a-868f-b420c72a260d',
```
pritishmishra: does getDeploymentInstances() response actually return cid? If not, we can remove it from the mock to avoid confusion.

jagadish-kb: the response does contain cid - https://bosh.io/docs/director-api-v1.html#list-instances. The response of /vms is the same as /instances except that cid & agent id would be null in case the vm is deleted.

pritishmishra: ok.. got it..

PR #224, opened 2018-03-12
Fixes following scheduler issues:

jagadish-kb:
1. SIGABRT signal recieved from worker and shutting down of scheduler.
   RCA : When too many backups are piled up (1000s), then within
  each ms, several deletes are fired. Post these many 1000s of deletes
  worker process is crashing. So to mitigate this, every delete has a
  configurable delete delay added. This should hopefully resolve the
  SIGABRT issue.
2. In Scheduler worker logs, used to see the following error during
   maintenance window- 'RangeError: Maximum call stack size exceeded'
   RCA: We were logging the DB models which are wrapped mongoose
   objects which can have complicated nestings and this led to
   recursive calls in rendering it, which led to this error. Now
   logging specific attribs instead of entire object.
3. For dev scenarios earlier the http header 'x-broker-api-version'
   was not made mandatory. However as part of OSBAPI impl, its
   made mandatory, which is not required perse. So reverting that.


config/settings.yml
```diff
@@ -163,7 +163,7 @@ defaults: &defaults
   ####################
   mongodb:
     provision:
-      plan_id: bc158c9a-7934-401e-94ab-057082a5073f
+      #plan_id: bc158c9a-7934-401e-94ab-057082a5073f
```
subhankarc: Why is this commented?

jagadish-kb: Unit tests check for mongodb not configured scenario as well. So that UT would fail with this being present.

subhankarc: ok.

Issue #212, opened 2018-02-26
Orphaned deployments created even when broker sends valid error


subhankarc:
Broker sends error when accepts_incomplete is not sent or when space_guid or organization_zuid missing. But the execution is not stopped there and it sends request to bosh which creates a bosh deployment.

**Steps to reproduce the issue:**
1. Do not send one of these above mentioned mandatory attributes and make a broker call.

PR #225, opened 2018-03-13
Fix for the issue 212

subhankarc:
Changelist:

1. ServiceBrokerApiController validateRequest method throws error
   and stops the execution instead of only setting the response body.

2. errors.js UnprocessableEntity class constructor is enhanced to take
   'reason' as argument to accomodate AsyncRequired error message to be
    sent if accepts_incomplete is not sent as parameter.
    https://github.com/openservicebrokerapi/servicebroker/blob/v2.13/spec.md#asynchronous-operations

3. Removed message and set description and error in the middleware.js global error handler.
   Broker error response should have a description field.
   https://github.com/openservicebrokerapi/servicebroker/blob/v2.13/spec.md#broker-errors

4. Adjusted the tests accordingly

closes issue: #212


lib/errors.js
```diff
@@ -161,8 +161,8 @@ class PreconditionFailed extends HttpClientError {
 exports.PreconditionFailed = PreconditionFailed;
 
 class UnprocessableEntity extends HttpClientError {
-  constructor(message) {
-    super(422, 'Unprocessable Entity', message);
+  constructor(message, reason) {
+    super(422, reason || 'Unprocessable Entity', message);
```
sauravmndl: 422 --> if you can use it constant, better. Not a must fix!!

lib/middleware.js
```diff
@@ -73,7 +73,7 @@ exports.notFound = function () {
 
 exports.error = function (options) {
   options = options || {};
-  const properties = ['status', 'message'];
+  const properties = ['status'];
```
jagadish-kb: instead of removing message from here and setting explicitly in LN92. Can you rather add both message and reason in to this array and avoid LN 92 & 93?

lib/controllers/ServiceBrokerApiController.js
```diff
@@ -212,19 +213,14 @@ class ServiceBrokerApiController extends BaseController {
...
+      throw new UnprocessableEntity('This request requires client support for asynchronous service operations.', 'AsyncRequired');
     }
     const operationType = _.get(req, 'operation_type');
     if (_.includes([CONST.OPERATION_TYPE.CREATE], operationType) &&
       (!_.get(req.body, 'space_guid') || !_.get(req.body, 'organization_guid'))) {
-      res.status(CONST.HTTP_STATUS_CODE.BAD_REQUEST).send({
-        error: 'BadRequest',
-        description: 'This request is missing mandatory organization guid and/or space guid.'
-      });
+      throw new BadRequest('This request is missing mandatory organization guid and/or space guid.');
```
sauravmndl: do you want to put 'BadRequest' as second parameter just like L218?

lib/middleware.js
subhankarc: According to OSB API spec, the broker error should have description field, not message. https://github.com/openservicebrokerapi/servicebroker/blob/v2.13/spec.md#broker-errors

And for errors like Unprocessable Entity, the attribute is 'error', as defined here https://github.com/openservicebrokerapi/servicebroker/blob/v2.13/spec.md#asynchronous-operations

Hence, setting it explicitly.

jagadish-kb: Ok. Got it. Thanks!

lib/controllers/ServiceBrokerApiController.js
subhankarc: Hi @sauravmndl,

For Bad Request, there is no definite answer to what should be the error, and also OSB checker only checks for the error code 400. Hence, let's keep it as is.

lib/errors.js
subhankarc: Done!

PR #226, opened 2018-03-14
Run js actions in separate process

amitmalav:
Currently js actions in deployment hooks are running in broker's node process. When merged, these changes will make js actions run in separate process.

### DIFF
--- a/lib/fabrik/actions/ActionManager.js
+++ b/lib/fabrik/actions/ActionManager.js
@@ -12,6 +12,7 @@ const ScriptExecutor = require('../../utils/ScriptExecutor');
 class ActionManager {
   static getAction(phase, action) {
     try {
+      const actionScriptAbsPath = path.join(__dirname, 'js', `${action}.js`);
       const actionProcessor = require(`./js/${action}`);
       let actionHandler = actionProcessor[`execute${phase}`];
       if (typeof actionHandler !== 'function') {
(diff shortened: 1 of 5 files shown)


lib/utils/ScriptExecutor.js
```diff
@@ -9,13 +9,14 @@ const logger = require('../logger');
 const errors = require('../errors');
 
 class ScriptExecutor {
-  constructor(scriptAbsPath) {
+  constructor(scriptAbsPath, command) {
```
jagadish-kb: if scriptAbsPath is not used, does it make sense to even have it?

lib/fabrik/actions/JSExecuter.js
```diff
@@ -0,0 +1,19 @@
+'use strict';
+
+const Promise = require('bluebird');
+const action = process.argv[2];
+const phase = process.argv[3];
+let context = process.argv[4];
+
```
jagadish-kb: Do we want to add some validation here with regards to required args (atleast action/phase) and throw a meaningful error rather than dumping some JS trace which would say not defined etc..?

lib/fabrik/actions/ActionManager.js
```diff
@@ -12,14 +12,18 @@ const ScriptExecutor = require('../../utils/ScriptExecutor');
...
   static getAction(phase, action) {
     try {
+      const actionScriptAbsPath = path.join(__dirname, 'js', `${action}.js`);
       const actionProcessor = require(`./js/${action}`);
       let actionHandler = actionProcessor[`execute${phase}`];
       if (typeof actionHandler !== 'function') {
         return this.voidImplementation(action, phase, 'JS');
       }
       const execute = function () {
         logger.info('Invoking handler with args..', arguments);
```
jagadish-kb: can change this log method to `Invoking handler ${action} ${phase} with args..`

lib/fabrik/actions/JSExecuter.js
```diff
@@ -0,0 +1,19 @@
...
+const actionProcessor = require(`./js/${action}`);
+try {
+  context = JSON.parse(context);
+} catch (err) {
+  console.err('Error in parsing context ', context);
+}
+Promise.try(() => actionProcessor[`execute${phase}`](context))
```
jagadish-kb: can add a return here - LN14

Issue #228, opened 2018-03-15
IaaS errors are not bubbled up


muralisc:
**Description**

The error thrown by IaaS while deleting snapshot is not bubbled up.
https://github.com/cloudfoundry-incubator/service-fabrik-broker/blob/master/lib/iaas/CloudProviderClient.js#L179

**Steps to reproduce the issue:**
1. Create a backup
2. Delete the snapshot directly from the IaaS
3. Try deleting the backup

**Describe the results you received:**
The backup gets deleted.

**Describe the results you expected:**
The backup should not be deleted and a valid error should be thrown

subhankarc:
@muralisc closing it as PR is merged.

PR #229, opened 2018-03-15
Bubble up iaas error

muralisc:
During delete of a backup metadata, an attempt to delete the snapshot whose id 
is stored in the metadata made.
In the event of failure of the delete operation the error thrown will be bubbled up.
Closes Issue #228


lib/iaas/AzureClient.js
```diff
@@ -154,7 +154,10 @@ class AzureClient extends BaseCloudClient {
           .snapshots
           .deleteMethodAsync(this.settings.resource_group, snapshotName)
           .then(retval => logger.info(`Deleted snapshot ${snapshotName}`, retval))
-          .catch(err => logger.error(`Error occured while deleting snapshot ${snapshotName}`, err));
+          .catch(err => {
```
kamath-prasad: Just for my knowledge , wouldnt it be better if this catch be for the outer [try](https://github.com/muralisc/service-fabrik-broker/blob/5b07c0bd66fbf122dfc36f37cb5578f7b34f1f7b/lib/iaas/AzureClient.js#L152) ?

lib/iaas/AzureClient.js
```diff
@@ -154,7 +154,10 @@ class AzureClient extends BaseCloudClient {
...
+            logger.error('Error occured while deleting snapshot', err);
+            throw err;
```
kamath-prasad: here it would be better if we re-throw the error only if the err is not `Snapshot not found` - Need to check the exact error description when snapshot is not found ? Ideally this would not be the case .

lib/iaas/CloudProviderClient.js
```diff
@@ -176,7 +176,10 @@ class CloudProviderClient extends BaseCloudClient {
           })
           .promise()
           .then(retval => logger.info(`Deleted snapshot ${snapshotId}`, retval))
-          .catch(err => logger.error('Error occured while deleting snapshot', err));
+          .catch(err => {
+            logger.error('Error occured while deleting snapshot', err);
+            throw err;
```
kamath-prasad: same as above..

lib/iaas/AzureClient.js
muralisc: @kamath-prasad  the exact error is InvalidSnapshot.NotFound and is aws-sdk specific error.
we here catch for all errors , errors which we are not still aware of .
@jagadish-kb wanted to log the error. Easier to search the log with known string. Not necessary ?

jagadish-kb: @muralisc  OK

lib/iaas/CloudProviderClient.js
jagadish-kb: @muralisc OK

lib/iaas/AzureClient.js
muralisc: to my understanding in the current code above its the same .
also  we need a catch in the innermost promise chain ?

PR #230, opened 2018-03-15
Improve scheduler maintenance mode

jagadish-kb:
If SF maintenance mode is in failed state, JobScheduler used to
wait till the last maintenance mode turns into successful status.
However, there are scenarios where docker node update failures
results in SF being in failed maintenance status.
To address this, whenever SF maintenance is in any state other than success
or in-progress, SF's DB connection status is checked and if
SF is connected to DB, then JobScheduler is taken out of maintenance
and all Jobs are made to run. 
Via this check following is verified:
1. External APIs are registered successfully via Route Registrar and are up.
2. SF can talk to DB and DB is up. 
So all pre-reqs for JobScheduler to run are good, so can start JobScheduler.


JobScheduler.js
```diff
@@ -147,26 +148,36 @@ class JobScheduler {
...
-                  CONST.OPERATION.ABORTED,
-                  CONST.SYSTEM_USER)
-                .then(success)
-                .catch((err) => {
-                  logger.error('error occurred while updating maintenance info', err);
-                  reject(err);
-                })
-                .finally(() => {
-                  if (this.intervalTimer) {
+            const maintInfoAttrs = ['progress', 'state', 'completedAt', 'reason', 'toVersion', 'fromVersion', 'createdAt', 'updatedAt', 'createdBy', 'updatedBy'];
```
sauravmndl: is it better to put constants in dedicated constants.js file?

JobScheduler.js
```diff
@@ -147,26 +148,36 @@ class JobScheduler {
...
+            if (_.get(maintenanceInfo, 'state', '') !== CONST.OPERATION.IN_PROGRESS) {
+              logger.info(`+-> System is not in maintenance, but its current state is: ${_.get(maintenanceInfo, 'state', '')}, not as expected. `);
```
sauravmndl: Instead of 'system' we can write 'Service Fabrik Broker'. Just a suggestion!

JobScheduler.js
```diff
@@ -147,26 +148,36 @@ class JobScheduler {
...
+              logger.info('checking if service fabrik is up, inspite of unexpected maintenance state - ', _.pick(maintenanceInfo, maintInfoAttrs));
+              this.isServiceFabrikUp()
```
sauravmndl: do you need to 'return' this statement?

JobScheduler.js
```diff
@@ -147,26 +148,36 @@ class JobScheduler {
...
+                .then(status => {
+                  logger.info(`SF Connected to DB :- ${status}`);
+                  if (status === true) {
+                    logger.info('Service fabrik is up, going ahead with starting the scheduler.');
                     clearInterval(this.intervalTimer);
+                    return resolve();
                   }
```
sauravmndl: if status !== true --> what happens?

JobScheduler.js
```diff
@@ -147,26 +148,36 @@ class JobScheduler {
...
                 });
+            } else {
+              logger.info('+-> System is in maintenance :', _.pick(maintenanceInfo, maintInfoAttrs));
+              const currTime = moment();
+              const maintenanceStartTime = _.get(maintenanceInfo, 'createdAt');
+              if (maintenanceStartTime && currTime.diff(maintenanceStartTime) > config.scheduler.maintenance_mode_time_out) {
+                logger.warn(`System in maintenance since ${maintenanceInfo.createdAt}. Exceeds configured maintenance timeout  ${config.scheduler.maintenance_mode_time_out} (ms). Flagging the current maintenance window as aborted.`);
+                maintenanceManager
```
sauravmndl: do you need to 'return' this statement?

JobScheduler.js
```diff
@@ -147,26 +148,36 @@ class JobScheduler {
...
+                  .updateMaintenace(`System in maintenance beyond configured timeout time ${config.scheduler.maintenance_mode_time_out/1000/60} (mins). JobScheduler aborting it.`,
+                    CONST.OPERATION.ABORTED,
+                    CONST.SYSTEM_USER)
+                  .then(() => logger.info('JobScheduler Aborted current maintenance window'))
+                  .catch((err) => {
+                    logger.error('error occurred while aborting maintenance - ', err);
```
sauravmndl: all logger statement initiate with Upper case.

JobScheduler.js
sauravmndl: Also is it better to put catch statement, as  this.isServiceFabrikUp() might throw exception. Host unreachable, may be!!

JobScheduler.js
```diff
@@ -183,6 +194,18 @@ class JobScheduler {
     });
   }
 
+  isServiceFabrikUp() {
+    return serviceFabrikClient.getInfo()
+      .then(info => {
+        logger.info('SF Status :', info);
+        if (info.db_status === CONST.DB.STATE.CONNECTED) {
+          logger.info('SF is connected to DB, can go ahead with starting scheduler.');
```
sauravmndl: This logger statement, you can put in caller. As based on connected status, what to do, let caller decide. Just a thought !

lib/controllers/ServiceFabrikApiController.js
```diff
@@ -168,16 +168,23 @@ class ServiceFabrikApiController extends BaseController {
...
-        .json({
-          name: this.serviceBrokerName,
-          api_version: this.constructor.version,
-          ready: _.isEmpty(missingImages)
-        })
-      );
+      .then(missingImages => dockerMissingImages = missingImages)
+      .catch((err) => {
+        logger.info('error occurred while fetching docker images', err);
+        dockerMissingImages = ['Error occurred while fetching docker images'];
```
sauravmndl: better to leave as empty array or null object itself. As an error statement in array of images doesn't look good. If someone parses it and displays, then gets an image with that statement, which is invalid!

lib/fabrik/DBManager.js
```diff
@@ -14,23 +14,6 @@ const NotFound = errors.NotFound;
...
-  BIND_IN_PROGRESS: 'BIND_IN_PROGRESS',
-  CONNECTING: 'CONNECTING',
-  CONNECTED: 'CONNECTED',
-  DISCONNECTED: 'DISCONNECTED',
-  CONNECTION_FAILED: 'CONNECTION_FAILED',
-  CREATE_UPDATE_FAILED: 'CREATE_UPDATE_FAILED',
-  BIND_FAILED: 'BIND_FAILED',
-  SHUTTING_DOWN: 'SHUTTING_DOWN'
-};
-
```
sauravmndl: Good choice to move this to constants

lib/cf/ServiceFabrikClient.js
```diff
@@ -22,6 +22,22 @@ class ServiceFabrikClient extends HttpClient {
...
+      .getAccessToken()
+      .then(accessToken => this
+        .request({
+          method: 'GET',
+          url: '/api/v1/info',
+          auth: {
+            bearer: accessToken
+          },
+          json: true
+        }, 200)
```
sauravmndl: Better use value 200 from constants

lib/controllers/ServiceFabrikApiController.js
```diff
@@ -168,16 +168,23 @@ class ServiceFabrikApiController extends BaseController {
...
+      })
+      .finally(() => {
+        res.status(200)
```
sauravmndl: use from constants

JobScheduler.js
jagadish-kb: These are not constants, but rather attribs that we want to pick from the model object. Same logger was referring same set of attribs. THe reason to put stuff in constants is also in case they are reusable. Should not bloat constants file with everything in there.

JobScheduler.js
jagadish-kb: Its not just service fabrik broker when in maintenance its the whole system, broker, route registration, mongodb update etc..

JobScheduler.js
jagadish-kb: Not required to return. Because that return goes back to SetTimeout, so there is no effect perse. Also look at the code below if/else block, if ur returning back from there right away timer is never set. So no need to add it.

jagadish-kb: Agree on the catch part. Yes, can lead to unhandled exception error.

JobScheduler.js
jagadish-kb: Polling continues.

JobScheduler.js
jagadish-kb: same as [above](https://github.com/cloudfoundry-incubator/service-fabrik-broker/pull/230#discussion_r174901021)

lib/controllers/ServiceFabrikApiController.js
jagadish-kb: Fixed it in better way.

lib/controllers/ServiceFabrikApiController.js
jagadish-kb: Done

JobScheduler.js
jagadish-kb: This is personal perference and there is no guideline perse as code base has many loggers like the above. Will incorporate the above, if the the same is agreed upon and enforced as a guideline.

JobScheduler.js
jagadish-kb: Unnecessary logger. removing it.

lib/controllers/ServiceFabrikApiController.js
```diff
@@ -168,16 +168,21 @@ class ServiceFabrikApiController extends BaseController {
...
+      .then(missingImages => allDockerImagesRetrieved = _.isEmpty(missingImages))
+      .tap(missingImages => logger.info('missing images..', missingImages))
+      .catch(err => logger.info('error occurred while fetching docker images', err))
+      .finally(() => {
+        res.status(CONST.HTTP_STATUS_CODE.OK)
+          .json({
+            name: this.serviceBrokerName,
+            api_version: this.constructor.version,
+            ready: allDockerImagesRetrieved,
+            db_status: this.fabrik.dbManager.getState().status
```
kamath-prasad: can this break where-ever we do not have service-fabrik-mongodb (like non GA landscapes) , since the info endpoint is called from monitoring stack to check if sf is up or not ?

JobScheduler.js
```diff
@@ -183,6 +195,17 @@ class JobScheduler {
     });
   }
 
+  isServiceFabrikUp() {
+    return serviceFabrikClient.getInfo()
+      .then(info => {
+        logger.info('SF Status :', info);
+        if (info.db_status === CONST.DB.STATE.CONNECTED) {
```
kamath-prasad: can directly `return info.db_status === CONST.DB.STATE.CONNECTED`

JobScheduler.js
```diff
@@ -147,26 +148,37 @@ class JobScheduler {
...
-                .then(success)
-                .catch((err) => {
-                  logger.error('error occurred while updating maintenance info', err);
-                  reject(err);
-                })
-                .finally(() => {
-                  if (this.intervalTimer) {
+            const maintInfoAttrs = ['progress', 'state', 'completedAt', 'reason', 'toVersion', 'fromVersion', 'createdAt', 'updatedAt', 'createdBy', 'updatedBy'];
+            if (_.get(maintenanceInfo, 'state', '') !== CONST.OPERATION.IN_PROGRESS) {
+              logger.info(`+-> System is not in maintenance, but its current state is: ${_.get(maintenanceInfo, 'state', '')}, not as expected. `);
```
kamath-prasad: just for my understanding , at this point we cannot say that this state is not as expected right :) ?

Only if SF is up, we can say that , correct ? - which we check right below.

lib/controllers/ServiceFabrikApiController.js
jagadish-kb: No. It returns it as `"db_status":"NOT_CONFIGURED"`

JobScheduler.js
jagadish-kb: Ok

lib/controllers/ServiceFabrikApiController.js
kamath-prasad: Ok , then its fine !

PR #231, opened 2018-03-16
Fix the code to check quota in update flow only if it is a plan update

kamath-prasad:
Currently the quota check is performed in `cf update` flow even if there is no plan update. 

This approach blocks the `cf update` for users who do not want to perform a plan update and who do not have sufficient quota.

Hence ensuring that the quota check is skipped in the update flow if its not a plan update.

### DIFF
--- a/lib/middleware.js
+++ b/lib/middleware.js
@@ -135,6 +135,9 @@ exports.checkQuota = function () {
     if (utils.isServiceFabrikOperation(req.body)) {
       logger.debug('[Quota]: ServiceFabrikOperation : calling next handler..');
       next();
+    } else if ( CONST.HTTP_METHOD.PATCH === req.method && utils.isNotPlanUpdate(req.body)) {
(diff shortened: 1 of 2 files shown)


lib/middleware.js
```diff
@@ -135,6 +135,9 @@ exports.checkQuota = function () {
     if (utils.isServiceFabrikOperation(req.body)) {
       logger.debug('[Quota]: ServiceFabrikOperation : calling next handler..');
       next();
+    } else if (CONST.HTTP_METHOD.PATCH === req.method && utils.isNotPlanUpdate(req.body)) {
```
pritishmishra: shouldn't this be `req.method === CONST.HTTP_METHOD.PATCH` ?

kamath-prasad: It can be that way as well .

But its a best practice to use constant on the LHS - https://en.wikipedia.org/wiki/Yoda_conditions to avoid programming errors of using assignment operators`.

pritishmishra: ok.. this is fine then.

jagadish-kb: This check could have gone as an OR clause in Line135 as the code fragments of both the blocks are the same?

PR #232, opened 2018-03-16
Randomising schedule interval when input with hours

jagadish-kb:
When human-readable format of hours is specified as interval,
then the schedules are not randomized and all the jobs start
at the same 'n'th hour. So to address this issue, if the
input hour's multiples leads to 24 hrs, then we can effectively
create random cron schedules by randomizing the hrs between
0 to x-1 hour and repeating the hours till 24.
Examples:
'8 hours' : '23 0,8,16 * * *' ,   '10 2,10,18 * * *' 
'4 hours' : '59 1,5,9,13,17,21 * * *', '15 3,7,11,15,19,23 * * *'
'7 hours' : '7 hours' (not possible to randomize an hr whose multiple does not lead to 24)

### DIFF
--- a/lib/jobs/ScheduleManager.js
+++ b/lib/jobs/ScheduleManager.js
@@ -31,6 +31,9 @@ class ScheduleManager {
       if (interval === CONST.SCHEDULE.DAILY) {
         interval = this.getRandomDailySchedule();
         logger.info(`Schedule interval input as 'daily'. So setting following daily random schedule - ${interval}`);
+      } else if (interval.indexOf('hours') !== -1) {
+        interval = this.getRandomHourlySchedule(interval);
+        logger.info(`human hours interval input, so setting following hourly random schedule - ${interval}`);
       } else if (interval === CONST.SCHEDULE.RANDOM) {
         const JobDefinition = JobFabrik.getJob(jobType);
         interval = JobDefinition.getRandomRepeatInterval();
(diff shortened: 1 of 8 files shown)


lib/jobs/ScheduleManager.js
```diff
@@ -53,6 +56,22 @@ class ScheduleManager {
     return `${min} ${hr} * * *`;
   }
 
+  static getRandomHourlySchedule(interval) {
+    try {
+      const everyXhrs = parseInt(/^[0-9]+/.exec(interval)[0]);
+      logger.info(`schedule is to run every ${everyXhrs} hours`);
```
sauravmndl: it is better to put a not 'NaN' check before logging. Otherwise in case of input 'blah hours' log --> 'schedule is to run every blah hours' doesn't look good. if everyHhrs === NaN directly throw exception.

test/utils.spec.js
```diff
@@ -49,5 +49,13 @@ describe('utils', function () {
       expect(utils.getRandomCronForOnceEveryXDays(7)).to.eql('1 1 1,8,15,22 * *');
       expect(utils.getRandomCronForOnceEveryXDays(15)).to.eql('1 1 1,16 * *');
     });
+    it('should return a random cron expression for every x hours for the day', function () {
+      const AssertionError = require('assert').AssertionError;
+      expect(utils.getRandomCronForEveryDayAtXHoursInterval.bind(utils, 29)).to.throw(AssertionError);
+      //bind returns a ref to function which is executed and checked for if it threw exception.
+      expect(utils.getRandomCronForEveryDayAtXHoursInterval(8)).to.eql('1 1,9,17 * * *');
```
kamath-prasad: just for my knowledge , how can we be sure that the getRandomInt function returns these specific values in UT ?

lib/jobs/ScheduleManager.js
jagadish-kb: @sauravmndl the logger will not be printed. It will throw exception at LN61 already as the regex wont be evaulated and it leads to accessing 0th index of an non-existing array. I had put an assert but that seemed unnecessary when testing.

test/utils.spec.js
jagadish-kb: its [mocked](https://github.com/jagadish-kb/service-fabrik-broker/blob/3b318621a0516420760bfc9c815b572ee8335028/test/utils.spec.js#L39) not to return random numbers :-)

lib/jobs/ScheduleManager.js
jagadish-kb: @sauravmndl  - one more thing, dont approve the PRs even if there are minor comments, let them be answered and then you can approve, else you would then leave it upto the committer :-)

test/utils.spec.js
kamath-prasad: Ok , thanks :)

PR #240, opened 2018-03-28
scheduler optimizations

jagadish-kb:
1. During broker update both docker/broker udpates were being
considered as broker update. Now explicitly checking for
drain message from broker to see if broker is undergoing an update
or not. Only if broker is undergoing an update only then is
scheduler turned off and put into maintenance.

2.Currently every time broker or scheduler starts system jobs
are scheduled again and again without first chedking if they
are already scheduled. Fixiing this by first checking if there
is a schedule and if there are any changes in the schedule interval.
If there is a need to reschedule, only then they are rescheduled
else not.


lib/jobs/ScheduleManager.js
```diff
@@ -168,18 +170,32 @@ class ScheduleManager {
...
       return this
-        .schedule(
-          jobDefinition.name,
-          jobDefinition.type,
-          jobDefinition.interval,
-          jobDefinition.job_data,
-          CONST.SYSTEM_USER);
+        .getSchedule(jobDefinition.name, jobDefinition.type)
+        .then(jobDetails => scheduleSystemJob = jobDetails.repeatInterval !== jobDefinition.interval)
+        .catch(errors.NotFound, () => scheduleSystemJob = true)
```
kamath-prasad: just asking , do we want to catch other run time errors in case DB call returns error ?

lib/maintenance/MaintenanceManager.js
```diff
@@ -64,7 +64,13 @@ class MaintenanceManager {
       state: CONST.OPERATION.IN_PROGRESS
     };
     return Repository
-      .findOne(CONST.DB_MODEL.MAINTENANCE_DETAIL, criteria);
+      .findOne(CONST.DB_MODEL.MAINTENANCE_DETAIL, criteria)
+      .then(info => info === null ? null : _.set(_.clone(info), 'broker_update_initiated', this.isBrokerUpdateInitiated(info)));
```
kamath-prasad: can we have this as a constant `broker_update_initiated` instead of duplicating it ?

lib/jobs/ScheduleManager.js
jagadish-kb: We want to catch only known errors other runtime errors cant be handled gracefully and its better it be propagated up the chain and let global handler decide what is to be done (i.e. exit and restart etc..)

lib/maintenance/MaintenanceManager.js
jagadish-kb: This attrib removed completely with newer impl.

lib/jobs/ScheduleManager.js
kamath-prasad: Ok..

kamath-prasad:
@jagadish-kb : can you pls resolve conflicts ?

JobScheduler.js
```diff
@@ -172,9 +174,25 @@ class JobScheduler {
...
             } else {
               logger.info('+-> System is in maintenance :', _.pick(maintenanceInfo, maintInfoAttrs));
+              const downTimePhase = maintenanceManager.getDowntimePhase(maintenanceInfo, config.scheduler.downtime_maintenance_phases);
+              if (downTimePhase === undefined) {
+                logger.info('Current Maintenance phase does not affect scheduler downtime. So going ahead and starting it...');
+                clearInterval(this.intervalTimer);
+                return resolve();
+              }
+              logger.info(`Scheduler in downtime phase: ${downTimePhase}, will wait till maintenance state changes from in-progress to completed state`);
+              const phaseStartTimeInMs = Date.parse(downTimePhase.substring(downTimePhase.lastIndexOf('at') + 2));
```
kamath-prasad: instead of parsing this way , cant we get the start time of this phase depending on the `updatedAt` field of `maintenanceInfo` ?

lib/maintenance/MaintenanceManager.js
```diff
@@ -67,6 +67,17 @@ class MaintenanceManager {
       .findOne(CONST.DB_MODEL.MAINTENANCE_DETAIL, criteria);
   }
 
+  getDowntimePhase(maintenanceInfo, downTimePhases) {
```
kamath-prasad: if I understand correctly , `maintenanceInfo.progress` is an array which is filled up the way we have been updating maintenance , correct ?

Ideally we should check only for the last entry of that array and check if matches with any of the the `downTimePhases` right ?

JobScheduler.js
jagadish-kb: Actually thats what i wanted to do earlier, but the phases keep on getting updated. ex., SF Update, Registration update and then mongodb update. Now if i take the updated time into consideration, then it will be time of the last phase rather than the start of downtime phase. But yeah use updated time incase parsing of time fails.

jagadish-kb: Actually made it work with updatedAt field itself :-)

lib/maintenance/MaintenanceManager.js
jagadish-kb: yes you are right in a way, but the issue is its truly not a 'phase' info that we have. From progress info  we are inferring the phases. When updating the phase both start and end of the phases are logged in. So if we just check for the last entry then there could be some (corner) cases where we might be in a stage where in broker update is succeeded and we might just see that as not in maintenance and start the job. So just to avoid this having this. But yeah, if you insist i can just check for last entry or last but 1 entry which should take care even of the boundary case.

JobScheduler.js
kamath-prasad: Ok , fine.

lib/maintenance/MaintenanceManager.js
kamath-prasad: yes ok..

kamath-prasad:
@sauravmndl : can you please review this PR ?

PR #241, opened 2018-03-29
Bluebird Promise referred instead of default promise

jagadish-kb:
Promise.delay is used in scheduled OOB however bluebird promise was not referred. This resulted in runtime error `TypeError: Promise.delay is not a function` as seen in schedule worker logs. This PR should fix it.


kamath-prasad:
@sauravmndl : can you please review/approve this PR as it is needed today for test cycle fixes ?

PR #244, opened 2018-04-04
Fix for AWS RequestLimitExceeded issue

sauravmndl:


lib/iaas/ComputeClient.js
```diff
@@ -12,11 +13,20 @@ class ComputeClient {
...
-            region: settings.region
-          });
+          {
+            let ec2_config = {
+              accessKeyId: settings.keyId,
+              secretAccessKey: settings.key,
+              region: settings.region
+            };
+            if ((typeof settings.max_retries === 'number') &&
+              Number.isInteger(settings.max_retries) && settings.max_retries > 0) {
```
kamath-prasad: dont you want to make it stricter ?

`settings.max_retries > 4` ?

kamath-prasad:
@sauravmndl : Change is also needed in SF bosh-release , correct ?

lib/iaas/ComputeClient.js
```diff
@@ -1,22 +1,34 @@
...
       .try(() => {
         switch (settings.provider) {
         case 'amazon':
-          return new AWS.EC2({
-            accessKeyId: settings.keyId,
-            secretAccessKey: settings.key,
-            region: settings.region
-          });
+          {
+            let ec2_config = {
```
pritishmishra: I think we should use lowerCamelCase for object (variable) names. 
Suggesting as per: https://github.com/felixge/node-style-guide

lib/iaas/ComputeClient.js
```diff
@@ -1,22 +1,34 @@
...
+              accessKeyId: settings.keyId,
+              secretAccessKey: settings.key,
+              region: settings.region,
+              maxRetries: 10
+            };
+            if ((typeof settings.max_retries === 'number') &&
+              Number.isInteger(settings.max_retries) && settings.max_retries > 0) {
```
pritishmishra: We can use only `Number.isInteger(settings.max_retries)` and `typeof settings.max_retries` is not required here.

sauravmndl:
@ashishjain14 : incrementally goes up 2, 4, 8 ,16 like that.

lib/fabrik/DirectorManager.js
```diff
@@ -609,6 +609,9 @@ class DirectorManager extends BaseManager {
         secret: undefined,
         trigger: _.get(args, 'trigger', CONST.BACKUP.TRIGGER.ON_DEMAND)
       })
+      .assign(_.get(config, 'backup.provider.max_retries') ? {
```
jagadish-kb: Does not seem right to be  passing this arg from invoker to the client? Why shouldnt this be encapsulated within the agents configs? Agents configs flowing from broker does not seem right. 
Also have we change the Agents API version if we are modifying the signature? Let us be aware of agent versions ?

lib/fabrik/DirectorManager.js
```diff
@@ -802,13 +805,17 @@ class DirectorManager extends BaseManager {
...
-      snapshotId: backupMetadata.snapshotId
-    };
+    const backup = _.chain({
+        guid: args.backup_guid,
+        timeStamp: args.time_stamp,
+        type: backupMetadata.type,
+        secret: backupMetadata.secret,
+        snapshotId: backupMetadata.snapshotId
+      })
+      .assign(_.get(config, 'backup.provider.max_retries') ? {
```
jagadish-kb: same as above.

lib/iaas/ComputeClient.js
```diff
@@ -1,22 +1,34 @@
...
-            accessKeyId: settings.keyId,
-            secretAccessKey: settings.key,
-            region: settings.region
-          });
+          {
+            let ec2Config = {
+              accessKeyId: settings.keyId,
+              secretAccessKey: settings.key,
+              region: settings.region,
+              maxRetries: 10
```
jagadish-kb: Move hardcoding to constant. ( Isn't this a constant review comment from your side ;-))

lib/iaas/ComputeClient.js
```diff
@@ -1,22 +1,34 @@
...
+            };
+            if (Number.isInteger(settings.max_retries) &&
```
jagadish-kb: unnecessary check for Number. Nowhere in our code do we check for the validity of the data type in settings.yml. Whatever is the value in the settings.yml its assumed its set with right data type (for ex, so many durations in milliseconds). If you start doing such validations only at one place it seems to break the consistency. I would recommend removing this.

 So if you remove this explicit numeric check you can set the initial value directly to ec2Config variable this way:
```
 ...
 maxRetries: settings.max_retries || CONST.AWS_MAX_RETRIES
}
```

lib/fabrik/DirectorManager.js
kamath-prasad: Does not seem right to be passing this arg from invoker to the client? Why shouldnt this be encapsulated within the agents configs? Agents configs flowing from broker does not seem right.

--> @jagadish-kb we had discussed this , currently if we need to increase the number of retries , then we will need to update of deployments which will be a overhead , hence passing this info from broker would be better , right ?

Also this would be probably be needed in the future if we pass the creds from broker instead of storing them in the service vm ?

Agree that we would need to modify the version/signature of agent API.

lib/iaas/ComputeClient.js
sauravmndl: what if someone assigns string value ?

lib/iaas/ComputeClient.js
sauravmndl: done :)

lib/fabrik/DirectorManager.js
jagadish-kb: @kamath-prasad  This has strong code smells for couple of reasons:
1. Max_Retries is specific to AWS but this API is generic and not AWS specific, so sending in AWS specific params from generic API isnt right.
2. IMHO retries should an intrinsic part of Backup and Restore Agent library and it should have this within its scope. If some configs must change for that definitely its release should get updated. Somehow spilling something which should be within the confines of BnR library. Also, Agent implementations are just pass through layers for this param and. Asking them to pass it just for the ease the updating of the config on the fly does not seem right. Also we would not be changing these values so frequently, this will anyways be done post evaluation of runs from one sprint to another and tweak the values, at which time if required we can force update the agents.

sauravmndl: Actually I was thinking of implementing a generic framework of dynamic configuration. Keeping all monolithic configs in local env looks good for native apps. But in cloud there are multiple (actually many) micro-services running with similar configs (value changes, structure/templates same). Updating the many micro-services is time consuming. If there are code changes, yes definitely update is required. But for configs, the cloud is moving towards dynamic config via API calls. Here is an example for app's config changes https://docs.cloudfoundry.org/credhub/. So, I think it is a good move towards sending config on the fly. So to make the APIs generic we can send configs in API following way

............
config: {
  iaas: {
    aws: {
        max_retries: 10
    }
}
}
............

Later on we can extend the above signature to pass other dynamic config paras e.g. credentials. Please let me know if above looks good.

jagadish-kb: @kamath-prasad 
> Not entirely true . Retries will be done from broker code too since even it uses EC2 API's.
Agree but its Bnr Lib is python and Broker code is NodeJs. They are using different clients all together and configure them accordingly per their needs. They have different access keys, so they might have different retries as well. So saying they must also have same retry count is debatable. If managing multiple configs is a problem we always have bosh jobs exporting properties which can be imported. 

>At this point , this holds true . But in the future if we want to pass the credentials or use the retry count for other IAASes ?
IMHO, Credentials are different than max_retries. Credentials are mandatory, whereas retries are optional. At least i am of the opinion that internal implementation details of a module should never be made visible to invokers for ex.. like caching, retries,..

My basic crib is this : Here you are invoking just an agent API and saying take backup why should the broker care how many times the agent retries? 

You are free to agree to disagree &  you are free to override my disagreement :-) and I can approve your PR, but somehow just for the ease of having the configuration changed and propagating the way its done does not seem right to me and i tend to disagree.

kamath-prasad: @jagadish-kb : As discussed , we will go ahead with storing the config in agent and not go via broker passing it as for now this will be deviating from the current architectural pattern we follow for maintaining agent configs and max_retries is no different config from other existing configs .

I will create separate BL item to see how we can have better/faster update of configs in the service deployments since current approach of waiting for the outdated service deployments to be updated might not be scalable in the long run.

PR #246, opened 2018-04-04
Feature/preupdate agent

### LINKED ISSUE #245: Agent API for PreUpdate actions
Provide support for agent api, which can perform preupdate operations

amitmalav:
Closes  #245

### DIFF
--- a/lib/fabrik/DirectorManager.js
+++ b/lib/fabrik/DirectorManager.js
@@ -224,7 +224,7 @@ class DirectorManager extends BaseManager {
               return this
                 .getDeploymentIps(deploymentName)
                 .then(ips => this.agent.preUpdate(ips, updateActionContext))
-                .tap(response => preUpdateResponse = response.actionResponse)
+                .tap(response => preUpdateResponse = response)
                 .catch(FeatureNotSupportedByAnyAgent, ServiceInstanceNotOperational, err => {
                   logger.debug('+-> Catched expected error of feature \'preUpdate\':', err);
                   return;
--- a/lib/fabrik/DirectorManager.js
+++ b/lib/fabrik/DirectorManager.js
@@ -212,9 +212,12 @@ class DirectorManager extends BaseManager {
               }, opts.context);
               return;
             })
-            .then(() => this.executePreUpdateActions(deploymentName, actionContext))
+            .then(() => {
+              let preUpdateContext = _.cloneDeep(actionContext);
+              return this.executePreUpdate(deploymentName, preUpdateContext);
+            })
             .tap(response => {
-              logger.info('PreUpdate action response is ...', response);
+              logger.info(`PreUpdate action response for ${deploymentName} is ...`, response);
               preUpdateAgentResponse = response;
             });
         case CONST.OPERATION_TYPE.CREATE:
@@ -245,7 +248,7 @@ class DirectorManager extends BaseManager {
       });
   }
 
-  executePreUpdateActions(deploymentName, context) {
+  executePreUpdate(deploymentName, context) {
     _.assign(context, {
       'instance_guid': this.getInstanceGuid(context.deployment_name)
     });
--- a/lib/fabrik/DirectorManager.js
+++ b/lib/fabrik/DirectorManager.js
@@ -224,7 +224,7 @@ class DirectorManager extends BaseManager {
               return this
                 .getDeploymentIps(deploymentName)
                 .then(ips => this.agent.preUpdate(ips, updateActionContext))
-                .tap(response => preUpdateResponse = response)
+                .tap(response => preUpdateResponse = response.actionResponse)
                 .catch(FeatureNotSupportedByAnyAgent, ServiceInstanceNotOperational, err => {
                   logger.debug('+-> Catched expected error of feature \'preUpdate\':', err);
                   return;
--- a/lib/fabrik/DirectorManager.js
+++ b/lib/fabrik/DirectorManager.js
@@ -195,55 +195,62 @@ class DirectorManager extends BaseManager {
     return Promise
       .try(() => {
         switch (action) {
-          case CONST.OPERATION_TYPE.UPDATE:
-            serviceLifeCycle = CONST.SERVICE_LIFE_CYCLE.PRE_UPDATE;
-            if (_.get(params, 'parameters.bosh_director_name') ||
-              username || password) {
-              throw new BadRequest(`Update cannot be done on custom BOSH`);
-            }
-            return this
-              .getDeploymentManifest(deploymentName)
-              .then(manifest => {
-                _.assign(actionContext.params, {
-                  'previous_manifest': manifest
-                });
-                _.assign(opts, {
-                  previous_manifest: manifest
-                }, opts.context);
-                return;
-              })
-              .then(() => {
-                let updateActionContext = _.cloneDeep(actionContext);
-                _.chain(updateActionContext)
-                  .set('instance_guid', this.getInstanceGuid(deploymentName))
-                  .value();
-                _.chain(updateActionContext.params)
-                  .set('service_id', this.service.id)
-                  .set('plan_id', this.plan.id)
-                  .value();
-                return this
-                  .getDeploymentIps(deploymentName)
-                  .then(ips => this.agent.preUpdate(ips, updateActionContext))
-                  .tap(response => preUpdateResponse = response)
-                  .catch(NotFound, () => {
-                    logger.info(`preUpdate is not implemented for ${this.service.name}`);
-                    return;
-                  });
+        case CONST.OPERATION_TYPE.UPDATE:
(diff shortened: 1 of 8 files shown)


lib/fabrik/DirectorManager.js
```diff
@@ -211,6 +211,31 @@ class DirectorManager extends BaseManager {
...
               }, opts.context);
               return;
+            })
+            .then(() => {
+              let updateActionContext = _.cloneDeep(actionContext);
+              _.chain(updateActionContext)
+                .set('instance_guid', this.getInstanceGuid(deploymentName))
+                .value();
+              _.chain(updateActionContext.params)
+                .set('service_id', this.service.id)
```
kamath-prasad: can you pass previous plan id as well ? Since the services might want to know if the update is from old plan to new plan ? Or can they check from their existing deployment's manifest ?

lib/fabrik/DirectorManager.js
```diff
@@ -211,6 +211,31 @@ class DirectorManager extends BaseManager {
...
+                .set('plan_id', this.plan.id)
+                .value();
+              return this
+                .getDeploymentIps(deploymentName)
+                .then(ips => this.agent.preUpdate(ips, updateActionContext))
+                .tap(response => preUpdateResponse = response.actionResponse)
+                .catch(FeatureNotSupportedByAnyAgent, ServiceInstanceNotOperational, err => {
+                  logger.debug('+-> Catched expected error of feature \'preUpdate\':', err);
```
kamath-prasad: `Catched` -> `Caught`

lib/fabrik/DirectorManager.js
```diff
@@ -211,6 +211,31 @@ class DirectorManager extends BaseManager {
...
+                  return;
+                })
+                .catch(err => {
+                  if (err.status === CONST.HTTP_STATUS_CODE.NOT_FOUND) {
+                    logger.debug('+-> Catched expected error of feature \'preUpdate\':', err);
```
kamath-prasad: same as above.

lib/fabrik/DirectorManager.js
```diff
@@ -211,6 +211,31 @@ class DirectorManager extends BaseManager {
...
+                    return;
+                  }
+                  throw err;
```
kamath-prasad: can you log the error here before throwing back ? Or will the calling layer log it ?

lib/fabrik/DirectorManager.js
amitmalav: params object, which is part of cf request, has previous_values (service_id, plan_id, org_id, space_id). previous plan can be extracted as `updateActionContext .params.previous_values.plan_id`

lib/fabrik/DirectorManager.js
amitmalav: Will get logged here: https://github.com/cloudfoundry-incubator/service-fabrik-broker/pull/246/files#diff-52f07cb68d46ec974943b67d2c207a47R262

kamath-prasad: ok ..

lib/fabrik/DirectorManager.js
kamath-prasad: ok , thanks for the clarification .

lib/fabrik/DirectorManager.js
```diff
@@ -211,6 +211,31 @@ class DirectorManager extends BaseManager {
...
+              let updateActionContext = _.cloneDeep(actionContext);
+              _.chain(updateActionContext)
+                .set('instance_guid', this.getInstanceGuid(deploymentName))
+                .value();
+              _.chain(updateActionContext.params)
+                .set('service_id', this.service.id)
+                .set('plan_id', this.plan.id)
+                .value();
+              return this
+                .getDeploymentIps(deploymentName)
```
jagadish-kb: add this pre-update as a supported feature. If the agents dont support this feature then the action must not be invoked. It must be implemented inline with deprovision.

lib/fabrik/DirectorManager.js
```diff
@@ -211,6 +211,31 @@ class DirectorManager extends BaseManager {
                 previous_manifest: manifest
               }, opts.context);
               return;
+            })
+            .then(() => {
+              let updateActionContext = _.cloneDeep(actionContext);
+              _.chain(updateActionContext)
```
jagadish-kb: using chain but using a single operation? why not use a single set operation directly?

jagadish-kb: also why are we cloning this?

lib/fabrik/DirectorManager.js
```diff
@@ -211,6 +211,31 @@ class DirectorManager extends BaseManager {
...
+              return this
+                .getDeploymentIps(deploymentName)
+                .then(ips => this.agent.preUpdate(ips, updateActionContext))
+                .tap(response => preUpdateResponse = response)
+                .catch(FeatureNotSupportedByAnyAgent, ServiceInstanceNotOperational, err => {
+                  logger.debug('+-> Caught expected error of feature \'preUpdate\':', err);
+                  return;
+                })
+                .catch(err => {
+                  if (err.status === CONST.HTTP_STATUS_CODE.NOT_FOUND) {
```
jagadish-kb: if feature not supported first check in settings, rather than invoking and letting it error & then suppressing such error.

lib/fabrik/DirectorManager.js
```diff
@@ -211,6 +211,31 @@ class DirectorManager extends BaseManager {
                 previous_manifest: manifest
               }, opts.context);
               return;
+            })
```
jagadish-kb: Refactor lines 200-238 to a new method. The logic in-between these lines breaks the code flows and readability. Methods should generally be small and concise and if there too much specifics going in, then such things must be generally moved into a seperate method of its own.

lib/fabrik/DirectorManager.js
```diff
@@ -423,7 +448,8 @@ class DirectorManager extends BaseManager {
       previous_manifest: opts.previous_manifest,
       multi_az_enabled: config.multi_az_enabled,
       stemcell: this.stemcell,
-      actions: preDeployResponse
+      actions: preDeployResponse,
+      preUpdateActionResponse: preUpdateResponse
```
jagadish-kb: Dont think these params `preDeployResponse` & `preUpdateResponse` are intuitive enough and might affect the code readability. Can we try to think of naming / grouping them in a better way if possible? if we cant then fine :-)

lib/fabrik/DirectorManager.js
amitmalav: We get list of supported features from `/info` end point of agent. Is there any way to get service specific supported features from settings also?

lib/fabrik/DirectorManager.js
amitmalav: Done!

lib/fabrik/DirectorManager.js
amitmalav: Changed `preUpdateResponse` to `preUpdateAgentResponse`

lib/fabrik/DirectorManager.js
```diff
@@ -240,6 +245,30 @@ class DirectorManager extends BaseManager {
       });
   }
 
+  executePreUpdateActions(deploymentName, context) {
```
jagadish-kb: can you rename the function to `executePreUpdate` ? by dropping the actions suffix, want to consciously keep this separate from the deployment hooks a.k.a. actions. Also this should have been just `executePreUpdateAction` in the first place as its just one action :-)

lib/fabrik/DirectorManager.js
```diff
@@ -240,6 +245,30 @@ class DirectorManager extends BaseManager {
...
+    _.assign(context, {
```
jagadish-kb: You are not cloning context. So the same is passed to executeActions which will missout instance id in create flow, is that ok? I would rather not modify state of a pass by reference which is conditional. why not do this setting when actionContext is getting created, that way its consistent and same flows across.

lib/fabrik/DirectorManager.js
```diff
@@ -211,6 +211,11 @@ class DirectorManager extends BaseManager {
                 previous_manifest: manifest
               }, opts.context);
               return;
+            })
+            .then(() => this.executePreUpdateActions(deploymentName, actionContext))
+            .tap(response => {
+              logger.info('PreUpdate action response is ...', response);
```
jagadish-kb: logger will become very difficult to decipher without context info of `deploymentName` in log statement. Please include deployment name in logger. 
```
logger.info(`PreUpdate action response is for ${deploymentName}...`, response);
```

lib/fabrik/DirectorManager.js
amitmalav: The idea behind adding extra setting to actionContext in executeActions was that it avoids duplication of settings in all the lifecycles which might be calling executeActions. But yeah, better to clone context as it is passed by reference (missed this part, thanks)

PR #247, opened 2018-04-05
Add agent preupdate API documentation

amitmalav:
Changes in `src/css/service-fabrik.css` are:
dl dt{float:left;width:115px;clear:left;text-align:right;overflow-wrap:break-word;font-weight:700}


src/content/api/agent/v1.1/index.apib
```diff
@@ -97,6 +97,34 @@ The agent is asked to perform any necessary operations before the broker execute
...
+Whenever someone asks the broker to update a service instance, it commissions an agent to perform any operations required to be executed before the actual update of the instance gets triggered.
+
+## PreUpdate [POST /preupdate]
+Called when ```cf update-service``` is invoked on a BOSH service instance.
+The agent is asked to perform any necessary operations before the broker executes the update of the service instance.
+
++ Parameters
+    + BODY.deployment_name: `` (string) - Deployment Name
+    + BODY.instance_guid: `` (string) - Instance Guid of service instance
+    + BODY.params: `` (object) - params that comes as part of cf update request, includes context, plan_id, service_id, previous_values and previous_manifest
```
jagadish-kb: Need more elaboration here. I would be better if you expand the schema of this params object here like:
```
BODY.params.context - OSBAPI spec context object information
BODY.params.plan_id - 
BODY.params. service_id - 
BODY.params. previous_values - 
BODY.params.previous_manifest - 
Additionally any command line json args as passed in the `cf update-service` operation can be found under BODY.params
```

src/content/api/agent/v1.1/requests/preupdate.apib
```diff
@@ -0,0 +1,93 @@
...
+                        plan_id: '12345678-1234-5678-1234-123456781234',
+                        service_id: '12345678-1234-5678-1234-123456781234',
+                        organization_id: '12345678-1234-5678-1234-123456781234',
+                        space_id: '12345678-1234-5678-1234-1234567812341'
+                    },
+                    "plan_id": "12345678-1234-5678-1234-123456781234",
+                    "service_id": "12345678-1234-5678-1234-123456781234",
+                    "previous_manifest": {
+                        ...
+                    }
```
jagadish-kb: also include some command line args input as part of cf update-service operation.

src/content/api/agent/v1.1/index.apib
```diff
@@ -97,6 +97,34 @@ The agent is asked to perform any necessary operations before the broker execute
 + Response 500
 <!-- include(responses/500.apib) -->
 
+# Group PreUpdate
+Whenever someone asks the broker to update a service instance, it commissions an agent to perform any operations required to be executed before the actual update of the instance gets triggered.
+
+## PreUpdate [POST /preupdate]
```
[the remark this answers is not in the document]
amitmalav: No, if preupdate is kept under `lifecycle` then it becomes mandatory for services to implement `preupdate` api if they are using lifecycle feature (eg. if some service wants deprovision but not preupdate ), hence created a separate feature. Or are you just talking about changing api route?

src/content/api/agent/v1.1/index.apib
amitmalav: Agree, missed that part. I'll add

src/content/api/agent/v1.1/index.apib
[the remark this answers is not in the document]
amitmalav: Changed feature name to `lifecycle.preupdate` and route is `lifecycle/preupdate`

PR #254, opened 2018-04-11
changes to fix nocontainer issue related to docker & SF timeout issues

challa:
Problem Cause
-----------------

During delete of the docker service (i.e docker container) the delete call goes from CF->SF-> Docker Swarm manager->Docker Node.

Due to a problem between SF->Swarm Manager communication we get Container Not Found issue

Normally Docker Node & Swarm takes around 3-6 Seconds to delete the container and it corresponding volume. SF make two calls to swarm manager for this
-one call to delete container
-one call to delete volume

for each call 'Service-fabrik' makes to swarm manager, SF waits for 60 seconds before it terminate/hungs up the call in case swarm manager takes more than 60 seconds to delete  and after finishing the delete call it cannot communicate back to service-fabrik as the call is already terminated.

if either of the delete calls (container/volume) takes more than 60 seconds SF terminates the call and returns error to CF i.e from CF perspective service delete failed and service stays cf but in docker node the container(i.e service) is deleted.

So for the service of CF there is no container as its already deleted and for any new delete call from CF to SF its returns 404 error to CF as container is already deleted.

Fix
---
In-case container is already deleted and CF (via user) make delete call again, SF will return HTTP 410 (Gone) response code so that CF can delete its orphan service instance.

To Reduce the Frequency of this issue happening we increase the SF  to docker HTTP call time-out from 60 seconds to 170 seconds

### DIFF
--- a/lib/docker/DockerClient.js
+++ b/lib/docker/DockerClient.js
@@ -22,7 +22,7 @@ class DockerClient extends Docker {
    */
   constructor(opts) {
     super(_.defaults({}, opts, config.docker, {
-      timeout: 60 * 1000
+      timeout: 170 * 1000
     }));
     this.constructor.monkeyPatchModem(this.modem);
   }
--- a/lib/docker/DockerClient.js
+++ b/lib/docker/DockerClient.js
@@ -4,6 +4,7 @@ const _ = require('lodash');
 const Promise = require('bluebird');
 const config = require('../config');
 const logger = require('../logger');
+
 Promise.promisifyAll([
   require('dockerode/lib/docker'),
   require('dockerode/lib/container'),
@@ -22,7 +23,7 @@ class DockerClient extends Docker {
    */
   constructor(opts) {
     super(_.defaults({}, opts, config.docker, {
-      timeout: 170 * 1000
+      timeout: config.http_timeout
     }));
     this.constructor.monkeyPatchModem(this.modem);
   }
--- a/test/acceptance/service-broker-api.instances.docker.spec.js
+++ b/test/acceptance/service-broker-api.instances.docker.spec.js
@@ -205,6 +205,29 @@ describe('service-broker-api', function () {
             });
         });
 
+        it('returns 410 Gone', function () {
+          mocks.docker.inspectContainer(instance_id, {
+            Config: {
+              Env: ['context={"platform":"cloudfoundry"}']
+            }
+          },404);
+          return chai.request(app)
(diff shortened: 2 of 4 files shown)


test/acceptance/service-broker-api.instances.docker.spec.js
```diff
@@ -205,6 +205,29 @@ describe('service-broker-api', function () {
             });
         });
 
+        it('returns 410 Gone', function () {
+          mocks.docker.inspectContainer(instance_id, {
```
jagadish-kb: not that it matters, but shouldnt the second param (options) to this method be `{}` or `undefined` ? To help the readability.

test/acceptance/service-broker-api.instances.docker.spec.js
```diff
@@ -205,6 +205,29 @@ describe('service-broker-api', function () {
...
+            .auth(config.username, config.password)
+            .catch(err => err.response)
+            .then(res => {
+              expect(res).to.have.status(410);
+              expect(res.body).to.eql({});
+              mocks.verify();
+            });
+        });
+
+
```
jagadish-kb: unnecessary new line

lib/docker/DockerClient.js
```diff
@@ -22,7 +22,7 @@ class DockerClient extends Docker {
    */
   constructor(opts) {
     super(_.defaults({}, opts, config.docker, {
-      timeout: 60 * 1000
+      timeout: 170 * 1000
```
jagadish-kb: can we move the magic number to a `constants.js`? Also a comment at the place where this is defined as to how this number was arrived at would help in future.

test/mocks/docker.js
```diff
@@ -114,18 +114,18 @@ function startContainer() {
     .reply(204);
 }
 
-function deleteContainer(guid) {
+function deleteContainer(guid, status) {
```
sauravmndl: Remove status as not used.

test/acceptance/service-broker-api.instances.docker.spec.js
```diff
@@ -205,6 +205,29 @@ describe('service-broker-api', function () {
...
+            .set('X-Broker-API-Version', api_version)
+            .auth(config.username, config.password)
+            .catch(err => err.response)
+            .then(res => {
+              expect(res).to.have.status(410);
+              expect(res.body).to.eql({});
+              mocks.verify();
+            });
+        });
+
```
sauravmndl: Remove extra line

PR #262, opened 2018-04-19
Modularize broker code

### LINKED ISSUE #270: Modularize service-fabrik-broker code base
Currently all the processes running (broker, report, scheduler) are using the same codebase even if some of the files are not required by the process.

We can restructure our code in such a way that each process can pick only required files from the service-fabrik-package.

We will be having multiple modules at parent level (ex. broker, common, deployment-hook, backup-manager etc.)

amitmalav:
Fixes #270 
* Made project structure changes
* Added missing UTs

Following structure changes are done as part of this PR.
Broker codebase is refactored into two modules. `common` and `broker`
`common` module can be referenced by `broker` or modules that will be introduced in future like `deployment-hook` module
Directory tree:
.
├── broker
│   ├── bin
│   ├── config
│   ├── lib
│   ├── JobScheduler.js
│   ├── ReportServer.js
│   └── server.js
├── common
│   ├── config
│   ├── errors.js
│   ├── logger.js
│   ├── middleware.js
│   ├── config.js
│   ├── constants.js
│   ├── controllers
│   ├── db
│   ├── EventLogInterceptor.js
│   ├── HttpServer.js
│   ├── app.js
│   ├── public
│   ├── utils
│   └── views
 |── test
├── store
├── ISSUE_TEMPLATE.md
├── LICENSE
├── NOTICE
├── README.md
├── jsdoc.json
├── package-lock.json
├── package.json


broker/lib/routes/admin.js
```diff
@@ -3,76 +3,77 @@
 const express = require('express');
 const config = require('../config');
 const middleware = require('../middleware');
+const common_middleware = require('../../../common/middleware');
```
kamath-prasad: do we follow snake-casing or camel casing ?

ketakig12: Can we not use class inheritance for middleware like Controllers? So that we do not have to require common middleware and modules' own middleware separately and use separate variables.

common/EventLogInterceptor.js
```diff
@@ -14,7 +14,7 @@ class EventLogInterceptor {
   constructor(eventType, appType) {
     const configFileName = `eventlog-config-${appType}.yml`;
     const eventConfAbsPath = process.env.CONF_DIR ? path.join(process.env.CONF_DIR, configFileName) :
-      path.join(__dirname, '..', 'config', configFileName);
+      path.join(__dirname, 'config', configFileName);
```
ketakig12: Aren't files common/config/eventlog-config-*.yml, broker specific? Shouldn't they be in broker/config? In here common/config , maybe we can have some common configs.

amitmalav: `common` module is not aware of the module that is calling eventLogInterceptor.
We can make it mandatory to set CONF_DIR in testing environment but this will be an extra overhead while testing for developers (We will have to set it explicitly before running test similar to SETTINGS_PATH).
Also having CONF_DIR set to specific module configs will force testing to be done at module level. As CONF_DIR will be `broker/lib/config` for broker and `deployment-hooks/lib/config` for deployment hook module. Going forward we can make some changes in our UT framework and incorporate independent testing for each module.

broker/lib/cf/TokenInfo.js
```diff
@@ -1,6 +1,6 @@
 'use strict';
```
jagadish-kb: i feel we could probably move the module `cf` to common. I see scheduler also split from broker down the line and scheduler uses `cf` so it would be better to move this to common.

broker/lib/middleware.js
```diff
@@ -0,0 +1,75 @@
...
+const _ = require('lodash');
+const errors = require('../../common/errors');
+const logger = require('../../common/logger');
+const quota = require('../lib/quota');
+const quotaManager = quota.quotaManager;
+const CONST = require('../../common/constants');
+const Forbidden = errors.Forbidden;
+const BadRequest = errors.BadRequest;
+const utils = require('../../common/utils');
+const catalog = require('./models/catalog');
```
jagadish-kb: should we move `models` also to common? This is something that could get shared with all?

broker/lib/cf/TokenInfo.js
amitmalav: So there is some code that will be used by all the bosh jobs (for ex. config.js), we put that in common, it will have minimal files.
Some code that is gonna be used by only few jobs and not all (Like cf, fabrikStatusPoller), used by only broker and scheduler.
I suggest we can have `cf` also on same level as `common` rather than inside `common` when we modularise `broker/` further.

broker/lib/middleware.js
amitmalav: I feel that `model`, `catalog`, `plan` etc. are very `broker` specific. Just to give an example, service definitions will not be a part of deployment hook job `config`

broker/server.js
```diff
@@ -3,14 +3,14 @@
 console.log('Starting Service Fabrik...');
```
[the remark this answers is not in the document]
amitmalav: @sauravmndl I felt that `server.js` is broker process specific file. And `App.js` is a generic `ExpressApp` that can be used by any process, be it `broker` or `report`  That's why kept under `common`.
Is your concern from node project structure standards perspective?

broker/lib/cf/TokenInfo.js
jagadish-kb: as discussed these common modules can move at root level as they can probably be used by other modules on further refactoring of broker code, for now this is good.

broker/lib/middleware.js
jagadish-kb: same as above.

amitmalav:
There were some high entropy strings present in the code. As these files were touched by this PR, trufflehog is failing.
https://github.com/cloudfoundry-incubator/service-fabrik-broker/blob/master/lib/quota/TokenInfo.js#L3 

To avoid the issue I have changed the base commit of trufflehog to this commit
https://github.com/cloudfoundry-incubator/service-fabrik-broker/pull/262/commits/92b774d3cb30e33df996e6c111251212c5174ac4

`INITIAL_COMMIT="92b774d3cb30e33df996e6c111251212c5174ac4"`
`high_entropy=$(trufflehog --json --entropy true --since_commit "$INITIAL_COMMIT" --regex "$TRAVIS_BUILD_DIR")`
Please do not squash the commits while merging this PR as it will change the commit hash of the initial commit.

PR #264, opened 2018-04-20
Add eventmesh server

muralisc:
* Added settings for etcd
* Added constants for state
* Introduced EventMeshServer interface
* Added Etcd3EventMeshServer , implementation of EventMeshServer to use
  etcd v3
* Added new node package etcd3
* Added unit tests.


lib/eventmesh/Etcd3EventMeshServer.js
```diff
@@ -0,0 +1,132 @@
...
+const etcd = new Etcd3({
+  hosts: config.etcd.url[0]
+});
+
+class Etcd3EventMeshServer extends EventMeshServer {
+
+  registerService(resourceType, serviceId, serviceAttributesValue, servicePlansValue) {
+    const serviceFolderName = `services/${resourceType}/${serviceId}`;
+    const optionKey = `${serviceFolderName}/attributes`;
+    logger.debug('Registering Service', serviceAttributesValue, servicePlansValue);
```
kamath-prasad: can format this log statement better , and also print the resource type and serviceId.

lib/eventmesh/Etcd3EventMeshServer.js
```diff
@@ -0,0 +1,132 @@
...
+} = require('etcd3');
+
+const etcd = new Etcd3({
+  hosts: config.etcd.url[0]
+});
+
+class Etcd3EventMeshServer extends EventMeshServer {
+
+  registerService(resourceType, serviceId, serviceAttributesValue, servicePlansValue) {
+    const serviceFolderName = `services/${resourceType}/${serviceId}`;
```
kamath-prasad: Can refactor to have function which will return serviceFolderName on passing it resoureType and serviceId instead of duplicating the same concatenation in all methods in this class ?

lib/eventmesh/Etcd3EventMeshServer.js
```diff
@@ -0,0 +1,132 @@
...
+    return etcd.get(statusKey).string();
+  }
+
+  getResourceState(resourceType, resourceId) {
+    return this.getResourceKey(resourceType, resourceId, 'state');
+  }
+
+  registerWatcher(key, callback, isRecursive) {
+    if (isRecursive === true) {
+      logger.debug(`Registering reccursive watcher on prefix: ${key}`);
```
kamath-prasad: typo `recursive` .

config/settings.yml
```diff
@@ -192,6 +192,12 @@ defaults: &defaults
       key: <%= certificate('broker.key') %>
       cert: <%= certificate('broker.crt') %>
   ####################
+  # ETCD SETTINGS #
```
subhankarc: We would ideally have different settings for each process as discussed. Most likely etcd would be common for all processes, still would make sense to see that segregation. Not sure if this is in scope of this pr. If not, it’s fine.

lib/constants.js
```diff
@@ -197,6 +197,13 @@ module.exports = Object.freeze({
       DESC: -1
     }
   },
+  RESOURCE_STATE: {
+    IN_QUEUE: 'in queue',
+    IN_PROGRESS: 'in progress',
+    DEPLOYED: 'deployed',
```
subhankarc: Deployed may not be necessary. It was previously added to handle the security group scenario. Now that it is handled by platform managers, it won’t be needed.

lib/eventmesh/Etcd3EventMeshServer.js
```diff
@@ -0,0 +1,132 @@
...
+const config = require('../config');
+const logger = require('../logger');
+const CONST = require('../../lib/constants');
+const EventMeshServer = require('./EventMeshServer');
+const {
+  Etcd3
+} = require('etcd3');
+
+const etcd = new Etcd3({
+  hosts: config.etcd.url[0]
```
subhankarc: Handle multiple hosts here.

lib/eventmesh/EventMeshServer.js
```diff
@@ -0,0 +1,81 @@
...
+
+class EventMeshServer {
+  constructor() {}
+
+  registerService(resourceType, serviceId, serviceAttributesValue, servicePlansValue) {
+    /* jshint unused:false */
+    throw new NotImplementedBySubclass('registerService');
+  }
+
+  getServiceAttributes(resourceType, serviceId) {
```
subhankarc: What is this used for?

lib/eventmesh/Etcd3EventMeshServer.js
```diff
@@ -0,0 +1,132 @@
...
+});
+
+class Etcd3EventMeshServer extends EventMeshServer {
```
subhankarc: Would make sense to outline the properties/attributes of a resource and an annotation as comment. Either here or the super class.

config/settings.yml
muralisc: Will make those changes for the backup manager pr

lib/eventmesh/EventMeshServer.js
muralisc: To get the object of service details common to all plans 
like service name , service guid etc.

lib/eventmesh/Etcd3EventMeshServer.js
muralisc: Done

lib/eventmesh/Etcd3EventMeshServer.js
```diff
@@ -0,0 +1,135 @@
+'use strict';
+
+const config = require('../config');
+const logger = require('../logger');
+const CONST = require('../../lib/constants');
+const EventMeshServer = require('./EventMeshServer');
+
```
ketakig12: Remove extra new line.

lib/eventmesh/Etcd3EventMeshServer.js
```diff
@@ -0,0 +1,135 @@
...
+const {
+  Etcd3
+} = require('etcd3');
+
```
ketakig12: Remove extra new line.

lib/eventmesh/Etcd3EventMeshServer.js
```diff
@@ -0,0 +1,135 @@
...
+const etcd = new Etcd3({
+  hosts: config.etcd.url
+});
+
+class Etcd3EventMeshServer extends EventMeshServer {
+
```
ketakig12: Remove extra new line.

lib/eventmesh/Etcd3EventMeshServer.js
```diff
@@ -0,0 +1,135 @@
...
+    const annotationFolderName = this.getAnnotationFolderName(resourceType, resourceId, annotationName, operationType, opId);
+    logger.debug(`Getting annotation key: ${key} for ${annotationFolderName}`);
+    const statusKey = `${annotationFolderName}/${key}`;
+    return etcd.get(statusKey).string();
+  }
+
+  getAnnotationState(resourceType, resourceId, annotationName, operationType, opId) {
+    return this.getAnnotationKey(resourceType, resourceId, annotationName, operationType, opId, 'state');
+  }
+
```
ketakig12: Remove extra new line.

lib/eventmesh/EventMeshServer.js
```diff
@@ -0,0 +1,104 @@
...
+  getAnnotationKey(resourceType, resourceId, annotationName, operationType, opId, key) {
+    /* jshint unused:false */
+    throw new NotImplementedBySubclass('getAnnotationKey');
+  }
+
+  getAnnotationState(resourceType, resourceId, annotationName, operationType, opId) {
+    /* jshint unused:false */
+    throw new NotImplementedBySubclass('getAnnotationState');
+  }
+
```
ketakig12: Remove extra new line.

test/eventmesh.Etcd3EventMeshServer.spec.js
```diff
@@ -0,0 +1,210 @@
+'use strict';
+
+const Promise = require('bluebird');
+const CONST = require('../lib/constants');
+const eventmesh = require('../lib/eventmesh');
+
```
ketakig12: Remove extra new line.

test/eventmesh.Etcd3EventMeshServer.spec.js
```diff
@@ -0,0 +1,210 @@
...
+      it('should get the annotation state', () => {
+        return eventmesh.server.getAnnotationState('fakeResourceType', 'fakeResourceId', 'fakeAnnotationName', 'fakeOperationType', 'fakeOperationId')
+          .then(() => {
+            expect(getstub.args[0][0]).to.equal('deployments/fakeResourceType/fakeResourceId/fakeAnnotationName/fakeOperationType/fakeOperationId/state');
+          });
+      });
+    });
+
+  });
+});
```
ketakig12: Should we also not check for the response to be in specific format (in every function)? E.g. in 'registerService' function, response should be from second put, but if I add one more 'then' and return etcd.get(), the response will change but unit test will still pass.

muralisc: done

lib/eventmesh/Etcd3EventMeshServer.js
```diff
@@ -0,0 +1,135 @@
...
+    return this.checkValidState(stateValue)
+      .then(() => this.updateResourceKey(resourceType, resourceId, 'state', stateValue));
+  }
+
+  updateResourceKey(resourceType, resourceId, key, value) {
+    const resourceFolderName = this.getResourceFolderName(resourceType, resourceId);
+    const statusKey = `${resourceFolderName}/${key}`;
+    logger.debug(`Updating resource key: ${key} for ${resourceId}`);
+    return etcd.put(statusKey).value(value);
+  }
```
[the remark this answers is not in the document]
muralisc: Abstracted the function so this wont be the case any more

lib/eventmesh/Etcd3EventMeshServer.js
```diff
@@ -0,0 +1,135 @@
...
+  createResource(resourceType, resourceId, val) {
+    logger.debug(`Creating Resource ${resourceType}/${resourceId}`);
+    const resourceFolderName = this.getResourceFolderName(resourceType, resourceId);
+    const optionKey = `${resourceFolderName}/options`;
+    const statusKey = `${resourceFolderName}/state`;
+    const lastOperationKey = `${resourceFolderName}/lastoperation`;
+    return etcd.put(optionKey).value(val)
+      .then(() => etcd.put(statusKey).value(CONST.RESOURCE_STATE.IN_QUEUE))
+      .then(() => etcd.put(lastOperationKey).value(''));
+  }
```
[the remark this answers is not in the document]
muralisc: done

lib/eventmesh/Etcd3EventMeshServer.js
muralisc: Added at [diff-link](https://github.com/cloudfoundry-incubator/service-fabrik-broker/pull/264/files#diff-16bc3b001161ff904934a6b2aa09415aR11)

lib/eventmesh/Etcd3EventMeshServer.js
```diff
@@ -0,0 +1,131 @@
...
+        const statusKey = `${annotationFolderName}/${CONST.ANNOTATION_KEYS.STATE}`;
+        return etcd.put(statusKey).value(CONST.RESOURCE_STATE.IN_QUEUE);
+      });
+  }
+
+  updateAnnotationState(resourceType, resourceId, annotationName, operationType, opId, stateValue) {
+    return this.updateAnnotationKey(resourceType, resourceId, annotationName, operationType, opId, CONST.ANNOTATION_KEYS.STATE, stateValue);
+  }
+
+  updateAnnotationKey(resourceType, resourceId, annotationName, operationType, opId, key, value) {
```
ketakig12: Can this function be directly called from outside of this class? If yes, then we need validation on key, if not, is there a way to make it private?

eventmesh/Etcd3EventMeshServer.js
```diff
@@ -0,0 +1,131 @@
+'use strict';
+
+const config = require('../broker/lib/config');
```
subhankarc: We should have a different config for eventmesh. Let's not use the broker config.

eventmesh/Etcd3EventMeshServer.js
```diff
@@ -0,0 +1,131 @@
...
+const logger = require('../broker/lib/logger');
+const CONST = require('../broker/lib/constants');
```
subhankarc: Import logger and CONST directly from common.

muralisc: done

eventmesh/Etcd3EventMeshServer.js
muralisc: Added new config.js and settings.yml. Updated README

test/eventmesh.Etcd3EventMeshServer.spec.js
```diff
@@ -0,0 +1,351 @@
...
+const CONST = require('../broker/lib/constants');
+const eventmesh = require('../eventmesh');
+const {
+  Etcd3
+} = require('etcd3');
+
+
+describe('eventmesh', () => {
+  describe('Etcd3EventMeshServer', () => {
+    let prefixWatcherStub, keyWatcherStub;
```
subhankarc: There is incorrect restoring of stub which is hindering in other test cases. Can you do the following?

```
    let sandbox, valueStub, stringStub, jsonStub, putstub, getstub, prefixWatcherStub, keyWatcherStub;
    before(() => {
      sandbox = sinon.sandbox.create();
      valueStub = sandbox.stub();
      stringStub = sandbox.stub();
      jsonStub = sandbox.stub();
      putstub = sandbox.stub(Etcd3.prototype, 'put', () => {
        return {
          value: (val) => Promise.resolve(valueStub(val))
        };
      });
      getstub = sandbox.stub(Etcd3.prototype, 'get', () => {
        return {
          json: () => Promise.resolve(jsonStub()),
          string: () => Promise.resolve(stringStub()),
        };
      });

      prefixWatcherStub = sandbox.stub().returns({
        create: () => Promise.resolve({
          on: () => Promise.resolve('prefixWatcherStubResponse')
        }),
      });
      keyWatcherStub = sandbox.stub().returns({
        create: () => Promise.resolve({
          on: () => Promise.resolve('keyWatcherStubResponse')
        }),
      });
      sandbox.stub(Etcd3.prototype, 'watch', () => {
        return {
          prefix: prefixWatcherStub,
          key: keyWatcherStub
        };
      });
    });

    afterEach(function () {
      valueStub.reset();
      prefixWatcherStub.reset();
      keyWatcherStub.reset();
      putstub.reset();
      getstub.reset();
      jsonStub.reset();
      stringStub.reset();
    });

    after(function () {
      sandbox.restore();
    });

```

It works for me. Just replace the code and see if it works.

muralisc: Updated

eventmesh/Etcd3EventMeshServer.js
```diff
@@ -0,0 +1,131 @@
+'use strict';
+
+const config = require('./config');
+const logger = require('../common/logger');
```
amitmalav: `common/logger.js` imports `common/config.js` which loads config using `SETTINGS_PATH` env. Ideally it should use `SF_EVENTMESH_SETTINGS_PATH`

eventmesh/EventMeshServer.js
```diff
@@ -0,0 +1,129 @@
+'use strict';
+
+const _ = require('lodash');
+const CONST = require('../broker/lib/constants');
```
amitmalav: `../common/constants` can be used.

eventmesh/settings.yml
```diff
@@ -0,0 +1,17 @@
+# This file should not be used in deployed environments. Instead, set
+# the SETTINGS_PATH environment variable to point to a configuration
```
amitmalav: SETTINGS_PATH => `SF_EVENTMESH_SETTINGS_PATH`

test/eventmesh.Etcd3EventMeshServer.spec.js
```diff
@@ -0,0 +1,354 @@
+'use strict';
+
+const Promise = require('bluebird');
+const CONST = require('../broker/lib/constants');
```
amitmalav: `../common/constants` can be used.

test/eventmesh.EventMeshServer.spec.js
```diff
@@ -0,0 +1,141 @@
+'use strict';
+const EventMeshServer = require('../eventmesh/EventMeshServer');
+const CONST = require('../broker/lib/constants');
```
amitmalav: `../common/constants` can be used.

eventmesh/Etcd3EventMeshServer.js
amitmalav: As discussed.
We can segregate the tests similar to this: https://github.com/amitmalav/service-fabrik-broker/tree/integrate/d-h

kamath-prasad:
@muralisc / @subhankarc : Can this be merged now ? There is a travis build failing.

eventmesh/Etcd3EventMeshServer.js
```diff
@@ -0,0 +1,131 @@
...
+        const statusKey = `${annotationFolderName}/${CONST.ANNOTATION_KEYS.STATE}`;
+        return etcd.put(statusKey).value(CONST.RESOURCE_STATE.IN_QUEUE);
+      });
+  }
+
+  updateAnnotationState(resourceType, resourceId, annotationName, operationType, opId, stateValue) {
+    return this.updateAnnotationKey(resourceType, resourceId, annotationName, operationType, opId, CONST.ANNOTATION_KEYS.STATE, stateValue);
+  }
+
+  updateAnnotationKey(resourceType, resourceId, annotationName, operationType, opId, key, value) {
```
jagadish-kb: annotation/operation/key -- there are 3 identifiers here. Having some example helps.
Also, its generally not a good practice to have more than 4 or 5 arguments to a function. When the list grows beyond this count, its always preferred to send an Object. Atleast in Java this was followed, can look at in JS.

Also your methods are not validating the values here. Since JS is lenient with args, should there be mandatory check validations done here in these methods and throw exceptions ?

jagadish-kb:
@subhankarc 
I had a basic doubt. If a manager has a watch on path `/deployments/<resourceType>/<resourceId>` will the same watch be called even if there is a change in the annotation i.e. subpath? For ex., will the createManager be invoked even when a backup is triggered via annotation `/deployments/<resourceType>/<resourceId>/<annotationName>/<annotationType>/<annotationId>`? 
Wouldn't this lead to performance issues later when multiple managers will all be getting fired and all managers except one would is supposed to process it?

subhankarc:
@jagadish-kb @muralisc 

Based on feedback from Jagadish, I feel we can restructure our event keys a bit to ensure two managers do not listen to same path and do not get unnecessary events.

For example, 

SF BOSH provision manager listens to `/deployments/bosh` and look for a key `/deployments/bosh/<service-instance-id>`

SF's default Backup manager can look for a key `/backups/default/<service-instance-id>/<backup-id>` and can listen to `/backups/default/`

In a generic way, every Provision Manager watches on /deployments/<provision-type> and every operation manager watches on `/<operation-name>/<operation-type>`

@muralisc @jagadish-kb do you see any issue with this?

jagadish-kb:
@subhankarc @muralisc 

I also had similar thoughts as you have posted in the last comment. No need to ensure every piece of data is under one root. Data can be segregated under different roots and if need be for retrieval / reporting they can be collated. So basically each manager has its own unique path, so that we dont have multiple watchers getting fired and each watcher having to figure out if he needs to do something or not. Rather it should be given that if a watcher is invoked that watcher must do necessary.

Also, another border/boundary condition during restarts. Does etcd ensure that all handlers are fired? for ex., for whatsoever reason the watcher had crashed by the time API server puts in the request and when the watcher job comes back up does etcd identify this and invoke the watcher?

subhankarc:
@jagadish-kb 

So during every restart the watch is registered by the manager. so while the manager is up, it will receive the event. 

If it was down, it would not receive the events during which it was down. But that would not be needed as well, because we would not stagger any request and wait for managers to pick it later. It would reject the request if not immediately picked up by the manager. 

Does it answer your question?

jagadish-kb:
@subhankarc - when you say " It would reject the request if not immediately picked up by the manager." - you mean API server polls for inqueue status (for x duration) and within this xduration if its not picked up by manager it will send back failure response? -- If you meant so its fine.

So while rejecting this request then API server should also ensure that the state value is set to rejected. Just in case to avoid manager to pick it up sometime later and process it which could result in orphaned deployment.

subhankarc:
@jagadish-kb 

That is correct. API server will make the resource status as IN_QUEUE and wait/poll for it to change. Managers pick it up and make it IN Process and only then API server responds. It will be determined by the different status of resource. 

We have so far decided on in_queue, in_process, completed and failed state. We plan to define the possible state values as a spec for any Controller so that writing new controller would be easy.

eventmesh/Etcd3EventMeshServer.js
```diff
@@ -0,0 +1,131 @@
...
+  }
+
+  annotateResource(resourceType, resourceId, annotationName, operationType, opId, val) {
+    const annotationFolderName = this.getAnnotationFolderName(resourceType, resourceId, annotationName, operationType, opId);
+    const optionKey = `${annotationFolderName}/${CONST.ANNOTATION_KEYS.OPTONS}`;
+    logger.debug(`Annotating Resource ${resourceType}/${resourceId} for annotation: ${annotationName}`);
+    return etcd.put(optionKey).value(val)
+      .then(() => {
+        const statusKey = `${annotationFolderName}/${CONST.ANNOTATION_KEYS.STATE}`;
+        return etcd.put(statusKey).value(CONST.RESOURCE_STATE.IN_QUEUE);
```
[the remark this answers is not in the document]
muralisc: Both are a promises
get.string: https://mixer.github.io/etcd3/classes/builder_.singlerangebuilder.html#string
get.json: https://mixer.github.io/etcd3/classes/builder_.singlerangebuilder.html#json
put.value: https://mixer.github.io/etcd3/classes/builder_.putbuilder.html#value

eventmesh/Etcd3EventMeshServer.js
muralisc: Changed the function to accept 1 object instead of multiple parameters.
Added assertions for the expected arguments.

eventmesh/EventMeshServer.js
```diff
@@ -0,0 +1,129 @@
...
+        throw new errors.NotFound(`Could not find state ${state}`);
+      }
+    });
+  }
+
+  getServiceFolderName(resourceType, serviceId) {
+    return `services/${resourceType}/${serviceId}`;
+  }
+
+  getResourceFolderName(resourceType, resourceId) {
```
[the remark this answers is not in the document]
muralisc: Keys do not have subkeys, folders have subkeys.

eventmesh/Etcd3EventMeshServer.js
```diff
@@ -0,0 +1,131 @@
...
+    const resourceKey = `${resourceFolderName}/${key}`;
+    return etcd.get(resourceKey).string();
+  }
+
+  getResourceState(resourceType, resourceId) {
+    return this.getResourceKey(resourceType, resourceId, CONST.RESOURCE_KEYS.STATE);
+  }
+
+  registerWatcher(key, callback, isRecursive) {
+    if (isRecursive === true) {
```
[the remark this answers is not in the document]
muralisc: The callback is the construct of etcd3 package .
The package is already promisified, but they have used a callback mechanism for watcher alone.
https://mixer.github.io/etcd3/classes/watch_.watchbuilder.html

eventmesh/Etcd3EventMeshServer.js
```diff
@@ -0,0 +1,131 @@
...
+    const resourceKey = `${resourceFolderName}/${key}`;
+    logger.debug(`Updating resource key: ${key} for ${resourceId}`);
+    return etcd.put(resourceKey).value(value);
+  }
+
+  getResourceKey(resourceType, resourceId, key) {
+    logger.debug(`Getting resource key: ${key} for ${resourceType}/${resourceId}`);
+    const resourceFolderName = this.getResourceFolderName(resourceType, resourceId);
+    const resourceKey = `${resourceFolderName}/${key}`;
+    return etcd.get(resourceKey).string();
```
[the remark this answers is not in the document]
muralisc: Update the name to `getResourceKeyValue` and `getAnnotationKeyValue`

eventmesh/EventMeshServer.js
```diff
@@ -0,0 +1,129 @@
...
+   *
+   * Resources:
+   *    - /deployments/<resourceType>/<resourceId>
+   * Resource attributes:
+   *    - /deployments/<resourceType>/<resourceId>/options
+   *    - /deployments/<resourceType>/<resourceId>/state
+   *    - /deployments/<resourceType>/<resourceId>/lastoperation
+   *
+   * Annotations:
+   *    - /deployments/<resourceType>/<resourceId>/<annotationName>/<annotationType>/<annotationId>
```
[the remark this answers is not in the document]
muralisc: Added examples

eventmesh/Etcd3EventMeshServer.js
```diff
@@ -0,0 +1,131 @@
...
+        .create()
+        .then(watcher => {
+          return watcher
+            .on('put', callback);
+        });
+    }
+  }
+
+  annotateResource(resourceType, resourceId, annotationName, operationType, opId, val) {
+    const annotationFolderName = this.getAnnotationFolderName(resourceType, resourceId, annotationName, operationType, opId);
```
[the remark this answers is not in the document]
muralisc: Renamed the variables

muralisc:
@subhankarc @jagadish-kb 
As discussed ,
SF's default Backup manager can look for a key `/backups/default/<resource-type>/<service-instance-id>/<backup-id>` and can listen to `/backups/default/`
While provision manager listens to `/deployments/<resource-type>` 
The above change will be made as the part of backupManager changes.

PR #266, opened 2018-04-22
Feature/iptables manager

### LINKED ISSUE #265: Isolate network traffic across sf bosh services subnet
Currently the backing services subnet is in open state (i.e. non standard / ports higher than 1024 are open for other instances in the same subnet) and because of this one service can reach other service instance on these higher ports. This is a security threat where a user with access to some of the tools like netcat, curl, wget etc can try to exploit if there exists a vulnerability. 

To address this issue, the following must be done:
1. Traffic between 2 service instances is fenced
2. No restrictions for Traffic between JB and service instances
3. Service instances are accessible by CF subnet & CF Apps
4. Service instances are reachable by BOSH
5. Traffic across VMs in the cluster is allowed
6. Solution persists across resurrection

jagadish-kb:
Fixes #265

### DIFF
--- a/lib/bosh/manifest/Addons.js
+++ b/lib/bosh/manifest/Addons.js
@@ -37,7 +37,7 @@ class Addons {
       name: CONST.ADD_ON_JOBS.IP_TABLES_MANAGER,
       jobs: [{
         name: CONST.ADD_ON_JOBS.IP_TABLES_MANAGER,
-        release: _.get(config, 'release_name', 'service-fabrik'),
+        release: _.get(config, 'release_name', CONST.SERVICE_FABRIK_PREFIX),
         properties: {
           allow_ips_list: allowIpList.join(','),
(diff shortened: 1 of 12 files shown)


lib/bosh/manifest/Addons.js
```diff
@@ -0,0 +1,50 @@
...
+
+class Addons {
+  constructor(context) {
+    this.networks = Array.isArray(context.networks) ? context.networks : [context.networks];
+  }
+
+  getAll() {
+    const addOnList = _.get(config, 'service_addon_jobs', []);
+    const addOns = [];
+    _.each(addOnList, (addOn) => addOns.push(this.getAddOn(addOn)));
```
muralisc: can use _.map and avoid the push ?

jagadish-kb: done!

PR #268, opened 2018-04-24
Avoid quota check if in cf update-service old and new plans belong to…

ketakig12:
… the same sku

### DIFF
--- a/lib/middleware.js
+++ b/lib/middleware.js
@@ -132,8 +132,8 @@ exports.isFeatureEnabled = function (featureName) {
 
 exports.checkQuota = function () {
   return function (req, res, next) {
-    if (utils.isServiceFabrikOperation(req.body) || (CONST.HTTP_METHOD.PATCH === req.method && utils.isNotPlanUpdate(req.body))) {
-      logger.debug('[Quota]: Check skipped as it is ServiceFabrikOperation or a normal instance update ...: calling next handler..');
+    if (utils.isServiceFabrikOperation(req.body) || (CONST.HTTP_METHOD.PATCH === req.method && (utils.isNotPlanUpdate(req.body) || utils.isSameSkuUpdate(req.body, catalog)))) {
+      logger.debug('[Quota]: Check skipped as it is ServiceFabrikOperation or a normal instance update or plan update with same sku...: calling next handler..');
       next();
     } else {
       const platform = _.get(req, 'body.context.platform');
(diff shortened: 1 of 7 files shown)


test/utils.spec.js
```diff
@@ -58,4 +59,151 @@ describe('utils', function () {
...
+          plan_id: plan_id,
+          service_id: service_id
+        }
+      };
+      expect(utils.isSameSkuUpdate(queryParams, catalog)).to.be.false;
+    });
+  });
+
+  describe('#getSkuNameForPlan', function () {
+    it('should return sku names for plan', function () {
```
kamath-prasad: is `-` part of sku ? ideally sku name will only be xsmall , small etc.. right ?

ketakig12: We do sku regex matching. We added '-' in sku so that the sku for plans like v1.0-small and v1.0-xsmall (since small is substring of xsmall) do not match.

kamath-prasad: ok , got it. thanks !

kamath-prasad:
@jagadish-kb : Can you pls review this PR ? This is a must go for the current sprint.

lib/middleware.js
```diff
@@ -132,33 +132,34 @@ exports.isFeatureEnabled = function (featureName) {
...
+    if (utils.isServiceFabrikOperation(req.body) || (CONST.HTTP_METHOD.PATCH === req.method && (utils.isNotPlanUpdate(req.body) || utils.isSameSkuUpdate(req.body, catalog)))) {
+      logger.debug('[Quota]: Check skipped as it is ServiceFabrikOperation or a normal instance update or plan update with same sku...: calling next handler..');
       next();
     } else {
       const platform = _.get(req, 'body.context.platform');
       if (platform === CONST.PLATFORM.CF) {
         const orgId = req.body.organization_guid || req.body.context.organization_guid || req.body.previous_values.organization_id;
         if (orgId === undefined) {    
           next(new BadRequest(`organization_id is undefined`));   
+        } else {
```
jagadish-kb: I see that QuotaManager also has the logic to skip quota based on whitelisted org or not. So shouldn't we move the above checks that are in `if` clause apart from serviceFabrikOperation operation into the `quotaManager` so that we avoid all these util functions. To me util functions are something reusable but in this case i am seeing what should ideally be part of quota manager logic is getting spilled into utils. would recommend keeping everything within QuotaManager. (Maybe SF operation check is an exception and can be kept outside as is but others i think must be encapsulated within QM class) 

Also figuring out skuname is part of quota manager. So now thats duped & also spilled onto utils.

ketakig12:
@jagadish-kb Please check now.

PR #275, opened 2018-05-21
Feature/bnr status poller 

sauravmndl:
1. Now FabrikStatus poller doesn't depend on state
2. When broker gets restarted, it goes through all the deployments and checks if any there is any ongoing backup for a deployment.
3. Introduced state for BackupStatus poller. Saved as a job in mongodb and runs in every 2 or 3 minutes.
4. So, if broker restarts, then no issues of fetching all deployments.


broker/lib/jobs/BnRStatusPollerJob.js
```diff
@@ -0,0 +1,170 @@
...
+                    .then(() => {
+                      operationStatusResponse.state = CONST.OPERATION.ABORTING;
+                      return operationStatusResponse;
+                    });
+                } else {
+                  // Operation aborted
+                  const currentTime = new Date();
+                  const abortDuration = (currentTime - new Date(instanceInfo.abortStartTime));
+                  if (abortDuration < config.backup.abort_time_out) {
+                    logger.info(`backup abort is still in progress on : ${deployment} for guid : ${backup_guid}`);
```
kamath-prasad: can you use instance id itself in the logs here for consistency.

broker/lib/jobs/BnRStatusPollerJob.js
```diff
@@ -0,0 +1,170 @@
...
+                    operationStatusResponse.state = CONST.OPERATION.ABORTING;
+                  } else {
+                    operationStatusResponse.state = CONST.OPERATION.ABORTED;
+                    logger.info(`Abort Backup timed out on : ${deployment} for guid : ${backup_guid}. Flagging backup operation as complete`);
```
kamath-prasad: same here.

broker/lib/jobs/BnRStatusPollerJob.js
```diff
@@ -0,0 +1,170 @@
+'use strict';
```
kamath-prasad: Just a question , why is this job called `BnRStatusPollerJob` ? Shouldnt it be just `BackupStatusPollerJob` ? For restore we depend upon CF , correct ?

broker/lib/jobs/BnRStatusPollerJob.js
sauravmndl: Changed to deployment instead of instance id. Also job name saved as <deployment>_backup_<backup_guid>. Done keeping cross-consumption scenario in mind. As instance_guid might not be unique across platforms.

broker/lib/jobs/BnRStatusPollerJob.js
sauravmndl: Yes, today we do. But as in SF 2.0 async call dependency on platform is removed, and if the same status poller is enhanced for restore, that's why kept the name like that.

broker/lib/controllers/ServiceFabrikApiController.js
```diff
@@ -10,7 +10,8 @@ const backupStore = require('../iaas').backupStore;
 const filename = backupStore.filename;
 const errors = require('../errors');
 const FabrikBaseController = require('./FabrikBaseController');
-const FabrikStatusPoller = require('../fabrik/FabrikStatusPoller');
+//const FabrikStatusPoller = require('../fabrik/FabrikStatusPoller');
```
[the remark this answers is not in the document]
sauravmndl: done

broker/lib/fabrik/DirectorManager.js
```diff
@@ -725,20 +748,28 @@ class DirectorManager extends BaseManager {
...
-          })
-          .then(() => {
-            metaUpdated = true;
-            const instanceInfo = _.chain(data)
+            instanceInfo = _.chain(data)
               .pick('tenant_id', 'backup_guid', 'instance_guid', 'agent_ip', 'service_id', 'plan_id')
               .set('deployment', deploymentName)
               .set('started_at', backupStartedAt)
               .value();
+            return DirectorManager.registerBnRStatusPoller({
```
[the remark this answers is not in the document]
kamath-prasad: Ok , got it . thanks for the explanation !

broker/lib/jobs/BnRStatusPollerJob.js
kamath-prasad: ok , makes sense.

broker/lib/jobs/BnRStatusPollerJob.js
kamath-prasad: ok , then in that case I think we need to make few logs / variables etc. generic . , will comment.

broker/lib/jobs/BnRStatusPollerJob.js
kamath-prasad: can you use : 

`Abort _.get(options, 'operation') timed out on : ${deployment} for guid : ${backup_guid}. Flagging backup _.get(options, 'operation') as complete`

broker/lib/jobs/BnRStatusPollerJob.js
```diff
@@ -0,0 +1,170 @@
...
+
```
kamath-prasad: Also why couldn't we refactor and reuse OperationStatusPoller itself ? Because it already has everything needed to support this use case , right ?

sauravmndl: OperationStatusPoller interacts with OOBManager instead of DirectorManager. Also for objectstore access the paths of metadata is different. Would be if-else in all steps. Factory pattern would lead to two more extra classes for cleaner approach. So, separated in two jobs. Also deployment lock-unlock is not present in oob deployments.

broker/lib/jobs/BnRStatusPollerJob.js
sauravmndl: done.

common/constants.js
```diff
@@ -64,6 +64,7 @@ module.exports = Object.freeze({
     BACKUP_STATUS_POLLER: 'FabrikStatusPoller',
```
kamath-prasad: we can remove this , right ?

kamath-prasad:
Few more comments , which I missed last night : 

- Why have you not removed FabrikStatusPoller.js ?
- How will reconciliation be done for existing backups under execution when the new code goes in ?

broker/lib/fabrik/DirectorManager.js
```diff
@@ -668,7 +691,7 @@ class DirectorManager extends BaseManager {
     const backupStartedAt = new Date().toISOString();
     const data = _
       .chain(opts)
-      .pick('service_id', 'plan_id', 'organization_guid', 'instance_guid', 'username')
+      .pick('deployment', 'service_id', 'plan_id', 'organization_guid', 'instance_guid', 'username')
```
muralisc: is the  `data.deployment` used ? 
Its being set again at https://github.com/cloudfoundry-incubator/service-fabrik-broker/pull/275/files#diff-b2c19de3bcc8e2e1fb90955f336cdbaeR757
and is not used anywhere else above.
If so can we remove it from here ?

sauravmndl: Okay. would check if data.deployment is nowhere used. If not would remove.

common/constants.js
sauravmndl: Supporting on going backups using FabrikStatusPoller. FabrikStatusPoller would poll status for on-going backups before rolling out BnRStatusPoller. FabrikStatusPoller can be remove anytime after 24 hrs of rolling out of BnRStatusPoller.

PR #277, opened 2018-05-23
Bug/dashboard

jayeshjk:
* Bug description: 
1. Dashboard rendering fails when service plan is updated as the corresponding dashboard URL is not updated in cloud controller.
2.  In case of migration from one bosh to other, dashboard rendering code might query to older bosh to get tasks hence the dashboard rendering would fail.

* To resolve the part 1 of the bug, a call to cloud controller to get the current plan of the instance was added in both DockerInstance and DirectorInstance.

* To resolve part 2 of the bug, a new function getTasksForDashboard was added in BoshDirectorClient which queries only correct director instead of looping through all the primaryConfigs as original getTasks does. The 'correct' director is fetched from the cache using deployment name.

* Minor modification in docker dashboard acceptance test to accommodate extra call to CC.

* Added one unit test for getTasksForDashboard.

### DIFF
--- a/broker/lib/bosh/BoshDirectorClient.js
+++ b/broker/lib/bosh/BoshDirectorClient.js
@@ -621,46 +621,37 @@ class BoshDirectorClient extends HttpClient {
 
   /*  Task operations */
 
-  getTasksForDashboard(options) {
+  getTasks(options, fetchDirectorForDeployment) {
     const query = _.assign({
       limit: 1000
     }, options);
+    let shouldFetch = fetchDirectorForDeployment === true;
+    var configPromise = new Promise((resolve, reject) => {
+      if (shouldFetch) {
+        var fetchedConfigs = [];
+        this.getDirectorConfig(options.deployment)
+          .then(directorConfig => {
+            fetchedConfigs.push(directorConfig);
+            resolve(fetchedConfigs);
+          });
+      } else {
+        resolve(this.primaryConfigs);
+      }
+    });
 
-    //Get the corresponding directorConfig first and then request to that config directly.
-    return this.getDirectorConfig(options.deployment)
-      .then(configForDeployment => {
-        return this.makeRequestWithConfig({
-          method: 'GET',
-          url: '/tasks',
-          qs: _.pick(query, ['limit', 'state', 'deployment'])
-        }, 200, configForDeployment);
+    return configPromise.map(directorConfig => {
+        return this
+          .makeRequestWithConfig({
+            method: 'GET',
+            url: '/tasks',
+            qs: _.pick(query, ['limit', 'state', 'deployment'])
+          }, 200, directorConfig)
+          .then(res => JSON.parse(res.body))
+          .map(task => {
+            task.id = `${options.deployment}_${task.id}`;
+            return task;
+          });
       })
-      .then(res => JSON.parse(res.body))
-      .map(task => {
-        task.id = `${options.deployment}_${task.id}`;
-        return task;
-      })
-      .reduce((all_tasks, tasks) => all_tasks.concat(tasks), []);
-  }
-
-  getTasks(options) {
-    const query = _.assign({
-      limit: 1000
-    }, options);
-
-    return Promise
-      .map(this.primaryConfigs, directorConfig => this
-        .makeRequestWithConfig({
-          method: 'GET',
-          url: '/tasks',
-          qs: _.pick(query, ['limit', 'state', 'deployment'])
-        }, 200, directorConfig)
-        .then(res => JSON.parse(res.body))
-        .map(task => {
-          task.id = `${options.deployment}_${task.id}`;
-          return task;
-        })
-      )
       .reduce((all_tasks, tasks) => all_tasks.concat(tasks), []);
   }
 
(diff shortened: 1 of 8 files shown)


broker/lib/bosh/BoshDirectorClient.js
```diff
@@ -621,6 +621,28 @@ class BoshDirectorClient extends HttpClient {
 
   /*  Task operations */
 
+  getTasksForDashboard(options) {
```
kamath-prasad: as discussed , pls see if we can re-use existing `getTask(option)` but refactoring it to `getTask(option,fetchDirectorForDeployment)`

kamath-prasad:
@abh1kg and @sauravmndl : please review this PR.

broker/lib/bosh/BoshDirectorClient.js
```diff
@@ -621,24 +621,37 @@ class BoshDirectorClient extends HttpClient {
 
   /*  Task operations */
 
-  getTasks(options) {
+  getTasks(options, fetchDirectorForDeployment) {
     const query = _.assign({
       limit: 1000
     }, options);
+    let shouldFetch = fetchDirectorForDeployment === true;
```
sauravmndl: I think, not required to introduce another variable for the falg.

broker/lib/bosh/BoshDirectorClient.js
```diff
@@ -621,24 +621,37 @@ class BoshDirectorClient extends HttpClient {
...
+    var configPromise = new Promise((resolve, reject) => {
+      if (shouldFetch) {
```
sauravmndl: Directly you can do if(fetchDirectorForDeployment){...}

broker/lib/bosh/BoshDirectorClient.js
```diff
@@ -621,24 +621,37 @@ class BoshDirectorClient extends HttpClient {
 
   /*  Task operations */
 
-  getTasks(options) {
+  getTasks(options, fetchDirectorForDeployment) {
     const query = _.assign({
       limit: 1000
```
sauravmndl: should it be hard-coded or config driven ?or  you can put it in constant. Just give a think. If you strongly feel it should be like this. It's fine.

broker/lib/bosh/BoshDirectorClient.js
```diff
@@ -621,24 +621,37 @@ class BoshDirectorClient extends HttpClient {
...
     }, options);
+    let shouldFetch = fetchDirectorForDeployment === true;
+    var configPromise = new Promise((resolve, reject) => {
```
sauravmndl: I think, not required to create a promise. You can directly write like

return Promise
     .try(()=>shouldFetch ? this.getDirectorConfig(options.deployment) : this.primaryConfigs)
      .map(directorConfig => {
+        return this
+          .makeRequestWithConfig({
+            method: 'GET',
+            url: '/tasks',
+            qs: _.pick(query, ['limit', 'state', 'deployment'])
+          }, 200, directorConfig)
+          .then(res => JSON.parse(res.body))
+          .map(task => {
+            task.id = `${options.deployment}_${task.id}`;
+            return task;
+          });
+      })
       .reduce((all_tasks, tasks) => all_tasks.concat(tasks), []);

broker/lib/bosh/BoshDirectorClient.js
```diff
@@ -621,24 +621,37 @@ class BoshDirectorClient extends HttpClient {
...
-        })
-      )
+    return configPromise.map(directorConfig => {
+        return this
+          .makeRequestWithConfig({
+            method: 'GET',
+            url: '/tasks',
+            qs: _.pick(query, ['limit', 'state', 'deployment'])
+          }, 200, directorConfig)
+          .then(res => JSON.parse(res.body))
```
sauravmndl: You can put the request making part inside a method.

abh1kg:
@sauravmndl has requested for changes. Further, there are some JSHint issues with this pull.

broker/lib/fabrik/DirectorInstance.js
```diff
@@ -321,18 +322,28 @@ class DirectorInstance extends BaseInstance {
...
-        files: [{
-          id: 'status',
-          title: 'Status',
-          language: 'yaml',
-          content: yaml.dump(deploymentInfo)
-        }]
-      }));
+      .spread((instance, deploymentInfo) => {
+        return Promise.all([
+          instance,
```
abh1kg: Using a Promise.all against an already resolved variable looks a bit odd. Can this be optimized a bit further? Essentially, here, we are depending on the corollary that a constant is a resolved promise.
Can we not club the plan fetch operation with the instance fetch promise above? 
Something like 
```
this.cloudController.getServiceInstance(this.guid).then(instanceInfo => {
  this.cloudController.getServicePlan(instanceInfo.entity.service_plan_guid, {})
})
```

Of course, you need to preserve `instanceInfo` as well

broker/lib/fabrik/DirectorInstance.js
```diff
@@ -318,21 +319,34 @@ class DirectorInstance extends BaseInstance {
...
-        instance: _.set(instance, 'task', deploymentInfo),
-        files: [{
-          id: 'status',
-          title: 'Status',
-          language: 'yaml',
-          content: yaml.dump(deploymentInfo)
-        }]
-      }));
+      .spread((consolidated, deploymentInfo) => {
+        let instance = consolidated.instance;
```
sauravmndl: You can rename 'consolidated' ==> 'instanceInfo'/'instanceDetails'

jayeshjk:
Yes, I will look into why the build is failing, I think it's the trufflehog issue. Will finalise it ASAP.

PR #283, opened 2018-05-24
Feature/seccomp

amitmalav:
Added `add_seccomp_filters.c` file
It sets seccomp filters at runtime for the running process
Provided flag `enable_syscall_filters ` to determine whether to run processes in sandbox

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -38,7 +38,7 @@ script:
      # truffleHog check fails to detect changed because travis merge creates a detached HEAD
      # Creating a branch to fix this issue
      git checkout -b travis-merge
-     INITIAL_COMMIT="fb2aeb0a16da5a25f8063558ff9f7159a3fe8c0f"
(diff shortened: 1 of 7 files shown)


amitmalav:
**Please don't squash and merge** Do rebase and merge

shashankmjain:
What is the criteria of whitelisting the sys calls?

amitmalav:
@shashankmjain 
Blacklisting syscalls is just a broad preventive measure.
More from operation pov, the concern was how can `service-fabrik` operators review action scripts that are being on-boarded effectively?
For example, If `service-fabrik` does not want to allow network access from `service-fabrik` vm then all the network related syscalls can be blacklisted. But issues like, sf wants to allow network calls specific to some network, this still remains unsolved.
The default `whitelisted_syscall` list has all the syscalls that are required by execution framework itself, like spawning a child process and run the on boarded scripts and get the result from the script.

PR #287, opened 2018-05-29
Change/preupdate

### LINKED ISSUE #286: Preupdate agent api will fail if agent credentials are updated
If broker-agent credentials are updated in service catalog, updates can't go through as preupdate agent api will get UnAuthorised error from agent if requested with updated password.

To fix this issue, preupdate call to agent can be made with agent credential from previous manifest (which is currently deployed manifest)

amitmalav:
Fixes #286

### DIFF
--- a/broker/lib/fabrik/DirectorManager.js
+++ b/broker/lib/fabrik/DirectorManager.js
@@ -260,9 +260,12 @@ class DirectorManager extends BaseManager {
       .set('plan_id', this.plan.id)
       .set('agent_properties', _.omit(agentProperties, 'auth', 'provider'))
       .value();
+    const agentCredBeforeUpdate = utils.getBrokerAgentPropertiesFromManifest(context.params.previous_manifest);
+    // Making preupdate request with agent credentials from previous manifest
+    // In case agent passwords are being updated as part of this update
     return this
       .getDeploymentIps(deploymentName)
-      .then(ips => this.agent.preUpdate(ips, context))
+      .then(ips => this.agent.preUpdate(ips, context, agentCredBeforeUpdate))
       .catch(FeatureNotSupportedByAnyAgent, ServiceInstanceNotOperational, err => {
         logger.debug('+-> Caught expected error of feature \'preUpdate\':', err);
         return {};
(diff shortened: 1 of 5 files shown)


broker/lib/fabrik/DirectorManager.js
```diff
@@ -260,9 +260,12 @@ class DirectorManager extends BaseManager {
       .set('plan_id', this.plan.id)
       .set('agent_properties', _.omit(agentProperties, 'auth', 'provider'))
       .value();
+    const agentCredsBeforeUpdate = utils.getBrokerAgentCredsFromManifest(context.params.previous_manifest);
```
jagadish-kb: Its better if this is moved to agent. Preupdate method anyways is sent with context. Director need not be aware of this. This logic should be within agent. i.e. which creds to use.

kamath-prasad: @jagadish-kb  : Didn't get you . are you saying that the /info call returns the creds to be used by broker while calling the /preupdate API ?

jagadish-kb: what i meant is the line 263, can be pushed into the function Agent.preUpdate(). Reason being, directorManager should not care what creds are to be used and ideally thats something that must be encapsulated within the Agent class. Dont want that responsibility spilling outside of it.

kamath-prasad: Ok , got it . Yes , it makes sense. When you said agent , thought it was service agent :)

amitmalav: Makes sense. I'll make the change.

amitmalav:
@ashishjain14 You are right, we should use agent credentials from currently deployed manifest only even for other operations like backup and restore. But that will increase load on bosh as we will be making one call to bosh every time we make request to bosh.
One solution that @jagadish-kb proposed is to force-update whenever services team make changes in agent credentials.

broker/lib/fabrik/DirectorManager.js
amitmalav: Done!

ashishjain14:
@amitmalav since we already maintain the deployments cache, why can't we manage agent credentials as well? Secondly is force update available as a feature?

PR #288, opened 2018-05-29
Fix parsing of space_guid from context during security group creation of Update Service flow

subhankarc:
- ensureTenantId accepts whole of options and not only space_guid

### DIFF
--- a/broker/lib/fabrik/CfPlatformManager.js
+++ b/broker/lib/fabrik/CfPlatformManager.js
@@ -69,7 +69,7 @@ class CfPlatformManager extends BasePlatformManager {
       .tap(() => logger.info('+-> Security group exists'))
       .catch(SecurityGroupNotFound, () => {
         logger.warn('+-> Security group does not exist. Trying to create it again.');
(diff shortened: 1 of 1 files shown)


broker/lib/fabrik/CfPlatformManager.js
```diff
@@ -69,7 +69,7 @@ class CfPlatformManager extends BasePlatformManager {
       .tap(() => logger.info('+-> Security group exists'))
       .catch(SecurityGroupNotFound, () => {
         logger.warn('+-> Security group does not exist. Trying to create it again.');
-        return this.ensureTenantId(options.context.space_guid)
+        return this.ensureTenantId(options)
```
sauravmndl: One more fix we have to do ensureTenantId(options) {
    return Promise
      `.try(() => _.get(options, 'context.space_guid') ? options.context.space_guid : this.cloudController`
        .getServiceInstance(options.guid)
        .then(instance => instance.entity.space_guid)
      );
  }

PR #289, opened 2018-05-30
Updating docker port registry in case of start error

sauravmndl:
In HA scenario of SF broker, when failover happens new master doesn't have updated docker port registry info.
Suppose old master created a container allocated port e.g. 36555. After failover new master doesn't have 36555 in its cache entry. Now if new creation of container happens and 36555 is taken from free ports, container creation might fail if goes to same host.

To mitigate if container creation fails now we update cache.

### DIFF
--- a/broker/lib/fabrik/DockerInstance.js
+++ b/broker/lib/fabrik/DockerInstance.js
@@ -245,6 +245,7 @@ class DockerInstance extends BaseInstance {
           .startContainer()
           .catch(DockerError.ServerError, err => self
             .removeContainer()
+            .then(() => docker.updatePortRegistry())
             .throw(new ContainerStartError(err.message))
           )
         );
--- a/test/test_broker/acceptance/service-broker-api.instances.docker.spec.js
+++ b/test/test_broker/acceptance/service-broker-api.instances.docker.spec.js
@@ -84,6 +84,39 @@ describe('service-broker-api', function () {
             });
         });
 
+        it('returns 201 Created - start fails once internally', function () {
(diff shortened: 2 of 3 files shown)


broker/lib/fabrik/DockerInstance.js
```diff
@@ -245,6 +245,7 @@ class DockerInstance extends BaseInstance {
           .startContainer()
           .catch(DockerError.ServerError, err => self
             .removeContainer()
+            .then(() => docker.updatePortRegistry())
```
jagadish-kb: Dont you get a specific error in case of port already in use? Ideally we want to updatePortRegistry only on that scenario and not all error scenarios right?

ketakig12: As Jagadish said, if possible, we can update cache on specific issue.
Also, this means, the create service instance still fails first time and next time onwards, it will not happen. Not sure if this is acceptable.

sauravmndl: I tested and the port already in use issue comes as 500 and start api of docker throws following codes
    statusCodes: {
      200: true, // unofficial, but proxies may return it
      204: true,
      304: 'container already started',
      404: 'no such container',
      500: 'server error'
    },

jagadish-kb: I am ok, with the above in case we are not getting specific error codes. Approving this PR.

ketakig12: Also, as we retry 3 times to create container, it will not be create instance failure from user point of view. Hence, sounds good. Approving PR.

PR #290, opened 2018-05-31
Graceful Riemann connection and error handling

ketakig12:
Since riemann client emits events on **connect**, after calling `createClient`, we should wait for initialization to happen properly i.e. for **connect** event before sending event. Else they might get lost.
Hence events coming during this initialization time also need to be queued and retried after connect event is emitted.

### DIFF
--- a/common/constants.js
+++ b/common/constants.js
@@ -12,6 +12,11 @@ module.exports = Object.freeze({
     MAX_QUEUE_SIZE: 100,
     MAX_SEND_RETRIES: 2
   },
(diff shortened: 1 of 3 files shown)


ashishjain14:
@ketakig12 how are we handling the queue in case of a failover?

ketakig12:
@ashishjain14 We queue the requests only when riemann client is getting initialized or we fail to send event. Currently, the events in the queue will be lost if failover happens, but if we don't queue them, they anyways error out and are lost. In case, we want to persist them even after failover, we will need a persistent queue and not in-memory queue.

PR #291, opened 2018-06-02
Allow context-based rate limits against BOSH task submissions

abh1kg:
- Broker performs percentage-based workers per type of operation
- Use etcd store for staggering director requests
- Use timer for repeated pull and execution of tasks


abh1kg:
@Reviewers: Requesting a preliminary review for this change. If you feel anything needs more attention, please let me know. Haven't added tests to this pull yet, so expect build failures

broker/lib/fabrik/DirectorManager.js
```diff
@@ -177,7 +189,145 @@ class DirectorManager extends BaseManager {
...
+    _.omit(params, '_serviceFabrikDbUpdate');
+
+    const decisionMaker = {
+      'shouldRunNow': false,
+      'cached': false,
+      'dbUpdate': false,
+      'policyApplicable': config.enable_bosh_rate_limit
+    };
+
+    return this.executePolicy(scheduled, action, deploymentName, dbUpdate, policyApplicable)
```
kamath-prasad: didnt understand how this function takes arguments ? should decisionMaker be used as an argument to this function ?

broker/lib/fabrik/DirectorTaskPoller.js
```diff
@@ -0,0 +1,51 @@
...
+
+const bosh = require('../bosh');
+const DirectorManager = require('./DirectorManager');
+const logger = require('../logger');
+const catalog = require('../models/catalog');
+const pubsub = require('pubsub-js');
+const CONST = require('../constants');
+const config = require('../config');
+const boshCache = bosh.BoshOperationCache;
+const TIME_POLL = 1 * 60 * 1000;
```
kamath-prasad: maybe have a config for this ?

broker/lib/bosh/BoshDirectorClient.js
```diff
@@ -344,7 +344,59 @@ class BoshDirectorClient extends HttpClient {
...
+      return this.getDirectorConfig(deploymentName);
+    }
+  }
+
+  /** 
+   * get the current tasks in the director (in processing, cancelling state)
+   * task count should be retrieved for ALL types of operations
+   * 
+   */
+  getCurrentTasks(action, directorConfig) {
```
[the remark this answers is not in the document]
abh1kg: Should be possible...will make this change

broker/lib/bosh/BoshOperationCache.js
```diff
@@ -0,0 +1,243 @@
...
+
+  /**
+   * Fetches a limited number of BOSH deployment names from the Etcd Store
+   * 
+   * @param {Number of entries to fetch} numEntries 
+   */
+  getNEntries(numEntries) {
+    logger.debug(`Getting the first ${numEntries} entries in cache`);
+    const wrapper = new Promise((resolve, reject) => {
+      etcdConnector().getAll().prefix(CACHE_KEY).sort("Create", "None").limit(numEntries).keys().then(out => {
```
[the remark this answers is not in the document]
abh1kg: These are the sort + target options for Etcd3 (https://mixer.github.io/etcd3/classes/builder_.multirangebuilder.html#sort ). 
Will add them to constants

broker/lib/bosh/BoshOperationCache.js
```diff
@@ -0,0 +1,243 @@
...
+    return wrapper;
+  }
+
+  /**
+   * Fetches all BOSH deployment names in the Etcd store currently
+   */
+  getDeploymentNames() {
+    logger.debug('Getting the current deployment cache...');
+    const wrapper = new Promise((resolve, reject) => {
+      etcdConnector().getAll().prefix(CACHE_KEY).sort('Create', 'None').keys().then(out => {
```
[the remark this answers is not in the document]
abh1kg: Will do

broker/lib/bosh/BoshDirectorClient.js
```diff
@@ -344,7 +344,59 @@ class BoshDirectorClient extends HttpClient {
...
+    };
+    return this.makeRequestWithConfig({
+      method: 'GET',
+      url: '/tasks',
+      qs: query
+    }, 200, directorConfig).then(out => {
+      // out is the array of currently running tasks
+      let currentTotal = out.length;
+      let taskGroup = _.groupBy(out, (entry) => {
+        if (entry.context_id === CONST.BOSH_FABRIK_SCHEDULED_OP) {
```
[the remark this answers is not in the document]
abh1kg: Yes... will make this change

broker/lib/fabrik/DirectorManager.js
abh1kg: the "decisionMaker" variable is updated based on the operations. There's a bug here that I'll push in the next change- the `policyApplicable` variable is not set. It should follow the feature value in the config

broker/lib/fabrik/DirectorTaskPoller.js
abh1kg: Will do...

broker/lib/fabrik/DirectorManager.js
```diff
@@ -177,7 +189,145 @@ class DirectorManager extends BaseManager {
...
+      .return(output);
+  }
+
   createOrUpdateDeployment(deploymentName, params, args) {
+    logger.info(`Checking rate limits against bosh for deployment `);
+    const previousValues = _.get(params, 'previous_values');
+    const action = _.isPlainObject(previousValues) ? CONST.OPERATION_TYPE.UPDATE : CONST.OPERATION_TYPE.CREATE;
+    const scheduled = _.get(params, 'scheduled') || false;
+    const dbUpdate = _.get(params, '_serviceFabrikDbUpdate') || false;
+    _.omit(params, 'scheduled');
```
ashishjain14: What is dbupdate here?

broker/lib/fabrik/DirectorManager.js
```diff
@@ -177,7 +189,145 @@ class DirectorManager extends BaseManager {
...
+    const decisionMaker = {
+      'shouldRunNow': false,
+      'cached': false,
+      'dbUpdate': false,
+      'policyApplicable': config.enable_bosh_rate_limit
+    };
+
+    return this.executePolicy(scheduled, action, deploymentName, dbUpdate, policyApplicable)
+      .then(shouldRunNow => {
+        if (!decisionMaker.policyApplicable) {
```
ashishjain14: Is this check again required since it was already checked in executePolicy method?

broker/lib/bosh/BoshOperationCache.js
```diff
@@ -0,0 +1,243 @@
...
+    return wrapper;
+  }
+
+  /**
+   * Deletes multiple BOSH deployment names from the Etcd store
+   * 
+   * @param {Deployment Names} deployments 
+   */
+  deleteDeploymentsFromCache(...deployments) {
+    const deploymentIdentifiers = _.flattenDeep(deployments);
```
ashishjain14: Why would a case arise where we may need to delete multiple deployments from cache?

broker/lib/bosh/BoshOperationCache.js
```diff
@@ -0,0 +1,243 @@
...
+    return Promise.all(deletePromises);
+  }
+
+  /**
+   * Stores the BOSH task ID corresponding to the service instance ID when the task is submitted to the BOSH Director
+   * 
+   * @param {Service Instance ID} serviceInstanceId 
+   * @param {BOSH Task ID corresponding to the operation} taskId 
+   */
+  storeBoshTask(serviceInstanceId, taskId) {
```
ashishjain14: Is this for the case where we do not want to hit the BOSH director?

broker/lib/fabrik/DirectorManager.js
abh1kg: This signifies the deployment update for the mongodb instance. For the MongoDB update, rate limits do not apply

broker/lib/fabrik/DirectorManager.js
abh1kg: This is needed because the `executePolicy` will return `runNow = true` but since the flag `policyApplicable` is responsible for the entire execution, it needs to be used subsequently

broker/lib/bosh/BoshOperationCache.js
abh1kg: Yes...

broker/lib/bosh/BoshOperationCache.js
abh1kg: This was put in for the future monitoring requirements.

broker/lib/fabrik/DirectorManager.js
```diff
@@ -177,7 +189,146 @@ class DirectorManager extends BaseManager {
...
 
+  executePolicy(scheduled, action, deploymentName, dbUpdate, policyApplicable) {
+    if (!policyApplicable) {
+      return Promise.resolve(true);
+    }
+    if (dbUpdate) {
+      return Promise.resolve(true);
+    }
+    const targetDirectorConfig = this.director.getDirectorForOperation(action, deploymentName);
+    this.director.getCurrentTasks(action, targetDirectorConfig).then(out => {
```
muralisc: A return missing here ?

broker/lib/fabrik/DirectorManager.js
```diff
@@ -177,7 +189,146 @@ class DirectorManager extends BaseManager {
...
+      let maxTasks = targetDirectorConfig.max_workers;
+      if (allTasks >= maxTasks) {
+        //no slots left anyway
+        return false;
+      }
+      if (scheduled) {
+        currentTasks = out.scheduled;
+        maxWorkers = targetDirectorConfig.policies.scheduled.max_workers;
+      } else {
+        currentTasks = out.user;
```
muralisc: can we rename `out` to `tasksCount` ?

broker/lib/bosh/BoshOperationCache.js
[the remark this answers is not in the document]
kamath-prasad: Ok , thanks for the info !!

broker/lib/bosh/BoshDirectorClient.js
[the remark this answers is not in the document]
kamath-prasad: ok.

broker/lib/bosh/BoshOperationCache.js
```diff
@@ -0,0 +1,243 @@
...
+  /**
+   * Stores the deployment operation (user-triggered) into the Etcd store as a JSON document
+   * Key is prefixed with a specific format and the deployment name is used as the identifier
+   * 
+   * @param {Service plan ID for the service instance} planId 
+   * @param {Calculated bosh deployment name} deploymentName 
+   * @param {Parameters used for the operation} params 
+   * @param {Arguments used for the operation} args 
+   */
+  store(planId, deploymentName, params, args) {
```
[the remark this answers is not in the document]
kamath-prasad: just trying to understand , since all BOSH based operations will put the instance/deployment in "In progress" state user cannot trigger lets say create followed by update or delete etc.. I think hence the current key structure should work.

Similarly thing should hold true for fabrik initiated operations.

maybe we need a differentiator between user triggered vs fabrik triggered for the same deployment . But even here since there is a lock on the instance, dont see both occurring in tandem.

broker/lib/fabrik/DirectorManager.js
kamath-prasad: yes correct. now it looks fine.

broker/config/settings.yml
```diff
@@ -430,6 +431,12 @@ defaults: &defaults
     skip_ssl_validation: true
     default_task_poll_interval: 10000
     lock_deployment_max_duration:  86400 # lock ttl for deployments in seconds. Its kept at 24 hrs - 60 * 60 * 24
+    max_workers: 24
+    policies:
+      user:
+        max_workers: 50
+      scheduled:
+        max_workers: 50
```
ashishjain14: Currently the approach is to give 50 percent workers each to user triggered and scheduled task. But this approach still may have a limitation on question of fare share for various request served by SF. For example if a user triggers 12 long running upgrades, followed by a create. This will result in create being blocked until all the upgrades are completed.

broker/lib/bosh/BoshOperationCache.js
ashishjain14: Why is that we have to maintain a task submitted to BOSH in the cache? The understanding is that the task which cannot be currently published to BOSH because of not enough capacity are kept in the cache

broker/lib/bosh/BoshOperationCache.js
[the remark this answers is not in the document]
ashishjain14: But there could be cases where we will not have mutually exclusive operations for example bind and backup are both read operations as per the SF 2.0 locking implementation. If these are the 2 requests which would simultaneously arrive would not have same key name be a problem.

broker/lib/bosh/BoshOperationCache.js
abh1kg: The task ID needs to be put into Etcd because:
- If the request is staggered initially, the response `ACCEPTED` (202) would not contain the task ID
- CC will call the last operation endpoint of the broker repeatedly with the information (this needs to indicate that the request was cached)
- The poller would submit to BOSH at one point in time
- The subsequent last operation call must figure out the task ID and the only way to do so is via ETCD

broker/lib/fabrik/DirectorManager.js
abh1kg: Will do

broker/config/settings.yml
abh1kg: Yes but that's the problem of BOSH having limited resources. True fair share cannot be achieved either way

broker/lib/bosh/BoshOperationCache.js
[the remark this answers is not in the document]
abh1kg: But, bind and backup never go via the Director. Exclusively, operations that create bosh tasks will be limited via the concept. Everything else happens as is

broker/config/settings.yml
jagadish-kb: @ashishjain14 @kamath-prasad @abh1kg 
I thought the original intention of the problem was to address what ashish mentioned in the above. Else what is the use of queuing in SF Broker side, we might as well let it queue on BOSH. Dont see the real benefit then?

My understanding was, if there are 24 workers then say 50-70% of it would be max used by major version upgrade. So at any point in time, then we must check for processing/queued tasks and if their combined value is less than the configured percentage only then we push that request to BOSH. Else queue it. As an example, 
out of 24 workers, assume we say 20 workers can be used for major version update and rest 4 workers we leave it free for any other adhoc operation that comes in.

Now before pushing a major version upgrade we see how many tasks are in processing/queued. If this count is less than 20, then we let this upgrade to bosh, else queue it. This way, we will push major version upgrade only when we feel it does not hog BOSH completely and it leaves for other operations also to take place.

This is the main use case which mongodb wants us to solve. May be please double check with MongoDB team.

broker/lib/fabrik/DirectorManager.js
abh1kg: Fixed

broker/lib/fabrik/DirectorManager.js
abh1kg: Done

abh1kg:
@jagadish-kb : Changes for per-op limits are done. Could you please review again?
@subhankarc : The lock status poller is added with this pull. Could you please review?

subhankarc:
@abh1kg You may want to raise the BOSH-release changes before merging it.

broker/lib/fabrik/DirectorManager.js
```diff
@@ -177,7 +189,179 @@ class DirectorManager extends BaseManager {
...
+            if (inCache) {
+              decisionMaker.cached = true;
+            }
+            return Promise.resolve();
+          });
+        }
+        if (!scheduled) {
+          return Promise.resolve(true);
+        }
+        throw new DeploymentDelayed(deploymentName);
```
jagadish-kb: why is this exception being thrown in case of scheduled and should not be runNow? for user tasks why is exception not being thrown?

broker/lib/fabrik/DirectorManager.js
```diff
@@ -177,7 +189,179 @@ class DirectorManager extends BaseManager {
...
+        }
+        const shouldRunNow = checkResults.shouldRunNow;
+
+        if (!decisionMaker.policyApplicable) {
+          return Promise.resolve();
+        }
+        if (dbUpdate) {
+          decisionMaker.shouldRunNow = true;
+          decisionMaker.dbUpdate = true;
+          return Promise.resolve();
```
jagadish-kb: explicitly `return false`

broker/lib/fabrik/DirectorManager.js
```diff
@@ -177,7 +189,179 @@ class DirectorManager extends BaseManager {
...
+        } else {
+          throw new DeploymentDelayed(deploymentName);
+        }
+      })
+      .then(() => this._createOrUpdateDeployment(deploymentName, params, args, scheduled))
+      .then(taskId => {
+        const out = {
+          'cached': false
+        };
+        if (decisionMaker.cached) {
```
jagadish-kb: when will this be true? if cached wont you throw the exception?see LN 324 - that should ensure no entries exist in etcd if its to run now?

broker/lib/fabrik/DirectorManager.js
```diff
@@ -177,7 +189,179 @@ class DirectorManager extends BaseManager {
...
+          return Promise.resolve();
+        }
+        if (dbUpdate) {
+          decisionMaker.shouldRunNow = true;
+          decisionMaker.dbUpdate = true;
+          return Promise.resolve();
+        }
+        if (shouldRunNow) {
+          decisionMaker.shouldRunNow = shouldRunNow;
+          return boshOperationCache.containsDeployment(deploymentName).then(inCache => {
```
jagadish-kb: if it should run now, it is also in etcd queued? is this a new operation and the one in etcd is for older operation?

broker/lib/fabrik/DirectorManager.js
```diff
@@ -177,7 +189,179 @@ class DirectorManager extends BaseManager {
...
+    const action = _.isPlainObject(previousValues) ? CONST.OPERATION_TYPE.UPDATE : CONST.OPERATION_TYPE.CREATE;
+    const scheduled = _.get(params, 'scheduled') || false;
+    const dbUpdate = _.get(params, '_serviceFabrikDbUpdate') || false;
+    const policyApplicable = config.enable_bosh_rate_limit;
+    _.omit(params, 'scheduled');
+    _.omit(params, '_serviceFabrikDbUpdate');
+
+    const decisionMaker = {
+      'shouldRunNow': false,
+      'cached': false,
```
jagadish-kb: instead of `cached` its better to name it `queued`

broker/lib/fabrik/DirectorManager.js
```diff
@@ -177,7 +189,179 @@ class DirectorManager extends BaseManager {
...
+            return Promise.resolve();
+          });
+        }
+        if (!scheduled) {
+          return Promise.resolve(true);
+        }
+        throw new DeploymentDelayed(deploymentName);
+      })
+      .then((storeInCache) => {
+        if (storeInCache) {
```
jagadish-kb: if we are storing in cache, then we wont run immediately can collapse the promise chain. Seems unnecessary long chain. For ex:
```
if(queueRequest){
   //save it in etcd
  return {
          'cached': true
    };
} else{
   return this._createOrUpdateDeployment(deploymentName, params, args, scheduled);
}
```

broker/lib/fabrik/DirectorManager.js
```diff
@@ -177,7 +189,179 @@ class DirectorManager extends BaseManager {
...
+              return Promise.resolve();
+            })
+            .catch(err => {
+              logger.error('Error updating etcd cache', err);
+              throw new CacheUpdateError(deploymentName);
+            });
+        }
+      })
+      .then(() => {
+        if (decisionMaker.shouldRunNow && decisionMaker.cached) {
```
jagadish-kb: But once queued operation is complete, its anyways cleared. when will this occur? is this to handle failure in delete of queued entries? if such was the case, then it would be better to have a timestamp when the entry was queued and ensure its overrwritten or invalidated instead of explicit delete like the below to handle delete failure.

broker/lib/fabrik/DirectorManager.js
```diff
@@ -177,7 +189,179 @@ class DirectorManager extends BaseManager {
...
+          //the deployment was cached in etcd earlier; remove it from cache and proceed
+          return boshOperationCache.deleteDeploymentFromCache(deploymentName);
+        }
+      })
+      .then(() => {
+        if (decisionMaker.directorError) {
+          throw new DeploymentDelayed(deploymentName);
```
jagadish-kb: No need to throw this exception. Can be handled in a better way from `executePolicy`

broker/lib/fabrik/LockStatusPoller.js
```diff
@@ -0,0 +1,89 @@
...
+  /**
+   * Starts the timer
+   */
+  start() {
+    if (this.interval) {
+      throw new Error('timer already started');
+    }
+    this.interval = setInterval(() => {
+      this.lock = client.lock(this.lockName);
+      this.lock.do(() => {
```
jagadish-kb: everytime you are getting the lock, wont it reset the reference in LN 64? so how does this truly work in HA setup? May be  am not aware of how etcd lock works. If you can add some comments to this code here, it will help in readability on what the above is trying to achieve.

broker/lib/fabrik/DirectorTaskPoller.js
```diff
@@ -0,0 +1,49 @@
...
+    return boshCache.getDeploymentNames().mapSeries(deploymentName => {
+        return Promise.try(() => {
+            return boshCache.getDeploymentByName(deploymentName);
+          })
+          .then(cached => {
+            let catalogPlan = catalog.getPlan(cached.plan_id);
+            return DirectorManager.load(catalogPlan).createOrUpdateDeployment(deploymentName, cached.params, cached.args);
+          })
+          .catch(e => {
+            logger.error(`Error in automated deployment for ${deploymentName}`, e);
```
jagadish-kb: automated deployment?

broker/lib/fabrik/DirectorTaskPoller.js
```diff
@@ -0,0 +1,49 @@
...
+      })
+      .catch(e => logger.error('error in processing deployments', e));
+  }
+}
+
+const pollerInstance = new DirectorTaskPoller();
+
+pubsub.subscribe(CONST.TOPIC.APP_STARTUP, (eventName, eventInfo) => {
+  logger.debug('-> Received event ->', eventName);
+  if (eventInfo.type === 'external' && config.enable_bosh_rate_limit) {
```
jagadish-kb: this should be 'internal' as create/update are part of internal app.

common/config.js
```diff
@@ -75,6 +75,27 @@ function completeDirectorConfig(director) {
...
+    }
+
+    if (_.sum([userCreateWorkers, userUpdateWorkers, userDeleteWorkers, autoWorkers]) > maxWorkers) {
+      throw new Error('Invalid director config: policy shares add up to more than max_workers count');
+    }
+
+    _.set(director, 'policies.user.create.max_workers', userCreateWorkers);
+    _.set(director, 'policies.user.update.max_workers', userUpdateWorkers);
+    _.set(director, 'policies.user.delete.max_workers', userDeleteWorkers);
+    _.set(director, 'policies.scheduled.max_workers', autoWorkers);
```
jagadish-kb: why are these set required? you can as well retrieve the same as just `targetDirectorConfig.policies.user[action]`? seems dupe.

common/config.js
```diff
@@ -75,6 +75,27 @@ function completeDirectorConfig(director) {
...
   }
+  let boshRateLimitEnabled = config.enable_bosh_rate_limit;
+  if (boshRateLimitEnabled) {
+    const maxWorkers = _.get(director, 'max_workers', 6);
+    const userCreateWorkers = _.get(director, 'policies.user.create');
+    const userUpdateWorkers = _.get(director, 'policies.user.update');
+    const userDeleteWorkers = _.get(director, 'policies.user.delete');
+    const autoWorkers = _.get(director, 'policies.scheduled.max_workers', (maxWorkers / 2));
+
+    if (userCreateWorkers === undefined || userUpdateWorkers === undefined || userDeleteWorkers === undefined) {
```
jagadish-kb: can define defaults in bosh release so that this validation need not be done.

broker/lib/bosh/BoshDirectorClient.js
```diff
@@ -365,13 +429,21 @@ class BoshDirectorClient extends HttpClient {
         if (config === undefined) {
           throw new errors.NotFound('Did not find any bosh director config which supports creation of deployment');
         }
+        const reqHeaders = {
+          'Content-Type': 'text/yaml'
+        };
+        if (dbUpdate) {
+          reqHeaders[CONST.BOSH_RATE_LIMITS.BOSH_CONTEXT_ID] = CONST.BOSH_RATE_LIMITS.BOSH_FABRIK_OP_MONGO;
```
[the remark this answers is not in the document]
abh1kg: We decided to not apply any staggering for the MongoDB OOB updates. Hence, the special treatment

broker/lib/bosh/BoshOperationCache.js
```diff
@@ -0,0 +1,261 @@
...
+    credentials: {
+      rootCertificate: Buffer.from(config.etcd.ssl.ca, 'utf8'),
+      privateKey: Buffer.from(config.etcd.ssl.key, 'utf8'),
+      certChain: Buffer.from(config.etcd.ssl.crt, 'utf8')
+    }
+  });
+  return etcd;
+}
+
+function getKey(deploymentName) {
```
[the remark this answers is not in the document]
abh1kg: I did not want to expose the `getTaskKey` and `getKey` functions to the callers. These are also prefixed with a namespace which should ideally be impervious to the caller. 
I will rename the function to `getDeploymentKey`

broker/lib/bosh/BoshOperationCache.js
```diff
@@ -0,0 +1,261 @@
...
+    });
+    return wrapper;
+  }
+
+  /**
+   * Fetches all BOSH deployment names in the Etcd store currently
+   */
+  getDeploymentNames() {
+    logger.debug('Getting the current deployment cache...');
+    const wrapper = new Promise((resolve, reject) => {
```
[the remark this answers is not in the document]
abh1kg: The change from `new Promise` to `Promise.try` can be done but:
- The etcd3 library uses a separate native ES6-like Promise library of its own, which messes with the global Promise chains used. I did not want to complicate the promise hierarchy by using `promisify` or `try`. Of course, I can still use `try` to convert the output to a Bluebird promise but we may end up with some problems later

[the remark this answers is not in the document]
abh1kg: I would refactor the `getDeploymentNames` to the way suggested above. But, I would still like to keep the `getNEntries` function as a utility, which would delegate to the `getDeploymentNames` internally.

broker/lib/bosh/BoshOperationCache.js
```diff
@@ -0,0 +1,261 @@
...
+   * @param {Object} params - Parameters used for the operation
+   * @param {Object} args - Arguments used for the operation
+   */
+  store(planId, deploymentName, params, args) {
+    logger.info('Putting the current deployment operation into the cache', planId, deploymentName, params, args);
+    const wrapper = new Promise((resolve, reject) => {
+      const key = getKey(deploymentName);
+      const operation = new DeploymentOperation(planId, deploymentName, params, args);
+      this.containsDeployment(deploymentName).then(exists => {
+        if (exists) {
```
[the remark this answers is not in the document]
abh1kg: The operations that go through these checks are only for user-based ops. These ops which always create or update a BOSH deployment are treated asynchronously from the PoV of CC. Yes, they always return `202-ACCEPTED`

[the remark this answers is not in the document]
abh1kg: For the point on decoupling from CC, I would assume the proxy would still remain the same way since the concept of asynchronous provisioning/updates should not change. Of course, I would treat this for future refactoring if required

broker/lib/bosh/BoshOperationCache.js
```diff
@@ -0,0 +1,261 @@
...
+    this.params = params;
+    this.args = args;
+  }
+
+  toJson() {
+    return JSON.stringify(this);
+  }
+}
+
+class BoshOperationCache {
```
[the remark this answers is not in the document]
abh1kg: Will rename this and the refs to use `BoshOperationQueue`

broker/lib/fabrik/DirectorInstance.js
```diff
@@ -240,21 +243,60 @@ class DirectorInstance extends BaseInstance {
...
-      .catchThrow(NotFound, new ServiceInstanceNotFound(this.guid))
-      .then(task => {
-        assert.ok(_.endsWith(task.deployment, this.guid), `Deployment '${task.deployment}' must end with '${this.guid}'`);
-        this.networkSegmentIndex = this.manager.getNetworkSegmentIndex(task.deployment);
-        this.setOperationState(operation, task);
-        if (operation.state !== 'in progress') {
-          return this.finalize(operation);
+    const instanceId = this.guid;
+
+    const getBoshTaskOperation = (operation, taskId, removeFromCache) => {
```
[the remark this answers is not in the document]
abh1kg: Will do.

broker/lib/fabrik/DirectorInstance.js
```diff
@@ -240,21 +243,60 @@ class DirectorInstance extends BaseInstance {
...
+        .return(operation);
+    };
+
+    if (operation.task_id) {
+      return getBoshTaskOperation(operation, operation.task_id, false);
+    } else {
+      return Promise.try(() => {
+        return this.manager.getCurrentOperationState(this.guid);
+      }).then(state => {
+        const isCached = state.cached;
```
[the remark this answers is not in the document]
abh1kg: I actually thought of naming this as `queued` but named this `cached` instead to avoid a naming confusion with the bosh task's `queued` state. So, `cached` should indicate if the operation is staggered and available in etcd.

broker/lib/fabrik/DirectorInstance.js
```diff
@@ -240,21 +243,60 @@ class DirectorInstance extends BaseInstance {
...
+        const taskId = state.task_id;
+        if (isCached) {
+          return _.assign(operation, {
+            description: `${_.capitalize(operation.type)} deployment is still in progress`,
+            state: 'in progress'
+          });
+        } else {
+          return getBoshTaskOperation(operation, taskId, true);
```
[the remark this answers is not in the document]
abh1kg: There are three cases here:
- If the bosh rate limit feature is not enabled, only the task_id is available [reverts to current flow]
- If the bosh rate limit feature is enabled and there were empty slots in BOSH, only the task_id is available [reverts to current flow]
- If the bosh rate limit feature is enabled and there were no empty slots in BOSH, the operation is queued in Etcd and the director task poller runs every minute checking against BOSH
   - In between, the last operation request from CC could arrive at the broker. In this state, the broker checks against two etcd keys [mentioned below] and if it finds that a task_id has already been generated, it checks against BOSH for the status [current flow]. If it finds that a task_id has not been generated yet and the entry is still in ETCD, it returns back `in_progress`
   - As soon as a slot opens up in BOSH, the poller attempts the operation against BOSH and removes the initial entry from etcd. It now gets a task ID from BOSH and pushes it into another key in ETCD against the instance ID

So, if the task is not queued on BOSH yet, the task ID will not be available but in this case, the `getBoshTaskStatus` would be invoked and it would fail there anyway since an `assert` is already there

broker/lib/fabrik/DirectorManager.js
```diff
@@ -177,7 +189,179 @@ class DirectorManager extends BaseManager {
...
+      .then(directorConfig => {
+        targetDirectorConfig = directorConfig;
+        return this.director.getCurrentTasks(action, targetDirectorConfig);
+      })
+      .then(tasksCount => {
+        if (!policyApplicable) {
+          runOutput.shouldRunNow = true;
+          return Promise.resolve(runOutput);
+        }
+        if (dbUpdate) {
```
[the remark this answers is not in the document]
abh1kg: This was kept here to have some resilience against BOSH outage. If BOSH is down, the `tasks` API would fail and we queue in that case.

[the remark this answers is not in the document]
abh1kg: I will make this change. On the other hand, this would mean that the any operation with this `_runImmediately` flag set would not go through the `executePolicy` function at all. So, technically, MongoDB updates will not be resilient against BOSH outages. I guess that's ok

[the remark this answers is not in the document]
abh1kg: This was done to guard against BOSH failures

broker/lib/fabrik/DirectorManager.js
abh1kg: Yes

broker/lib/fabrik/DirectorManager.js
abh1kg: For user tasks, the exception is thrown later on once it's saved in ETCD. For scheduled tasks, we do not store in etcd because the scheduler will pick it up for processing later on - the exception is just thrown in the case of scheduled ops

broker/lib/fabrik/DirectorManager.js
abh1kg: No, this is for the case where the task poller runs, finds that slots are available in BOSH - then, the entry from etcd for the deployment should be removed and then processed. Note that this happens _before_ the task is submitted to BOSH

broker/lib/fabrik/DirectorTaskPoller.js
abh1kg: wanted to convey `error in scheduled deployment operation for ... `
I guess it came out wrong. Will change

broker/lib/fabrik/LockStatusPoller.js
abh1kg: Actually I want to reset the reference so that I forcibly can release the lock in the `stop` function below

common/config.js
abh1kg: Ok, I will make the change once the bosh release changes are done

broker/lib/fabrik/DirectorInstance.js
```diff
@@ -240,21 +243,60 @@ class DirectorInstance extends BaseInstance {
...
+          return this.manager.getTask(taskId);
+        })
+        .catchThrow(NotFound, new ServiceInstanceNotFound(instanceId))
+        .then(task => {
+          assert.ok(_.endsWith(task.deployment, this.guid), `Deployment '${task.deployment}' must end with '${this.guid}'`);
+          this.networkSegmentIndex = this.manager.getNetworkSegmentIndex(task.deployment);
+          this.setOperationState(operation, task);
+          if (operation.state !== 'in progress') {
+            return Promise.try(() => {
+                if (removeFromCache) {
```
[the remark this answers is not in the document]
abh1kg: @jagadish-kb : The actual caching logic + removal are in DirectorManager. However, even if we obscure by putting the output of `getCurrentOperationState` from DirectorInstance, the DirectorInstance would _still_ need to invoke the `finalizer`. Is that what you're looking for here?

abh1kg:
@jagadish-kb : The review comments should be addressed now. Can you please have a look?

broker/lib/bosh/BoshDirectorClient.js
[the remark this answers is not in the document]
jagadish-kb: ok.

broker/lib/fabrik/DirectorInstance.js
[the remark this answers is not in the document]
jagadish-kb: ok.

broker/lib/bosh/BoshDirectorClient.js
```diff
@@ -344,7 +344,71 @@ class BoshDirectorClient extends HttpClient {
...
+          'delete': taskGroup[CONST.OPERATION_TYPE.DELETE] ? taskGroup[CONST.OPERATION_TYPE.DELETE].length : 0,
+          'update': taskGroup[CONST.OPERATION_TYPE.UPDATE] ? taskGroup[CONST.OPERATION_TYPE.UPDATE].length : 0,
+          'scheduled': taskGroup[CONST.FABRIK_SCHEDULED_OPERATION] ? taskGroup[CONST.FABRIK_SCHEDULED_OPERATION].length : 0,
+          'uncategorized': taskGroup[CONST.UNCATEGORIZED] ? taskGroup[CONST.UNCATEGORIZED].length : 0,
+          'total': out.length
+        };
+      });
+  }
+
+  createOrUpdateDeployment(action, manifest, opts, scheduled, dbUpdate) {
```
jagadish-kb: this flag dbUpdate is now not being passed from invokers. Can you check and remove if not requiredl

broker/lib/bosh/BoshDirectorClient.js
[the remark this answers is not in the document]
jagadish-kb: now since you skip the policy check for runImmediately do you need this ?

broker/lib/fabrik/DirectorManager.js
```diff
@@ -177,7 +188,205 @@ class DirectorManager extends BaseManager {
...
+        if (allTasks >= maxTasks) {
+          //no slots left anyway
+          runOutput.shouldRunNow = false;
+          return runOutput;
+        }
+        if (scheduled) {
+          currentTasks = tasksCount.scheduled;
+          maxWorkers = targetDirectorConfig.policies.scheduled.max_workers;
+        } else {
+          currentTasks = tasksCount[action];
```
jagadish-kb: potential undefined. use lodash to avoid undefined runtime errors. `_.get(tasksCounts, action, 0)`

broker/lib/fabrik/DirectorManager.js
```diff
@@ -177,7 +188,205 @@ class DirectorManager extends BaseManager {
...
+          maxWorkers = targetDirectorConfig.policies.user[action].max_workers;
```
jagadish-kb: better to use lodash get with default to avoid undefined run time errors.

broker/lib/fabrik/DirectorManager.js
```diff
@@ -177,7 +188,205 @@ class DirectorManager extends BaseManager {
...
+      'cached': false
+    };
+    return this.executePolicy(scheduled, action, deploymentName)
+      .then(checkResults => this.enqueueOrTrigger(checkResults.shouldRunNow, scheduled, deploymentName))
+      .tap(res => {
+        decisionMaker.shouldRunNow = res.shouldRunNow;
+        decisionMaker.cached = res.cached;
+      })
+      .then(res => res.enqueue)
+      .then(enqueue => {
```
jagadish-kb: unnecessary promise chain here. can instead check `res.enqueue` right?

broker/lib/fabrik/DirectorManager.js
```diff
@@ -177,7 +188,205 @@ class DirectorManager extends BaseManager {
...
+              out.cached = true;
+              out.task_id = taskId;
+              return out;
+            });
+        } else {
+          out.task_id = taskId;
+          return out;
+        }
+      })
+      .catch(DeploymentDelayed, e => {
```
jagadish-kb: could have avoided this promise chain by doing something like this:

```
 if(enque){
   // Do what is required & return 'cached' true
}else {
 return Promise.try(()=> {
   if(shouldRunNow & Cached) {
      //delete cache entry
    }
    }).then(() => this._createOrUpdateDeployment(deploymentName, params, args, scheduled))
       ...
}

```
keeps the flow in a logical order. had this exception `DeploymentDelayed` been thrown from elsewhere as it would have made sense to have it this way, but this is similar to a `goto` stmt which could be avoided.

abh1kg: That's true but this leads to inner promise chains being elongated. Of course, I am not questioning this because the way to break out of promise chains early is a long-running discussion (https://github.com/petkaantonov/bluebird/issues/581). In fact, the creator of Bluebird prefers throwing an exception to break out [go figure! :) ]
I would keep this flow for now and refactor this in a latter PR. Hope this is fine

abh1kg:
@jagadish-kb : Changes as requested are done

broker/lib/fabrik/DirectorManager.js
jagadish-kb: ok

broker/lib/fabrik/DirectorManager.js
jagadish-kb: ok

PR #293, opened 2018-06-05
Remove broker process hard dependency on Docker

### LINKED ISSUE #280: Docker dependency on broker process
Docker:
 
If we call the info api without a docker deployed we always get “ready”: false:
 
curl https://sf.abapcp-staging.aws.sapcloud.io:9292/api/v1/info
 
{"name":"service-fabrik-broker","api_version":"1.0","ready":false,"db_status":"NOT_CONFIGURED"}
 
It seems that the broker still wants to [query docker](https://github.com/cloudfoundry-incubator/service-fabrik-broker/blob/2018.T04b/broker/lib/controllers/ServiceFabrikApiController.js#L170-L185) and returns false, because it can’t reach docker.
 
Also not setting docker config doesn’t seem to work, because in our current version (T02b) they are read in the [config](https://github.com/cloudfoundry-incubator/service-fabrik-broker/blob/2018.T02b/lib/config.js#L74-L105).  And in the latest version they are still used in a file, which is required by [ServiceFabrikApiController](https://github.com/cloudfoundry-incubator/service-fabrik-broker/blob/2018.T04b/broker/lib/controllers/ServiceFabrikApiController.js#L7).

amitmalav:
Fixes #280 
`getInfo` api now returns
`{"name":"service-fabrik-broker","api_version":"1.0","ready":true,"db_status":<db-status>}`
if `enable_swarm_manager` property is set to false

### DIFF
--- a/broker/lib/controllers/ServiceFabrikApiController.js
+++ b/broker/lib/controllers/ServiceFabrikApiController.js
@@ -167,20 +167,18 @@ class ServiceFabrikApiController extends FabrikBaseController {
   }
 
   getInfo(req, res) {
-    if (!config.enable_swarm_manager) {
-      return res.status(CONST.HTTP_STATUS_CODE.OK)
-        .json({
-          name: this.serviceBrokerName,
-          api_version: this.constructor.version,
-          ready: true,
-          db_status: this.fabrik.dbManager.getState().status
-        });
-    }
-    let allDockerImagesRetrieved = false;
-    return docker
-      .getMissingImages()
-      .then(missingImages => allDockerImagesRetrieved = _.isEmpty(missingImages))
-      .catch(err => logger.info('error occurred while fetching docker images', err))
+    let allDockerImagesRetrieved = true;
+    return Promise.try(() => {
+        if (config.enable_swarm_manager) {
+          return docker
+            .getMissingImages()
+            .then(missingImages => allDockerImagesRetrieved = _.isEmpty(missingImages));
+        }
+      })
+      .catch(err => {
+        allDockerImagesRetrieved = false;
+        logger.info('error occurred while fetching docker images', err);
+      })
       .finally(() => {
         res.status(CONST.HTTP_STATUS_CODE.OK)
           .json({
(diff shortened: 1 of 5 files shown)


broker/lib/controllers/ServiceFabrikApiController.js
```diff
@@ -167,6 +167,15 @@ class ServiceFabrikApiController extends FabrikBaseController {
   }
 
   getInfo(req, res) {
+    if (!config.enable_swarm_manager) {
+      return res.status(CONST.HTTP_STATUS_CODE.OK)
```
sauravmndl: Would suggest not have code duplication. Better initialise 
allDockerImagesRetrieved = true 
and execute following code 
```
return docker
       .getMissingImages()
       .then(missingImages => allDockerImagesRetrieved = _.isEmpty(missingImages))
```
only if(config.enable_swarm_manager)==> true or not undefined. Then subsequent code remains same.

amitmalav: My idea was to avoid making call to docker client => `getMissingImages()` if swarm is not available.

sauravmndl: to avoid if 
```
if (!config.enable_swarm_manager)
```
right ? So, I was suggesting the same.

kamath-prasad: Yes agree with Saurav . I think the value of "ready" is a bit misleading , in the sense it should be based on multiple other things in the future. 

So for eg: 
ready = allDockerImagesRetrieved && allBOSHRealsesUploaded ... etc. 

Hence it will be fine to initialise allDockerImagesRetrieved to true and avoid code duplication and just have 1 return statement.

ashishjain14: @sauravmndl  what if docker.getMissingImages() errors out would not this result in sending the status as true even when there are some images to be downloaded.

ashishjain14: so in the catch we may have to reset the status

PR #297, opened 2018-06-12
Fix/event backup complete

### LINKED ISSUE #295: Backup complete status event not logged
When backup is completed, the event is not sent to event server (e.g. Riemann), not logged into database. Backup report is generated based on the event of backup completion. If the event not found the backup is declared as not triggered even though status was completed and successful.
This happens as status polling code is moved to scheduler process instead of broker process. Currently in scheduler process there is no listener is subscribed to any event, even though the event is published. 
Happening in following build:
service fabrik: https://github.com/cloudfoundry-incubator/service-fabrik-broker/archive/2018.T05b.tar.gz

sauravmndl:
This fixes #295
Fix is to initialise all the event listener in service fabrik scheduler worker.

### DIFF
--- a/broker/JobScheduler.js
+++ b/broker/JobScheduler.js
@@ -9,6 +9,7 @@ const cpus = require('os').cpus();
 const CONST = require('./lib/constants');
 const logger = require('./lib/logger');
 const config = require('./lib/config');
+const utils = require('./lib/utils');
 const errors = require('./lib/errors');
 const maintenanceManager = require('./lib/maintenance').maintenanceManager;
 const serviceFabrikClient = require('./lib/cf').serviceFabrikClient;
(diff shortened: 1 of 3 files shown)


broker/JobScheduler.js
```diff
@@ -74,6 +75,8 @@ class JobScheduler {
     process.on('message', (msg) => this.handleMessage(msg));
     logger.info(`Starting Service Fabrik Batch Job worker: ${cluster.worker.id} - ${process.pid}  @${new Date()}`);
     require('./lib/jobs');
+    logger.info('Initializing event listeners..');
```
jagadish-kb: Can you please remove this logger? or make it debug? in the long run we see too many unnecessary log statements in our app.

PR #300, opened 2018-06-15
Enable service fabrik integration with credhub

jagadish-kb:
IAAS credentials need not be stored as part of the service. Supports credhub integration.
Passes credhub co-ordinates during service deployment.

### DIFF
--- a/broker/config/settings.yml
+++ b/broker/config/settings.yml
@@ -402,6 +402,17 @@ defaults: &defaults
     url: https://api.bosh-lite.com
     username: admin
     password: admin
+  
+  #############################
+  # CREDHUB SETTINGS #
+  #############################
+  cred_provider:
+    credhub_url: https://10.0.3.11:8443/api/
+    credhub_uaa_url: https://10.0.3.11:8443
+    credhub_client_id:  credhub-admin
+    credhub_client_secret: 
+    credhub_username: admin
+    credhub_user_password: admin
 
   ###################
   # DOCKER SETTINGS #
--- a/broker/config/settings.yml
+++ b/broker/config/settings.yml
@@ -491,6 +491,7 @@ defaults: &defaults
       provider: &provider_blueprint
         name: boshlite
         container: blueprint_backup
+        credhub_key: '/service-fabrik/iaas/blueprint'
         credhub_username: blueprint-admin
         credhub_user_password: blueprint-secret
     rabbitmq: &agent_rabbitmq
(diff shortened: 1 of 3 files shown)


broker/config/settings.yml
```diff
@@ -402,6 +402,17 @@ defaults: &defaults
...
     username: admin
     password: admin
+  
+  #############################
+  # CREDHUB SETTINGS #
+  #############################
+  cred_provider:
+    credhub_url: https://10.0.3.11:8443/api/
+    credhub_uaa_url: https://10.0.3.11:8443
+    credhub_client_id:  credhub-admin
```
ashishjain14: Shouldn't the UAA related configs be rename as uaa_ instead of credhub_

broker/config/settings.yml
```diff
@@ -402,6 +402,17 @@ defaults: &defaults
...
+    credhub_client_secret:
```
ashishjain14: Does this mean there will be a single uaa_client_id/secret for all the services on-boarded? What if each service has its own dedicated clientid/secret in-line with what we have in place for uaa username/password for token retrieval?

ketakig12: Same question.

broker/lib/models/Service.js
```diff
@@ -2,14 +2,23 @@
...
       .assign({
-        plans: _.map(options.plans, plan => new Plan(this, plan))
+        plans: _.map(options.plans, plan => {
+          if (plan.manager.name === CONST.INSTANCE_TYPE.DIRECTOR && config.cred_provider &&
+            _.get(plan, 'manager.settings.context.agent.provider.credhub_key', undefined) !== undefined) {
+            //Inject credhub config into agent properties in case credhub is configured for the service
+            _.assign(_.get(plan, 'manager.settings.context.agent.provider'),
+              _.omit(config.cred_provider, 'credhub_username', 'credhub_user_password'));
+          }
+          return new Plan(this, plan);
```
ketakig12: Can you please add unit tests for this?

broker/config/settings.yml
jagadish-kb: Answered in BnR [PR](https://github.com/cloudfoundry-incubator/service-fabrik-backup-restore/pull/35#discussion_r196321935)

broker/config/settings.yml
jagadish-kb: No. Each service has its own credential. This is master credentials that SF uses and it can fetch any creds. Individual services can just fetch their creds. Please review sf-produc-component PR to know about individual service users/creds creation.

broker/lib/models/Service.js
jagadish-kb: Done!

PR #307, opened 2018-06-24
Fix for BOSH deployments not being picked up from etcd queue

abh1kg:
- Rely on bluebird promise instead of native promise
- Use etcd json properly

### DIFF
--- a/broker/lib/fabrik/DirectorTaskPoller.js
+++ b/broker/lib/fabrik/DirectorTaskPoller.js
@@ -10,6 +10,7 @@ const config = require('../config');
 const boshCache = bosh.BoshOperationQueue;
 const TIME_POLL = 1 * 60 * 1000;
 const LockStatusPoller = require('./LockStatusPoller');
+const Promise = require('bluebird');
 
 class DirectorTaskPoller extends LockStatusPoller {
   constructor() {
--- a/broker/lib/fabrik/DirectorTaskPoller.js
+++ b/broker/lib/fabrik/DirectorTaskPoller.js
@@ -26,9 +26,7 @@ class DirectorTaskPoller extends LockStatusPoller {
           })
           .then(cached => {
             let catalogPlan = catalog.getPlan(cached.plan_id);
-            return Promise.try(() => {
-                return DirectorManager.load(catalogPlan);
-              })
+            return DirectorManager.load(catalogPlan)
               .then(manager => manager.createOrUpdateDeployment(deploymentName, cached.params, cached.args));
           })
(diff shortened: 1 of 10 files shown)


abh1kg:
@jagadish-kb , @ketakig12 : Can you have a look and merge if you don't see any concerns? This was found out as an issue after running a load test on BOSH with the tasks being exceeded

broker/lib/fabrik/DirectorTaskPoller.js
```diff
@@ -25,7 +26,10 @@ class DirectorTaskPoller extends LockStatusPoller {
           })
           .then(cached => {
             let catalogPlan = catalog.getPlan(cached.plan_id);
-            return DirectorManager.load(catalogPlan).createOrUpdateDeployment(deploymentName, cached.params, cached.args);
+            return Promise.try(() => {
+                return DirectorManager.load(catalogPlan);
+              })
+              .then(manager => manager.createOrUpdateDeployment(deploymentName, cached.params, cached.args));
```
jagadish-kb: can change the above (LN29-33) to the below 
```
return  DirectorManager.load(catalogPlan)
              .then(manager => manager.createOrUpdateDeployment(deploymentName, cached.params, cached.args));
```

broker/lib/jobs/ServiceInstanceUpdateJob.js
```diff
@@ -97,7 +97,13 @@ class ServiceInstanceUpdateJob extends BaseJob {
...
           .catch(err => {
             operationResponse.update_init = CONST.OPERATION.FAILED;
-            logger.error('Error occurred while updating service insance job :', err);
+            logger.error('Error occurred while updating service instance job :', err);
+            if (err instanceof errors.DeploymentAttemptRejected) {
+              //If deployment was staggered due to exhaustion of workers, reschedule update job
+              //Retry attempts do not count when deployment is staggered
+              trackAttempts = false;
+              err.statusMessage = 'Deployment attempt rejected due to BOSH overload. Update cannot be initiated';
+            }
```
jagadish-kb: The update retry is unbounded here but it would be a good idea to have a timeout based on the firstAttempt started at. And if the update run is after the next scheduled update run (ie. after 7 days), then dont have to retry because then you would have two competing runs on the same deployment and it might simply vie for the queue.

For now you might not need to really fix this as it might simply make the code a bit complex here. Just document a TODO and after extensive tests in canary if we need to fix this, lets patch it.

ketakig12:
Merging the PR as we have 2 approvals.

PR #309, opened 2018-07-02
Feature/fine grained sec groups

jayeshjk:
* Requirement is to enable the creation of security groups for specific ports at the time of service instance creation. Currently, application security groups are created for entire port range of 1024-65k, exposing service instance components to attack initiated from inside the bound apps.

* Modified the flow of building rules for security group creation to consume relevant exposed port information from services catalog.

* Tested on dev landscape for blueprint service and will run through the dev pipeline once catalogs of other services are also updated.

* This change does not take care of update of application security groups of existing instances.

### DIFF
--- a/broker/lib/fabrik/CfPlatformManager.js
+++ b/broker/lib/fabrik/CfPlatformManager.js
@@ -116,11 +116,11 @@ class CfPlatformManager extends BasePlatformManager {
 
   buildSecurityGroupRules(options) {
     let portRule = '1024-65535';
-    if (options.port.exposedPorts) {
-      if (options.port.range) {
-        portRule = `${_.first(options.port.exposedPorts)}-${_.last(options.port.exposedPorts)}`;
+    if (options.securityGroupParams && options.securityGroupParams.exposed_ports) {
+      if (options.securityGroupParams.range) {
+        portRule = `${_.first(options.securityGroupParams.exposed_ports)}-${_.last(options.securityGroupParams.exposed_ports)}`;
       } else {
-        portRule = _.size(options.port.exposedPorts) === 1 ? `${_.first(options.port.exposedPorts)}` : `${_.join(options.port.exposedPorts,',')}`;
+        portRule = _.size(options.securityGroupParams.exposed_ports) === 1 ? `${_.first(options.securityGroupParams.exposed_ports)}` : `${_.join(options.securityGroupParams.exposed_ports,',')}`;
       }
     }
     return {
--- a/broker/lib/fabrik/CfPlatformManager.js
+++ b/broker/lib/fabrik/CfPlatformManager.js
@@ -114,17 +114,10 @@ class CfPlatformManager extends BasePlatformManager {
       );
   }
 
-  /*
-  {
-    protocol: 'tcp',
-    ips: ['10.11.20.248', '10.11.20.255'],
-    applicationAccessPorts: ['8080']
-  }
-  */
   buildSecurityGroupRules(options) {
     let portRule = '1024-65535';
-    if (options.applicationAccessPorts && _.size(options.applicationAccessPorts) > 0) {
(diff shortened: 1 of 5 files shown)


broker/lib/fabrik/CfPlatformManager.js
```diff
@@ -114,11 +114,22 @@ class CfPlatformManager extends BasePlatformManager {
...
+  /*
+  {
+    protocol: 'tcp',
+    ips: ['10.11.20.248', '10.11.20.255'],
+    applicationAccessPorts: ['8080']
+  }
+  */
   buildSecurityGroupRules(options) {
+    let portRule = '1024-65535';
+    if (options.applicationAccessPorts && _.size(options.applicationAccessPorts) > 0) {
```
abh1kg: Shouldn't the check be `Array.isArray(options.applicationAccessPorts) && _.size(options.applicationAccessPorts) > 0`?

jayeshjk: Yes. Will update this check. A related query: Should we handle a case where options.applicationAccessPorts is not an array? Ideally this should not happen, but with this check (and even with earlier one) default portRule will be used, if options.applicationAccessPorts is not an array.

jayeshjk:
@kamath-prasad @abh1kg  I think modifying update instance flow inside broker to incorporate this will involve extra calls to CC and overall logic will also get messy. So maybe, if needed, we can do this update as a one time thing?

broker/lib/fabrik/CfPlatformManager.js
```diff
@@ -114,11 +114,22 @@ class CfPlatformManager extends BasePlatformManager {
       );
   }
 
+  /*
```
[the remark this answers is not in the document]
jayeshjk: Will do.

kamath-prasad:
@abh1kg / @jayeshjk : Ok , then let us avoid update of existing rules.

broker/lib/fabrik/CfPlatformManager.js
abh1kg: I will like to not complicate this much. The services must put in an array with one entry in this case. Let's not provide another option of just a single entry for the access ports

broker/lib/fabrik/CfPlatformManager.js
```diff
@@ -114,11 +114,22 @@ class CfPlatformManager extends BasePlatformManager {
...
+  {
+    protocol: 'tcp',
+    ips: ['10.11.20.248', '10.11.20.255'],
+    applicationAccessPorts: ['8080']
+  }
+  */
   buildSecurityGroupRules(options) {
+    let portRule = '1024-65535';
+    if (options.applicationAccessPorts && _.size(options.applicationAccessPorts) > 0) {
+      portRule = _.size(options.applicationAccessPorts) === 1 ? `${_.first(options.applicationAccessPorts)}` : `${_.join(options.applicationAccessPorts,',')}`;
```
sauravmndl: `_.join(options.applicationAccessPorts,',')` is enough. Even if array size is 1 produces same result.

jayeshjk: Okay. Will update it as well. Thanks

broker/lib/fabrik/CfPlatformManager.js
```diff
@@ -115,10 +115,14 @@ class CfPlatformManager extends BasePlatformManager {
   }
 
   buildSecurityGroupRules(options) {
+    let portRule = '1024-65535';
+    if (Array.isArray(options.applicationAccessPorts) && _.size(options.applicationAccessPorts) > 0) {
+      portRule = `${_.join(options.applicationAccessPorts,',')}`;
```
abh1kg: Need not have the "``" here.... Just `_.join(options.applicationAccessPorts, ',')` suffices

kamath-prasad:
@jayeshjk  : please check if you need to rebase your branch as the code coverage looks reduced now.

PR #311, opened 2018-07-03
Fix/default bm

amitmalav:
Fix method name in DefaultBackupManager
Fix some comments `// TODO - PR` => `//TODO-PR`


kamath-prasad:
@amitmalav  : coverage reduced ? :)

subhankarc:
@kamath-prasad We would have to write some unit tests for the DefaultBackupManager file.

There is one more fix coming. Hence I added do not merge.

PR #321, opened 2018-07-06
Fix/refactor

muralisc:
- Optimise Apiserver calls - Use single call to update state and response
- Add BaseDirectorService

### DIFF
--- a/managers/BaseDirectorService.js
+++ b/managers/BaseDirectorService.js
@@ -4,6 +4,14 @@ const _ = require('lodash');
 const utils = require('../broker/lib/utils');
 
 class BaseDirectorService {
+  constructor(plan) {
+    this.plan = plan;
+  }
+
+  get settings() {
+    return this.plan.manager.settings;
+  }
+
   static parseDeploymentName(deploymentName, subnet) {
     return _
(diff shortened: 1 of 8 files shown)


managers/BaseDirectorService.js
```diff
@@ -0,0 +1,24 @@
...
+class BaseDirectorService {
+  constructor(plan) {
+    this.plan = plan;
+  }
+
+  get settings() {
+    return this.plan.manager.settings;
+  }
+
+  static parseDeploymentName(deploymentName, subnet) {
```
kamath-prasad: duplicate [code](https://github.com/cloudfoundry-incubator/service-fabrik-broker/blob/master/managers/backup-manager/BackupService.js#L46-L53)

subhankarc: This method was removed from BackupService and moved to BaseDirectorService. @muralisc correct?

muralisc: @kamath-prasad its removed from [here](https://github.com/cloudfoundry-incubator/service-fabrik-broker/pull/321/files#diff-b35054ba7edff4ba9736b79267cb63a2L47)

PR #324, opened 2018-07-09
Add Base Manager

amitmalav:
Adding a `BaseManager` class that can be extended by various managers.
Member functions are:
 - `registerWatcher`
 - `_preProcessRequest`
 - `processRequest`
 - `_postProcessRequest`

Currently `DefaultBackupManager` extends `BaseManager` class

Any new Manager can just extend the BaseManager and only have to implement the processRequest part.


managers/BaseManager.js
```diff
@@ -7,54 +7,134 @@ const config = require('../common/config');
...
 class BaseManager {
 
-  registerWatcher() {
-    throw new NotImplementedBySubclass('registerWatcher');
+  registerWatcher(resourceGroup, resourceType, queryString) {
+    logger.debug(`Registering watcher for resourceGroup ${resourceGroup} of type ${resourceType} with queryString as ${queryString}`);
+    return eventmesh.apiServerClient.registerWatcher(resourceGroup, resourceType, this.handleResource, queryString)
+      .then(stream => {
+        logger.info(`Successfully set watcher with query string:`, queryString);
+        return setTimeout(() => {
```
abh1kg: I would prefer if you use `Promise.delay` instead of `setTimeout` here so that we don't mix promises and global timers.

subhankarc:
@amitmalav Can you please add some unit tests for the BaseManager and DefaultBackupManager?

managers/BaseManager.js
```diff
@@ -7,54 +7,134 @@ const config = require('../common/config');
...
+          try {
+            logger.info(`Refreshing stream after ${CONST.APISERVER.WATCHER_REFRESH_INTERVAL}`);
+            stream.abort();
+            return this.registerWatcher();
```
abh1kg: If you use `Promise.delay`, synchronously handling errors here is not required and you can use `Bluebird catch` for any internal errors

managers/BaseManager.js
```diff
@@ -7,54 +7,134 @@ const config = require('../common/config');
...
+      .tap((resource) => logger.info(`Successfully released processing lock for the request with options: ${JSON.stringify(changedOptions)}\n` +
         `Updated resource with annotations is: `, resource));
   }
 
+  _preProcessRequest(objectBody, processingLockStatus) {
+    const options = JSON.parse(objectBody.spec.options);
+    // Acquire processing lock so that in HA scenerio, only one backup-manager process processes the request
+    return Promise.try(() => {
+      if (!objectBody.metadata.annotations || objectBody.metadata.annotations.lockedByManager === '') {
+        return BaseManager.prototype._acquireProcessingLock(objectBody)
```
abh1kg: I understand the need to use the `prototype` directly here since you do not have an object to work with but can this be done in a better way?

managers/BaseManager.js
```diff
@@ -7,54 +7,134 @@ const config = require('../common/config');
...
+      } else {
+        processingLockStatus.conflict = true;
+        logger.info(`Request with options ${JSON.stringify(options)} is picked by other process with ip ${objectBody.metadata.annotations.lockedByManager}`);
+      }
+    });
+  }
+
+  _postProcessRequest(objectBody) {
+    const options = JSON.parse(objectBody.spec.options);
+    return BaseManager.prototype._releaseProcessingLock(objectBody)
```
abh1kg: Same comment as above

managers/BaseManager.js
```diff
@@ -7,54 +7,134 @@ const config = require('../common/config');
...
+          }
+        })
+        .catch(err => {
+          if (!processingLockStatus.conflict) {
+            logger.error(`Caught error while processing request with options ${JSON.stringify(changedOptions)}`, err);
+          }
+        })
+        .finally(() => {
+          if (!processingLockStatus.conflict) {
+            return BaseManager.prototype._postProcessRequest(changeObjectBody);
```
abh1kg: Same comment as above

managers/BaseManager.js
```diff
@@ -7,54 +7,134 @@ const config = require('../common/config');
...
-  worker() {
-    throw new NotImplementedBySubclass('registerWatcher');
-  }
-
-
   /**
    * @description Patches resource with annotation key lockedByManager and value broker ip
    */
-  static acquireProcessingLock(changeObjectBody) {
+  _acquireProcessingLock(changeObjectBody) {
```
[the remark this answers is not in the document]
amitmalav: Yes, we have added few internal methods in some classes which are not called by objects directly.

managers/BaseManager.js
amitmalav: If my understanding is correct it will mean that call will never get resolved.

abh1kg: You can get around this by using a catch in the promise chain. Basic code flow as below:
```
Promise.delay(CONST.APISERVER.WATCHER_REFRESH_INTERVAL)
  .then(()=>{
      logger.info(`Refreshing stream after ${CONST.APISERVER.WATCHER_REFRESH_INTERVAL}`); 
      stream.abort();
      return this.registerWatcher();
    })
    .catch(err=> logger.error('Failed registering watcher with error', e)
```

managers/BaseManager.js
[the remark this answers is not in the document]
subhankarc: This is something we have not used earlier that extensively, but recently started adopting it, thanks to @abh1kg  :)

I like this approach as you get an understanding of internal methods and exposed methods.

managers/BaseManager.js
subhankarc: One thing to ponder on is that we would like to co-locate similar Managers into a single BOSH jobs. For example, Backup manager and Restore Manager together will go to a job. In that case, there would be single file which calls both the managers, or instantiate both. Will that be possible if the registerwatcher is a recursive call?

managers/backup-manager/DefaultBackupManager.js
```diff
@@ -3,19 +3,33 @@
...
 const DBManager = require('../../broker/lib/fabrik/DBManager');
-const errors = require('../../common/errors');
-const Conflict = errors.Conflict;
 
 /* jshint nonew:false */
 new DBManager(); //to start the BnRStatusPoller
 
 class DefaultBackupManager extends BaseManager {
+  constructor() {
+    super();
```
subhankarc: Even though it serves the same purpose, having the registerwatcher in the constructor doesnt seem right. How about putting it in init() and calling init() explicitly by the caller?

subhankarc: Also, as I mentioned earlier, we should have the provision to run multiple manager under one node process.

kamath-prasad:
@amitmalav : can you please add description to this PR ? Difficult to understand what all things are achieved after this change.

subhankarc:
Merge it along with https://github.com/cloudfoundry-incubator/service-fabrik-boshrelease/pull/100

PR #326, opened 2018-07-12
Fix/remove sfv2 flag

muralisc:
This PR contains the following
 - removes the service-fabrik V2 flag
 - Changes in the getLastBackup which no longer fetches the status from backup store and instead get it from the backup resource of apiserver
 - BnRStatusPoller updates the log or error details in completion of the backup.


subhankarc:
@muralisc we need to change the sfpc also.

broker/lib/controllers/ServiceBrokerApiController.js
```diff
@@ -85,8 +85,6 @@ class ServiceBrokerApiController extends FabrikBaseController {
       if (req.instance.async) {
```
subhankarc: Now there will be no sync operation. We can as well remove the if.

PR #328, opened 2018-07-13
Enhance the broker agent bind response to include network details

jagadish-kb:
Bind response to contain details of the host and port as part of the endpoints attribute which makes it standard for all consumers and they need not parse the URI to figure out the same.


broker/config/route-registrar.json
```diff
@@ -1,20 +1,16 @@
...
-      "port": 9292,
-      "registration_interval": "20s",
-      "uris": [
-        "service-fabrik-broker.bosh-lite.com"
-      ]
-    }
-  ]
-}
+  "message_bus_servers": [{
+    "host": "10.244.0.129:4222",
```
jagadish-kb: There was no need to checkin this file, but with cf deployment this is the default ip of nats server. Else everyone has to update it every time.

broker/config/route-registrar.json
```diff
@@ -1,20 +1,16 @@
...
-      "name": "fabrik",
-      "port": 9292,
-      "registration_interval": "20s",
-      "uris": [
-        "service-fabrik-broker.bosh-lite.com"
-      ]
-    }
-  ]
-}
+  "message_bus_servers": [{
```
subhankarc: +1 for this check in

jagadish-kb:
@subhankarc  For now the network Id is fixed since services are only deployed on CF by SF. This is a constant thats given to us by NM Team. Later we will move this to settings yml which will be an IAC import from NM component.

PR #329, opened 2018-07-13
Package Refactoring part1

subhankarc:
The following package level refactoring is done as part of this.
 - Moved controllers and routes
 - Moved BOSH to data-access-layer folder
 - Moved iaas to data-access-layer

The final package structure is planned like the following

service-fabrik-broker
 - loggers
   - event loggers
 - managers
 - api-controllers
 - data-access-layer
   - cf
   - bosh
   - docker
   - db
   - iaas
   - eventmesh
 - jobs
 - ui
 - reports
 - quotas
 - common
   - constants
   - utils
   - models
   - config
   - middleware


sauravmndl:
Instead of naming the directory **api-controllers** we could have kept only controllers. Within that we can  create sub-directories as per need. Just a thought.

broker/lib/fabrik/VirtualHostRelationMapper.js
```diff
@@ -2,9 +2,9 @@
 
 const Promise = require('bluebird');
 const logger = require('../logger');
-const bosh = require('../bosh');
+const bosh = require('../../../data-access-layer/bosh');
```
sauravmndl: Can we directly name the sub package directory **data-access-layer** as **dal**?

api-controllers/ServiceFabrikAdminController.js
```diff
@@ -3,20 +3,20 @@
...
 const FabrikBaseController = require('./FabrikBaseController');
-const utils = require('../utils');
-const fabrik = require('../fabrik');
-const bosh = require('../bosh');
-const ScheduleManager = require('../jobs');
-const BackupReportManager = require('../reports');
-const CONST = require('../constants');
-const maintenanceManager = require('../maintenance').maintenanceManager;
+const utils = require('../broker/lib/utils');
+const fabrik = require('../broker/lib/fabrik');
```
sauravmndl: Some of the parts like **broker/lib/fabrik** can be renamed to **service/fabrik**. Or some other sub directories also. 
So, we can look package structure as **controllers, service, dal**.

broker/lib/fabrik/VirtualHostRelationMapper.js
subhankarc: It does not sound obvious, h nice the full name. It’s not that big hence I thought we can use it.

api-controllers/ServiceFabrikAdminController.js
subhankarc: This is only first part of the restructuring. Avoided only controllers as it also clashes with Kubernetes controllers. api-controllers is more explicit I guess.

subhankarc: The Fabrik folder would slowly go away as all the codes would move to managers.

We would remove the broker folder and have everything either directly under root or under different folders. I’ll give a final folder structure snippet that we agreed on in the last meeting in the description.

amitmalav:
Raising new PR with more changes.
https://github.com/cloudfoundry-incubator/service-fabrik-broker/pull/330

PR #331, opened 2018-07-18
Transition sfv2

muralisc:
Handle exiting backups when transitioning to sfv2.
- existing backup metadata is not available in apiserver
- check object store if the metadata is not available in apiserver
- For operations last backup and delete backup

### DIFF
--- a/api-controllers/ServiceFabrikApiController.js
+++ b/api-controllers/ServiceFabrikApiController.js
@@ -428,6 +428,8 @@ class ServiceFabrikApiController extends FabrikBaseController {
         })
       )
       .catch(NotFound, (err) => {
+        // This code block is specifically for the transition of Service Fabrik to v2
+        // Here we reffer to BackupService to get the lastBackup status
         logger.info('Backup metadata not found in apiserver, checking blobstore. Error message:', err.message);
         const tenantId = _.get(req, 'body.space_guid') ||
           _.get(req, 'query.space_guid') ||
@@ -671,15 +673,9 @@ class ServiceFabrikApiController extends FabrikBaseController {
         value: options
       })
       .catch(NotFound, (err) => {
+        // if not found in apiserver delete from blobstore
         logger.info('Backup metadata not found in apiserver, checking blobstore. Error message:', err.message);
-        const tenantId = _.get(req, 'body.space_guid') ||
-          _.get(req, 'query.space_guid') ||
-          _.get(req, 'query.tenant_id') ||
-          _.get(req, 'body.context.space_guid') ||
-          _.get(req, 'body.context.namespace');
-        return this.getPlanIdFromInstanceId(req.params.instance_id)
-          .then(plan_id => BackupService.createService(catalog.getPlan(plan_id)))
-          .then(backupService => backupService.deleteBackup(tenantId, req.params.instance_id));
+        return this.backupStore.deleteBackupFile(options);
       })
       .then(() =>
         eventmesh.apiServerClient.updateOperationState({
--- a/api-controllers/ServiceFabrikApiController.js
+++ b/api-controllers/ServiceFabrikApiController.js
@@ -15,16 +15,13 @@ const FabrikBaseController = require('./FabrikBaseController');
 const Unauthorized = errors.Unauthorized;
 const NotFound = errors.NotFound;
 const Gone = errors.Gone;
-const Timeout = errors.Timeout;
 const cf = require('../data-access-layer/cf');
 const Forbidden = errors.Forbidden;
 const BadRequest = errors.BadRequest;
-const Conflict = errors.Conflict;
 const UnprocessableEntity = errors.UnprocessableEntity;
 const ServiceInstanceNotFound = errors.ServiceInstanceNotFound;
 const JsonWebTokenError = jwt.JsonWebTokenError;
 const ContinueWithNext = errors.ContinueWithNext;
-const InternalServerError = errors.InternalServerError;
 const DeploymentAlreadyLocked = errors.DeploymentAlreadyLocked;
 const ScheduleManager = require('../jobs');
 const config = require('../common/config');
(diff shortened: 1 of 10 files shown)


api-controllers/ServiceFabrikApiController.js
```diff
@@ -433,33 +369,39 @@ class ServiceFabrikApiController extends FabrikBaseController {
...
+        resourceId: req.params.instance_id,
         operationName: CONST.OPERATION_TYPE.BACKUP,
         operationType: CONST.APISERVER.RESOURCE_TYPES.DEFAULT_BACKUP,
-        operationId: backupGuid,
-      }).then(state => {
-        // abort only if the state is in progress
-        if (state === CONST.APISERVER.RESOURCE_STATE.IN_PROGRESS) {
-          return eventmesh.apiServerClient.updateOperationState({
+      })
+      .then(backupGuid => {
```
jagadish-kb: The method invoked method is `getLastOperation` and its response is expected to be a `backupguid`? rather it would be better if it returned an `operation` object something like the below, which is generic:
```
{
   operation: backup
   operation_params: {
        backup_guid: xxxx
   }
}
```
In the above you can skip `operation_parms` and directly put those op specific params one level up.  
Or if the intention is to always return backup_guid then name the method accordingly.

api-controllers/ServiceFabrikApiController.js
```diff
@@ -433,33 +369,39 @@ class ServiceFabrikApiController extends FabrikBaseController {
...
-          logger.info(`Skipping abort as state is : ${state}`);
-        }
-      }).then(() => ServiceFabrikApiController.getResourceOperationStatus({
-        operationId: backupGuid,
-        start_state: CONST.OPERATION.ABORT,
-        started_at: backupStartedAt
-      }));
-    }).then(status => res.status(status.state === 'aborting' ? CONST.HTTP_STATUS_CODE.ACCEPTED : CONST.HTTP_STATUS_CODE.OK).send({}));
+          })
+          .then(state => {
```
jagadish-kb: why cant the methods apiServerClient.getLastOperation & apiServerClient.getOperationState be merged into one as a generic method like `getLastOperationState`? Just a thought, i think this might be useful at many other places in many other controllers/pollers as well later down the line.

api-controllers/ServiceFabrikApiController.js
```diff
@@ -433,33 +369,39 @@ class ServiceFabrikApiController extends FabrikBaseController {
...
+                operationName: CONST.OPERATION_TYPE.BACKUP,
+                operationType: CONST.APISERVER.RESOURCE_TYPES.DEFAULT_BACKUP,
+                operationId: backupGuid,
+                stateValue: CONST.OPERATION.ABORT
+              });
+            } else {
+              logger.info(`Skipping abort as state is : ${state}`);
+            }
+          })
+          .then(() => eventmesh.apiServerClient.getResourceOperationStatus({
```
jagadish-kb: here dont you have to poll wait for the status to change from `ABORT` to `ABORTING` like you do at other places where you wait for `in-progress`?

api-controllers/ServiceFabrikApiController.js
```diff
@@ -639,13 +581,18 @@ class ServiceFabrikApiController extends FabrikBaseController {
...
-    return eventmesh.apiServerClient.patchOperationOptions({
+    return eventmesh
+      .apiServerClient.patchOperationOptions({
         operationName: CONST.OPERATION_TYPE.BACKUP,
         operationType: CONST.APISERVER.RESOURCE_TYPES.DEFAULT_BACKUP,
         operationId: req.params.backup_guid,
         value: options
       })
-      .catchThrow(NotFound, new Gone('Backup does not exist or has already been deleted'))
+      .catch(NotFound, (err) => {
```
jagadish-kb: For restore you dont have to handle restore from object stores? I mean the same logic of getBackup should be applied even for restores right durign the migration period? Why is the file just deleted if no backup info found in etcd?

common/constants.js
```diff
@@ -105,7 +105,7 @@ module.exports = Object.freeze({
     SHUTDOWN_WAIT_TIME: 5000
   },
   BACKUP: {
-    BACKUP_START_TIMEOUT: 15, //SECONDS
+    BACKUP_START_TIMEOUT: 60, //SECONDS
```
jagadish-kb: Any reason for bumping this and keeping it at 60? its always better to put time units as part of variable name as it makes it readable at all places as in `BACKUP_START_TIMEOUT_IN_SECS`

data-access-layer/eventmesh/ApiServerClient.js
```diff
@@ -23,22 +24,35 @@ const apiserver = new kc.Client({
 });
 
 function buildErrors(err) {
```
jagadish-kb: shouldnt the method be named `buildErrorAndThrow` instead? or even better would be `convertToHttpErrorAndThrow(err)`
the action of just `build` would generally mean you return the same error and not throw.

data-access-layer/eventmesh/ApiServerClient.js
```diff
@@ -59,6 +73,61 @@ class ApiServerClient {
...
+      .then(() => this.getOperationState({
+        operationName: CONST.OPERATION_TYPE.BACKUP,
+        operationType: CONST.APISERVER.RESOURCE_TYPES.DEFAULT_BACKUP,
+        operationId: opts.operationId
+      })).then(state => {
+        const duration = (new Date() - opts.started_at) / 1000;
+        logger.info(`Polling for ${opts.start_state} duration: ${duration} `);
+        if (duration > CONST.BACKUP.BACKUP_START_TIMEOUT) {
+          logger.error(`Backup with guid ${opts.operationId} not picked up from the queue`);
+          throw new Timeout(`Backup with guid ${opts.operationId} not picked up from the queue`);
```
jagadish-kb: should you also not update the status of the backup as failed and ensure that backup never gets picked up by manager in background?

jagadish-kb: Also, shouldnt this check move to the else clause in LN112? in the last iteration assuming on boundary condition u r really timing out and it was processed (either in-progress or error), but you would be instead responding back as Timedout?

data-access-layer/eventmesh/ApiServerClient.js
```diff
@@ -130,6 +199,15 @@ class ApiServerClient {
         }));
   }
 
+  patchResourceStatus(resourceGroup, resourceType, resourceId, statusDelta) {
+    return Promise.try(() => this.init())
```
jagadish-kb: why is this explicity init() done here?

data-access-layer/eventmesh/ApiServerClient.js
```diff
@@ -310,6 +388,25 @@ class ApiServerClient {
...
 
+  /**
+   * @description Function to Update the error field
+   * @param {string} opts.operationName - Name of operation
+   * @param {string} opts.operationType - Type of operation
+   * @param {string} opts.operationId - Unique id of operation
+   * @param {Object} opts.error - Value to set as error
+   */
+  updateOperationError(opts) {
+    const change = {
```
jagadish-kb: would rather rename this attrib to `response` or `operationStatus`

data-access-layer/eventmesh/ApiServerClient.js
```diff
@@ -130,6 +199,15 @@ class ApiServerClient {
         }));
   }
 
+  patchResourceStatus(resourceGroup, resourceType, resourceId, statusDelta) {
```
jagadish-kb: why is the last param called `delta` its just `status` right?

managers/backup-manager/BackupService.js
```diff
@@ -249,6 +249,26 @@ class BackupService extends BaseDirectorService {
       });
   }
 
+  getLastBackup(tenant_id, instance_guid, noCache) {
```
jagadish-kb: from which invocation is `noCache` set? If this is never going to be used in V2, recommed it to be rather removed.

data-access-layer/eventmesh/ApiServerClient.js
muralisc: if we put status it may mean that the value sent will replace the current status, but that wont be the case.
the `statusDelta` is patched and not updated.
The existing status values are retained and only the changed keys are updated.

managers/backup-manager/BackupService.js
muralisc: Not currently used but this is an optimisation we can have in cases we need to avoid requests to agent.

data-access-layer/eventmesh/ApiServerClient.js
muralisc: Done

data-access-layer/eventmesh/ApiServerClient.js
muralisc: To ensure that loadSpec is already loaded.

data-access-layer/eventmesh/ApiServerClient.js
muralisc: Done

common/constants.js
muralisc: Yes in azure it takes 32 seconds for the backupService to start and update the apiserver.

api-controllers/ServiceFabrikApiController.js
```diff
@@ -339,6 +256,12 @@ class ServiceFabrikApiController extends FabrikBaseController {
       });
   }
 
+  getPlanIdFromInstanceId(instance_id) {
```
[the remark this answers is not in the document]
muralisc: Done

api-controllers/ServiceFabrikApiController.js
```diff
@@ -433,33 +369,39 @@ class ServiceFabrikApiController extends FabrikBaseController {
...
+            // abort only if the state is in progress
+            if (state === CONST.APISERVER.RESOURCE_STATE.IN_PROGRESS) {
+              return eventmesh.apiServerClient.updateOperationState({
+                operationName: CONST.OPERATION_TYPE.BACKUP,
+                operationType: CONST.APISERVER.RESOURCE_TYPES.DEFAULT_BACKUP,
+                operationId: backupGuid,
+                stateValue: CONST.OPERATION.ABORT
+              });
+            } else {
+              logger.info(`Skipping abort as state is : ${state}`);
```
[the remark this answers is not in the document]
muralisc: Done

api-controllers/ServiceFabrikApiController.js
muralisc: Yes, we are having a ApiServerClient refactoring coming up. It has too many wrappers now

api-controllers/ServiceFabrikApiController.js
muralisc: Aborting in sf v1 has been fire and forget . We had retained the same here.

api-controllers/ServiceFabrikApiController.js
```diff
@@ -418,13 +341,26 @@ class ServiceFabrikApiController extends FabrikBaseController {
...
-        return res
-          .status(CONST.HTTP_STATUS_CODE.OK)
-          .send(_.omit(result, 'secret', 'agent_ip', 'description'));
+        })
+      )
+      .catch(NotFound, (err) => {
+        // This code block is specifically for the transition of Service Fabrik to v2
+        // Here we reffer to BackupService to get the lastBackup status
+        logger.info('Backup metadata not found in apiserver, checking blobstore. Error message:', err.message);
+        const tenantId = _.get(req, 'body.space_guid') ||
```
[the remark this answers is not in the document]
muralisc: Not required , removed

api-controllers/ServiceFabrikApiController.js
muralisc: This is like a foreign key for last backup of the deployment. We are getting the last backup operation guid of the deployment `req.params.instance_id`.
The above details are then queried using the backup_guid

common/constants.js
muralisc: Renamed

api-controllers/ServiceFabrikApiController.js
muralisc: Restore is not yet sfv2

data-access-layer/eventmesh/ApiServerClient.js
muralisc: As you have suggested the backup start is now asynchronous.

muralisc: the check for timeout is made after the state check in the upcoming PR https://github.com/cloudfoundry-incubator/service-fabrik-broker/compare/master...muralisc:add-error-field?expand=1#diff-db2af8e82ceb9a902cfd3daa0b0706cfR112

api-controllers/ServiceFabrikApiController.js
jagadish-kb: ok

PR #340, opened 2018-07-22
Fix ongoing backup

muralisc:
- Create backup and deployment resources for backups running in background when service-fabrik transitions from v1 to v2
- Avoid call to backupstore while updating apiserver.

### DIFF
--- a/jobs/BnRStatusPollerJob.js
+++ b/jobs/BnRStatusPollerJob.js
@@ -64,7 +64,42 @@ class BnRStatusPollerJob extends BaseJob {
     const backup_guid = instanceInfo.backup_guid;
     const deployment = instanceInfo.deployment;
     const plan = catalog.getPlan(instanceInfo.plan_id);
-    return Promise.try(() => {
+    return Promise
+      .try(() => {
+        logger.info(`Check if the backup resouce ${backup_guid} and deployment resource ${instance_guid} already exists`);
+        // For transition of sf to sfv2 we need to create bakcup and deployment resource
+        // if they are not present
+        // Can be removed once SFv2 transition is over
+        return eventmesh.apiServerClient.getResource(
+            CONST.APISERVER.RESOURCE_GROUPS.BACKUP,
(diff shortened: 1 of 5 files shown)


jobs/BnRStatusPollerJob.js
```diff
@@ -64,7 +64,42 @@ class BnRStatusPollerJob extends BaseJob {
...
+            });
+          })
+          .then(() => eventmesh.apiServerClient.getResource(
+            CONST.APISERVER.RESOURCE_GROUPS.DEPLOYMENT,
+            CONST.APISERVER.RESOURCE_TYPES.DIRECTOR,
+            instance_guid))
+          .catch(() => {
+            logger.info(`Creating missing operation resource for instance ${instance_guid}`);
+            return eventmesh.apiServerClient.createDeployment(instance_guid, {});
+          })
```
amitmalav: Can create Deployment Resource and then create Backup Resource

managers/backup-manager/BackupService.js
```diff
@@ -285,14 +285,21 @@ class BackupService extends BaseDirectorService {
                 logs: logs,
                 snapshotId: lastOperation.snapshotId
               })
+              .return({
+                state: lastOperation.state,
+                logs: logs,
+                snapshotId: lastOperation.snapshotId,
+                finished_at: new Date(Date.now())
```
subhankarc: If we are setting the date here, will the finished at value in backupstore and in apiserver would be different isn't it?

muralisc: yes . but since out least count is in minutes , this should be acceptable ? 
Or we can pass down the date created here down the function chain .

subhankarc: fine then.

jobs/BnRStatusPollerJob.js
muralisc: Done.

kamath-prasad:
@muralisc : You need to raise a PR on 7A branch as well.

 @subhankarc : please merge the PR if changes are fine.

PR #343, opened 2018-07-23
Adding helper class EncryptionManager

subhankarc:
* Adding Encryptionmanager which has encrypt and decrypt method to decrypt based on aes-256

* Adding test cases

* Currently it is not used, to be used by ApiServerClient to encrypt the bind response

### DIFF
--- a/broker/config/settings.yml
+++ b/broker/config/settings.yml
@@ -193,6 +193,11 @@ defaults: &defaults
     ip: '127.0.0.1'
     port: 9443
 
+  apiserver:
+    ip: '127.0.0.1'
+    port: 9443
+    encryption:
+      password: 'ABCDEFGH'
 
   ##############################
   # REPORT ENDPOINT SETTINGS #
--- a/broker/config/settings.yml
+++ b/broker/config/settings.yml
@@ -197,7 +197,8 @@ defaults: &defaults
     ip: '127.0.0.1'
     port: 9443
     encryption:
-      password: 'ABCDEFGH'
+      password: 'ABCDEFGHABCDEFGHABCDEFGHABCDEFGH'
+      initialization_vector: 'ABCDEFGHABCDEFGH'
 
   ##############################
   # REPORT ENDPOINT SETTINGS #
--- a/broker/config/settings.yml
+++ b/broker/config/settings.yml
@@ -197,7 +197,7 @@ defaults: &defaults
     ip: '127.0.0.1'
     port: 9443
     encryption:
-      password: 'ABCDEFGHABCDEFGHABCDEFGHABCDEFGH'
+      key: 'ABCDEFGHABCDEFGHABCDEFGHABCDEFGH'
       initialization_vector: 'ABCDEFGHABCDEFGH'
 
   ##############################
--- a/common/constants.js
+++ b/common/constants.js
@@ -239,11 +239,6 @@ module.exports = Object.freeze({
     HOSTNAME: 'servicefabrik.io',
     NAMESPACE: 'default',
     API_VERSION: 'v1alpha1',
-    ENCRYPTION: {
-      AES_256_ALGORITHM: 'aes-256-ctr',
-      INPUT_ENCODING: 'utf8',
-      OUTPUT_ENCODING: 'hex'
-    },
     RESOURCE_GROUPS: {
       LOCK: 'lock',
       DEPLOYMENT: 'deployment',
@@ -287,7 +282,11 @@ module.exports = Object.freeze({
     ATTRIBUTES: 'attributes',
     PLANS: 'plans'
   },
-
+  ENCRYPTION: {
+    AES_256_ALGORITHM: 'aes-256-ctr',
+    INPUT_ENCODING: 'utf8',
+    OUTPUT_ENCODING: 'hex'
+  },
   PLATFORM: {
     CF: 'cloudfoundry',
     K8S: 'kubernetes'
--- a/common/constants.js
+++ b/common/constants.js
@@ -239,6 +239,11 @@ module.exports = Object.freeze({
     HOSTNAME: 'servicefabrik.io',
     NAMESPACE: 'default',
     API_VERSION: 'v1alpha1',
+    ENCRYPTION: {
+      AES_256_ALGORITHM: 'aes-256-ctr',
+      INPUT_ENCODING: 'utf8',
+      OUTPUT_ENCODING: 'hex'
+    },
     RESOURCE_GROUPS: {
       LOCK: 'lock',
       DEPLOYMENT: 'deployment',
--- a/common/utils/EncryptionManager.js
+++ b/common/utils/EncryptionManager.js
@@ -10,9 +10,9 @@ class EncryptionManager {
    * @param text test to be encrypted
    */
   encrypt(text) {
-    const cipher = crypto.createCipher(CONST.APISERVER.ENCRYPTION.AES_256_ALGORITHM, config.apiserver.encryption.password);
-    let crypted = cipher.update(text, CONST.APISERVER.ENCRYPTION.INPUT_ENCODING, CONST.APISERVER.ENCRYPTION.OUTPUT_ENCODING);
-    crypted += cipher.final(CONST.APISERVER.ENCRYPTION.OUTPUT_ENCODING);
+    const cipher = crypto.createCipheriv(CONST.ENCRYPTION.AES_256_ALGORITHM, config.apiserver.encryption.password, config.apiserver.encryption.initialization_vector);
+    let crypted = cipher.update(text, CONST.ENCRYPTION.INPUT_ENCODING, CONST.ENCRYPTION.OUTPUT_ENCODING);
+    crypted += cipher.final(CONST.ENCRYPTION.OUTPUT_ENCODING);
     return crypted;
   }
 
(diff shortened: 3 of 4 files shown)


broker/config/settings.yml
```diff
@@ -187,6 +187,11 @@ defaults: &defaults
       key: <%= certificate('broker.key') %>
       cert: <%= certificate('broker.crt') %>
 
+  apiserver:
+    ip: '127.0.0.1'
+    port: 9443
+    encryption:
+      password: 'ABCDEFGH'
```
abh1kg: This encryption key must remain constant once generated, right? So this needs to be marked as immutable. Is it possible to somehow make the encryption not dependent on an immutable constant?

common/constants.js
```diff
@@ -239,6 +239,11 @@ module.exports = Object.freeze({
     HOSTNAME: 'servicefabrik.io',
     NAMESPACE: 'default',
     API_VERSION: 'v1alpha1',
+    ENCRYPTION: {
```
abh1kg: Would prefer if it is moved out of the APISERVER object since it may be required later for some other operation

common/utils/EncryptionManager.js
```diff
@@ -0,0 +1,31 @@
...
+const config = require('../config');
+const CONST = require('../constants');
+
+class EncryptionManager {
+
+  /**
+   * @param text test to be encrypted
+   */
+  encrypt(text) {
+    const cipher = crypto.createCipher(CONST.APISERVER.ENCRYPTION.AES_256_ALGORITHM, config.apiserver.encryption.password);
```
abh1kg: Just a note: the createCipher function without an initialization vector is currently deprecated [see: https://nodejs.org/api/crypto.html#crypto_crypto_createcipher_algorithm_password_options ].

common/utils/EncryptionManager.js
```diff
@@ -0,0 +1,31 @@
...
+    let crypted = cipher.update(text, CONST.APISERVER.ENCRYPTION.INPUT_ENCODING, CONST.APISERVER.ENCRYPTION.OUTPUT_ENCODING);
```
abh1kg: Though not a big problem but note that this is a synchronous operation which blocks the event loop.

broker/config/settings.yml
subhankarc: good point...actually I was also thinking of the same. the problem with config is that one can regenerate and change it and then we may loose old bindings!

I can't think of any option as such. Do you have anything in mind? @jagadish-kb @kamath-prasad  may chime in here.

common/constants.js
subhankarc: Done!

common/utils/EncryptionManager.js
subhankarc: Changed to createCipheriv

common/utils/EncryptionManager.js
subhankarc: anything you would want to change?

subhankarc:
@abh1kg @jagadish-kb @kamath-prasad @amitmalav 

Keeping this discussion in a separate comment as @abh1kg 's comment became outdated.

Currently the initialization vector and password are kept in the config, which means one has to generate them. But the for decrypt to work, we would need the keys to be constant, as otherwise if the keys are renewed, the decryption would not work. How do we tackle this?

broker/config/settings.yml
kamath-prasad: @subhankarc : Not sure how exactly this methods will be used in the bind operation . One option I could think of is - can we store the keys/vector along with the encrypted string ? Maybe as part of bind credentials ? Then we can generate these keys on the fly ?

test/test_broker/utils.EncryptionManager.spec.js
```diff
@@ -0,0 +1,39 @@
...
+const EncryptionManager = require('../../common/utils/EncryptionManager');
+
+describe('utils', () => {
+  describe('EncryptionManager', () => {
+    /* jshint expr:true */
+    describe('encrypt-decrypt-simple', () => {
+      it('returns the same string after decryption', () => {
+        const manager = new EncryptionManager();
+        const testText = 'Hello World';
+        console.log(testText);
```
kamath-prasad: console.log required ?

test/test_broker/utils.EncryptionManager.spec.js
```diff
@@ -0,0 +1,39 @@
...
+        const encryptedText = manager.encrypt(testText);
+        console.log(encryptedText);
```
kamath-prasad: same here ?

broker/config/settings.yml
subhankarc: If the key is at the same place, it would be very easy to attack right?

common/constants.js
```diff
@@ -282,7 +282,11 @@ module.exports = Object.freeze({
     ATTRIBUTES: 'attributes',
     PLANS: 'plans'
   },
-
+  ENCRYPTION: {
+    AES_256_ALGORITHM: 'aes-256-ctr',
```
sauravmndl: Instead of using **aes-256-ctr** can we use **aes-256-gcm** as GCM achieves **confidentiality** and **integrity** both where CTR supports only confidentiality?

broker/config/settings.yml
```diff
@@ -187,6 +187,12 @@ defaults: &defaults
       key: <%= certificate('broker.key') %>
       cert: <%= certificate('broker.crt') %>
 
+  apiserver:
+    ip: '127.0.0.1'
+    port: 9443
+    encryption:
+      password: 'ABCDEFGHABCDEFGHABCDEFGHABCDEFGH'
+      initialization_vector: 'ABCDEFGHABCDEFGH'
```
sauravmndl: suggestion: while putting in actual config generate using random byte generator as IV is recommended as cryptographically random and unpredictable. something like following
`crypto.randomBytes(64).toString('hex')`

broker/config/settings.yml
```diff
@@ -187,6 +187,12 @@ defaults: &defaults
       key: <%= certificate('broker.key') %>
       cert: <%= certificate('broker.crt') %>
 
+  apiserver:
+    ip: '127.0.0.1'
+    port: 9443
+    encryption:
+      password: 'ABCDEFGHABCDEFGHABCDEFGHABCDEFGH'
```
sauravmndl: you can just name the parameter 'password' --> 'key' as to be consistent with **crypto.createCipheriv** and **crypto.createDecipheriv** methods' parameters-naming.

common/utils/EncryptionManager.js
abh1kg: No, let's keep it like this for now. There are asynchronous versions via C++ node modules available but that might add some build-level complexity.

broker/config/settings.yml
subhankarc: settings.yml is just an example.

jagadish-kb: +1 - this could be injected into from manifet.erb from product component. Agree above is just sample.

common/constants.js
jagadish-kb: might be good to check if there are some corp/org guidelines on the same.

subhankarc: change of plan!.....keeping it generic so that the consumer can choose the algorithm!

common/utils/EncryptionManager.js
```diff
@@ -0,0 +1,31 @@
...
+
+class EncryptionManager {
+
+  /**
+   * @param text test to be encrypted
+   */
+  encrypt(text) {
+    const cipher = crypto.createCipheriv(CONST.ENCRYPTION.AES_256_ALGORITHM, config.apiserver.encryption.password, config.apiserver.encryption.initialization_vector);
+    let crypted = cipher.update(text, CONST.ENCRYPTION.INPUT_ENCODING, CONST.ENCRYPTION.OUTPUT_ENCODING);
+    crypted += cipher.final(CONST.ENCRYPTION.OUTPUT_ENCODING);
```
jagadish-kb: Some code comments here about LN 14 & 15 will help. This could be std usage from lib functions, but just from a readability part, it would help as LN15 is concat with LN14.

common/utils/EncryptionManager.js
```diff
@@ -0,0 +1,31 @@
+'use strict';
+
+const crypto = require('crypto');
+const config = require('../config');
+const CONST = require('../constants');
+
+class EncryptionManager {
+
+  /**
```
jagadish-kb: Are these JS docs helpful? We started adding these in the last couple of sprints, but most of the time they seem to self explanatory from function signatures. The reason am questioning this is, this makes our code have JSDoc on some files and not have in others. So i would prefer to have consistency.
Also, if we are not auto generating api documentation from JSDOC, then probably i would omit it.

subhankarc: will remove it for this. ;)

common/utils/EncryptionManager.js
subhankarc: actually the final is not required. final is called only when you call update multiple times. I can concat them.

subhankarc: actually I can remove the final call.

subhankarc:
@jagadish-kb Can you please have a look and if all are fine then approve and merge it?

common/constants.js
jagadish-kb: Dont think its a good idea to parameterise as it leads to additional validation at EncryptionManager on the algorithm name passed. Also different modules could encrypt in different algorithms. Its best we enforce this for two reasons:
1. Standardisation across the product
2. Also, avoid errors which could occur when modules might change the algorithm across releases.

So if we want to fix it to `aes-256-ctr` lets go ahead and fix it in this PR and when we really want to start using the encryption then we can check for org guidelines.

common/utils/EncryptionManager.js
```diff
@@ -0,0 +1,25 @@
+'use strict';
+
+const crypto = require('crypto');
+const config = require('../config');
+const CONST = require('../constants');
+
+class EncryptionManager {
+
+  constructor(algorithm) {
+    this.algorithm = algorithm;
```
jagadish-kb: should not be doing this as explained above. as we need to additionally validate etc.

common/constants.js
subhankarc: Agree @jagadish-kb . 
I have done the changes accordingly.

Issue #347, opened 2018-08-01
Set LockTTL for ApiServer Locks


amitmalav:
**Description**
Currently there is no TTL for apiserver locks. In case of long running/faulty operations (ex: backup) the lock is never released, Which can lead to blocking of all other operations.

**Solution Approach**
There are 2 ways to achieve locking and unlocking mechanism.
1. To lock a deployment, we create a resource of type deploymentlock and to unlock this lock resource is deleted.
2. We can have one state label on lock resource which will reflect state of lock ('locked'/'unlocked')

Algorithm 1 is as follows

**lock**
```javascript
lock(instance_id, lockDetails):
  getLockResource with instance_id
  if found:
    checkLockTTL:
      if expired:
        updateLockResource with lockDetails
      else:
        return DeploymentAlreadyLocked error
  else:
    create lock resource with instance_id, lockDetails
```
**unlock**
```javascript
unlock(instance_id):
  delete lock resource with instance_id
```
Shortcomings of this approach.
While locking and unlocking we want to ensure atomicity. With this approach we may encounter race conditions discussed below with example of two backup operations where lock is already acquired by B1 and is about to expire.
```javascript
    B1                                     B2
-- Locked by B1    
-- unlock call(1)
-- lock expires
--                                  acquire lock
-- (1) deletes lock acquired by B2
```
Here the problem we see is that B1 is trying to force delete the lock resource which doesn't even belong to it anymore. Now we can solve this problem by associating lockId with the lock resource which is unique to the resource (resourceVersion) and caller will pass this id when unlocking. But now what? ApiServerDelete doesn't entertain any query based requests (like delete only if resourceVersion is X).
One Better way to achieve this would be to use patch rather than delete when unlocking and pass 
 lockId, which is essentially resourceVersion of lock resource at creation time, in patch call as metadata.resourceVersion. It will ensure that patch call won't go through if resource has been updated.

Proposed Algorithm

**lock**
```javascript
lock(instance_id, lockDetails):
  getLockResource with instance_id
  if found:
    checkLockTTL:
      if not expired && resource state !== unlocked:
         return DeploymentAlreadyLocked error
      else:
        updateLockResource with lockDetails and state as 'locked' and return resourceVersion of updated resource
  else:
    create lock resource with instance_id, lockDetails and state as 'locked' and return resourceVersion of created resource
```
**unlock**
```javascript
unlock(instance_id, lockId):
[...]

PR #348, opened 2018-08-02
Lock TTL Implementation for ApiServer locks

amitmalav:
Fixes: #347

### DIFF
--- a/api-controllers/FabrikBaseController.js
+++ b/api-controllers/FabrikBaseController.js
@@ -36,15 +36,17 @@ class FabrikBaseController extends BaseController {
     return (req, res, next) => {
       let resourceLocked = false;
       let processedRequest = false;
+      let lockId;
       return this._lockResource(req, operationType)
         .tap(() => resourceLocked = true)
-        .then(() => {
+        .then(lockResourceId => {
+          lockId = lockResourceId;
           const fn = _.isString(func) ? this[func] : func;
           return fn.call(this, req, res);
         })
         .tap(() => processedRequest = true)
-        .then(() => this._unlockIfReqfailed(operationType, processedRequest, req, res, next))
-        .catch(err => resourceLocked ? this._unlockIfReqfailed(operationType, processedRequest, req, res, next, err) : next(err));
+        .then(() => this._unlockIfReqfailed(operationType, processedRequest, lockId, req, res, next))
+        .catch(err => resourceLocked ? this._unlockIfReqfailed(operationType, processedRequest, lockId, req, res, next, err) : next(err));
     };
   }
 
@@ -59,14 +61,14 @@ class FabrikBaseController extends BaseController {
             resourceGroup: CONST.APISERVER.RESOURCE_GROUPS.DEPLOYMENT,
             resourceType: CONST.APISERVER.RESOURCE_TYPES.DIRECTOR,
             resourceId: req.params.instance_id,
-            operation: operationType ? operationType : req.query.operation.type // This is for the last operation call
+            operation: operationType
           }
         });
       }
     });
   }
 
-  _unlockIfReqfailed(operationType, processedRequest, req, res, next, err) {
+  _unlockIfReqfailed(operationType, processedRequest, lockId, req, res, next, err) {
     const plan_id = req.body.plan_id || req.query.plan_id;
     const plan = catalog.getPlan(plan_id);
     // If processed request
@@ -80,11 +82,11 @@ class FabrikBaseController extends BaseController {
               operationType === CONST.OPERATION_TYPE.CREATE && res.statusCode === CONST.HTTP_STATUS_CODE.CONFLICT || // PutInstance => unlock in case of 409
               operationType === CONST.OPERATION_TYPE.DELETE && res.statusCode === CONST.HTTP_STATUS_CODE.GONE // DeleteInstance => unlock in case of 410
             ) {
-              return lockManager.unlock(req.params.instance_id)
+              return lockManager.unlock(req.params.instance_id, lockId)
                 .catch(unlockErr => next(unlockErr));
             }
           } else {
-            return lockManager.unlock(req.params.instance_id)
+            return lockManager.unlock(req.params.instance_id, lockId)
               .then(() => {
                 if (err) {
                   if (err instanceof ContinueWithNext) {
--- a/api-controllers/ServiceBrokerApiController.js
+++ b/api-controllers/ServiceBrokerApiController.js
@@ -7,7 +7,6 @@ const logger = require('../common/logger');
 const utils = require('../common/utils');
 const catalog = require('../common/models/catalog');
 const FabrikBaseController = require('./FabrikBaseController');
-const lockManager = require('../data-access-layer/eventmesh').lockManager;
 const BadRequest = errors.BadRequest;
 const PreconditionFailed = errors.PreconditionFailed;
 const NotFound = errors.NotFound;
@@ -246,25 +245,18 @@ class ServiceBrokerApiController extends FabrikBaseController {
 
     function done(result) {
       const body = _.pick(result, 'state', 'description');
-      // Unlock resource if state is succeeded or failed
-      if (result.state === CONST.OPERATION.SUCCEEDED || result.state === CONST.OPERATION.FAILED) {
-        return lockManager.unlock(req.params.instance_id) //TODO-JB - to be migrated as all unlock would happen via unlock resource poller.
-          .then(() => res.status(CONST.HTTP_STATUS_CODE.OK).send(body));
-      }
       res.status(CONST.HTTP_STATUS_CODE.OK).send(body);
     }
 
     function failed(err) {
-      return lockManager.unlock(req.params.instance_id)
-        .then(() => res.status(CONST.HTTP_STATUS_CODE.OK).send({
-          state: CONST.OPERATION.FAILED,
-          description: `${action} ${instanceType} '${guid}' failed because "${err.message}"`
-        }));
+      res.status(CONST.HTTP_STATUS_CODE.OK).send({
+        state: CONST.OPERATION.FAILED,
+        description: `${action} ${instanceType} '${guid}' failed because "${err.message}"`
+      });
     }
 
     function gone() {
-      return lockManager.unlock(req.params.instance_id)
-        .then(() => res.status(CONST.HTTP_STATUS_CODE.GONE).send({}));
+      res.status(CONST.HTTP_STATUS_CODE.GONE).send({});
     }
 
     function notFound(err) {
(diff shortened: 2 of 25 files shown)


data-access-layer/eventmesh/LockManager.js
```diff
@@ -199,11 +221,15 @@ class LockManager {
...
-      return CONST.ETCD.LOCK_TYPE.WRITE;
+      return CONST.APISERVER.LOCK_TYPE.WRITE;
     } else {
-      return CONST.ETCD.LOCK_TYPE.READ;
+      return CONST.APISERVER.LOCK_TYPE.READ;
     }
   }
+  _getLockTTL(operation) {
+    /* jshint unused:false */
+    return 10000; // Shall be fetched from config
```
amitmalav: Will have to get it from config.

data-access-layer/eventmesh/LockManager.js
```diff
@@ -131,19 +127,28 @@ class LockManager {
             resourceGroup: CONST.APISERVER.RESOURCE_GROUPS.LOCK,
             resourceType: CONST.APISERVER.RESOURCE_TYPES.DEPLOYMENT_LOCKS,
             resourceId: resourceId,
-            options: opts
+            metadata: resource.metadata,
```
jagadish-kb: while updating shouldn't we send resource version as well? Looked at the code in apiserver.unlock and not sure how that is handling with versions? Because thats what we are to use to avoid concurrent writes.

data-access-layer/eventmesh/LockManager.js
```diff
@@ -167,28 +172,45 @@ class LockManager {
...
+      };
+      if (lockId) {
+        opts.metadata = {
+          resourceVersion: lockId
+        };
+      }
+      return eventmesh.apiServerClient.updateResource(opts)
         .tap(() => logger.debug(`Successfully unlocked resource ${resourceId} `))
         .catch(err => {
+          if (err instanceof Conflict) {
```
jagadish-kb: can leverage `.catch(conflict, (err)=> {})` instead of the above syntax for consistency across code base.

data-access-layer/eventmesh/LockManager.js
```diff
@@ -167,28 +172,45 @@ class LockManager {
...
+            logger.info(`Lock on resource ${resourceId} has been updated by some other operation because it expired, no need to unlock now`, err);
+            return;
+          }
           if (err instanceof NotFound) {
-            logger.debug(`Successfully Unlocked resource ${resourceId} `);
+            logger.debug(`Lock on resource ${resourceId} not found`, err);
```
jagadish-kb: When would this happen in the new flow? This error should never occur right as lock will always be present? So better to throw this error in this scenario?

data-access-layer/eventmesh/LockManager.js
jagadish-kb: Didn't understand the above? Will the same be updated in this PR itself?

jobs/BnRStatusPollerJob.js
```diff
@@ -92,26 +92,55 @@ class BnRStatusPollerJob extends BaseJob {
               }
             });
           })
-          .then(() => eventmesh.apiServerClient.getResource({
-            resourceGroup: CONST.APISERVER.RESOURCE_GROUPS.BACKUP,
-            resourceType: CONST.APISERVER.RESOURCE_TYPES.DEFAULT_BACKUP,
-            resourceId: backup_guid
+          .then(() => lockManager.lock(instance_guid, {
```
jagadish-kb: is this for putting lock for long running backups from older code? is this required? SF2.0 for backup is already in 6A, so this might not be required now right?

jobs/BnRStatusPollerJob.js
```diff
@@ -92,26 +92,55 @@ class BnRStatusPollerJob extends BaseJob {
...
+              resourceType: CONST.APISERVER.RESOURCE_TYPES.DEFAULT_BACKUP,
+              resourceId: backup_guid,
+              operation: CONST.OPERATION_TYPE.BACKUP
+            }
           }))
-          .catch(NotFound, () => {
-            logger.info(`Creating missing operation resource for backup ${backup_guid}`);
-            return eventmesh.apiServerClient.createResource({
+          .catch(err => {
+            if (err instanceof DeploymentAlreadyLocked) {
```
jagadish-kb: use the error type in catch itself instead of such checks like above. Also this might seem unnecessary as per above comment.

jobs/BnRStatusPollerJob.js
```diff
@@ -92,26 +92,55 @@ class BnRStatusPollerJob extends BaseJob {
...
+              metadata: {
+                labels: {
+                  lockId: lockId
+                }
               }
-            });
-          })
+            })
+            .catch(NotFound, () => {
+              logger.info(`Creating missing operation resource for backup ${backup_guid}`);
```
jagadish-kb: confusing, why are we doing this. I dont think this would be required.

api-controllers/FabrikBaseController.js
```diff
@@ -75,11 +77,11 @@ class FabrikBaseController extends BaseController {
               operationType === CONST.OPERATION_TYPE.CREATE && res.statusCode === CONST.HTTP_STATUS_CODE.CONFLICT || // PutInstance => unlock in case of 409
               operationType === CONST.OPERATION_TYPE.DELETE && res.statusCode === CONST.HTTP_STATUS_CODE.GONE // DeleteInstance => unlock in case of 410
             ) {
-              return lockManager.unlock(req.params.instance_id)
+              return lockManager.unlock(req.params.instance_id, req.lockId)
```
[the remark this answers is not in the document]
amitmalav: The idea is that whoever acquires lock should have the lockId as it is needed in PATCH call to handle race conditions.
So either it is passed to `lockManager.unlock()` (1) or unlock can get it from the resource who is trying to unlock(2)
In 2nd case also then we will have to pass `resourceGroup` and `resourceType` and `resourceId` as we can't rely on `spec.options.lockedResourceDetails` of lock resource as it might have changed (race condition)

api-controllers/FabrikBaseController.js
```diff
@@ -37,7 +37,9 @@ class FabrikBaseController extends BaseController {
       let processedRequest = false;
       return this._lockResource(req, operationType)
         .tap(() => resourceLocked = true)
-        .then(() => {
+        .then((lockId) => {
+          // Has to be put in deployment resource as label
+          req.lockId = lockId;
```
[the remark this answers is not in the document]
amitmalav: Whoever is acquiring the lock, it's their responsibility to store the lockId "somewhere" as they will be needing it while unlocking. It makes sense for lockManger to just return lockId and caller can patch it to their resource.

api-controllers/FabrikBaseController.js
```diff
@@ -56,7 +58,7 @@ class FabrikBaseController extends BaseController {
             resourceGroup: CONST.APISERVER.RESOURCE_GROUPS.DEPLOYMENT,
             resourceType: CONST.APISERVER.RESOURCE_TYPES.DIRECTOR,
             resourceId: req.params.instance_id,
-            operation: operationType ? operationType : req.query.operation.type // This is for the last operation call
+            operation: operationType
```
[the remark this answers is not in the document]
amitmalav: Yes.

api-controllers/ServiceBrokerApiController.js
```diff
@@ -136,22 +136,25 @@ class ServiceBrokerApiController extends FabrikBaseController {
       const body = _.pick(result, 'state', 'description');
       // Unlock resource if state is succeeded or failed
       if (result.state === CONST.OPERATION.SUCCEEDED || result.state === CONST.OPERATION.FAILED) {
-        return lockManager.unlock(req.params.instance_id)
+        // shall be fetched from deployment resource once lc managers are onboarded
+        return lockManager.unlock(req.params.instance_id, req.lockId)
```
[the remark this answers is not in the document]
amitmalav: please refer comment above

api-controllers/ServiceBrokerApiController.js
```diff
@@ -136,22 +136,25 @@ class ServiceBrokerApiController extends FabrikBaseController {
...
           .then(() => res.status(CONST.HTTP_STATUS_CODE.OK).send(body));
       }
       res.status(CONST.HTTP_STATUS_CODE.OK).send(body);
     }
 
     function failed(err) {
-      return lockManager.unlock(req.params.instance_id)
+      // shall be fetched from deployment resource once lc managers are onboarded
+      return lockManager.unlock(req.params.instance_id, req.lockId)
```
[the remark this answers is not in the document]
amitmalav: please refer comment above

api-controllers/ServiceBrokerApiController.js
```diff
@@ -136,22 +136,25 @@ class ServiceBrokerApiController extends FabrikBaseController {
...
         .then(() => res.status(CONST.HTTP_STATUS_CODE.OK).send({
           state: CONST.OPERATION.FAILED,
           description: `${action} ${instanceType} '${guid}' failed because "${err.message}"`
         }));
     }
 
     function gone() {
-      return lockManager.unlock(req.params.instance_id)
+      // shall be fetched from deployment resource once lc managers are onboarded
+      return lockManager.unlock(req.params.instance_id, req.lockId)
```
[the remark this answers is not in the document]
amitmalav: please refer comment above

api-controllers/ServiceFabrikApiController.js
```diff
@@ -279,13 +279,17 @@ class ServiceFabrikApiController extends FabrikBaseController {
                   operation: CONST.OPERATION_TYPE.BACKUP
                 }
               })
-              .then(() => {
+              .then((lockId) => {
```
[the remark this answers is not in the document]
amitmalav: Removed

api-controllers/ServiceFabrikApiController.js
```diff
@@ -279,13 +279,17 @@ class ServiceFabrikApiController extends FabrikBaseController {
...
                 lockedDeployment = true;
+                req.lockId = lockId;
                 return eventmesh.apiServerClient.createResource({
                   resourceGroup: CONST.APISERVER.RESOURCE_GROUPS.BACKUP,
                   resourceType: CONST.APISERVER.RESOURCE_TYPES.DEFAULT_BACKUP,
                   resourceId: backupGuid,
-                  parentResourceId: req.params.instance_id,
+                  labels: {
+                    instance_guid: req.params.instance_id,
+                    lockId: lockId
```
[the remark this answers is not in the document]
amitmalav: Not used anywhere. Just for consistency decided to keep it for backup resources also.

data-access-layer/eventmesh/LockManager.js
amitmalav: resourceVersion is part of metadata.

jagadish-kb:
Only UnlockResourcePoller should do the unlocking and lockid should not be returned by lock method as it can be potentially used by code at other places to unlock which should not be ideally unlocking.

PR #349, opened 2018-08-03
Support fabrik-internal services

steinri:
The service fabrik currently supports internal [service plans](https://github.com/cloudfoundry-incubator/service-fabrik-broker/blob/2018.T07b/common/models/Service.js#L40), but not internal services. 

This is used to define a different plan for the mongodb, which is used by the service fabrik, than for mongodbs provided by the service fabrik via the catalog. However if the service fabrik does not provide a mongodb service this mechanism does not work. Hence this adds an internal service in the same manner as the internal service plans.

### DIFF
--- a/common/models/catalog.js
+++ b/common/models/catalog.js
@@ -31,7 +31,7 @@ class Catalog {
 
   toJSON() {
     return {
-      services: this.services
+      services: _.filter(this.services, service => service.name.indexOf('-fabrik-internal') === -1)
     };
   }
 }


steinri:
The motivation behind this is for SF's usage, i.e. deploy a mongodb for the service fabrik (service-fabrik-mongodb).

If I am not mistaken this is currently not possible without having the mongodb plan in the catalog. However than you currently must provide a mongodb service, but can add internal plan. This just leverages the idea of an internal plan to an internal service, such that you don't need to provide a mongodb anymore to use the service fabrik with mongodb.

kamath-prasad:
@steinri : Ok . , got it.. 

I see that this is already handled [here](https://github.com/cloudfoundry-incubator/service-fabrik-broker/blob/master/common/models/Service.js#L40) .

So your change might not be needed.

steinri:
This is handled there for service plans, but not for whole services.
If I would create a mongodb service and mark every service plan as fabrik-internal it would still have the mongodb service in the catalog, but the plan section would be empty.

Hence the catalog would look like:

```
{
  "services": [
    { <my_service> },
    {
      "id": "foo-baar",
      "name": "mongodb-fabrik-internal",
      "description": "...",
      "bindable": true,
      "subnet": null,
      "tags": [
      ],
      "metadata": {
        "displayName": "MongoDB",
        "documentationUrl": "...",
        "longDescription": "....",
        "providerDisplayName": "...",
        "supportUrl": "...."
      },
      "requires": [],
      "plan_updateable": true,
      "dashboard_client": {
        "id": "....",
        "redirect_uri": "...,
        "secret": "...."
      },
      "plans": []  // <-- no plans
    }
  ]
}
```
instead of
```
{
  "services": [
    { <my_service> }
  ]
}
```

PR #355, opened 2018-08-08
Deleting Transaction Logs after successful backup

pritishmishra:
Changes include:
1. Adding feature of Transcation-Logs cleanup. (Delete transaction logs taken after the latest successful backup)
a. As part of this, we look for successful backup beyond the retention period and delete transaction logs from that backup time onwards in the past.
e.g. suppose u have a successful backup on 15th day.. it will delete transaction logs from 15th day onwards till the past
b. If there is no successful backup found, we delete the transaction logs older than retention period(+1day), (comparing the retention period from today).
e.g. If today is the 17th day, we will delete all transaction logs older than 15 days (taken on 1st & 2nd day). 
Assuming retention period is 14days here.
2. Modifying existing UTs to also handle transaction-logs cleanup
3. Adding new UTs to cover various scenarios of transaction-logs cleanup 

Related service-fabrik-boshrelease PR:
https://github.com/cloudfoundry-incubator/service-fabrik-boshrelease/pull/112

### DIFF
--- a/data-access-layer/iaas/BackupStore.js
+++ b/data-access-layer/iaas/BackupStore.js
@@ -34,8 +34,7 @@ class BackupStore {
 
   listFilenames(prefix, predicate, iteratees, container, dontParseFilename) {
     const options = {
-      prefix: prefix,
-      container: container
+      prefix: prefix
     };
 
     const self = this;
@@ -46,8 +45,7 @@ class BackupStore {
       level++;
       logger.debug(`Fetching recursively at level : ${level}`);
       const promise = new Promise(function (resolve, reject) {
-        self.cloudProvider
-          .list(options)
+        Promise.try(() => container ? self.cloudProvider.list(container, options) : self.cloudProvider.list(options))
           .then(files => {
             logger.debug('list of files recieved - ', files);
             if (files && files.length > 0) {
--- a/data-access-layer/iaas/BackupStore.js
+++ b/data-access-layer/iaas/BackupStore.js
@@ -32,7 +32,7 @@ class BackupStore {
     return _.nth(/^(.+)-broker$/.exec(this.containerName), 1);
   }
 
-  listFilenames(prefix, predicate, iteratees) {
+  listFilenames(prefix, predicate, iteratees, dontParseFilename) {
     const options = {
       prefix: prefix
     };
@@ -72,7 +72,7 @@ class BackupStore {
       .then(() =>
         _
         .chain(fileList)
-        .map(file => this.filename.parse(file.name))
+        .map(file => dontParseFilename ? file : this.filename.parse(file.name))
         .filter(predicate)
         .sortBy(iteratees)
         .value());
@@ -274,6 +274,22 @@ class BackupStore {
     return Promise
       .try(() => this.filename.isoDate(backupStartedBefore))
       .then(isoDate => this.listBackupFiles(options, getPredicate(isoDate), iteratees));
+  }
+
+  listTransactionLogsOlderThan(options, dateOlderThan) {
+    const iteratees = ['lastModified'];
+    let prefix = options.prefix;
+
+    function getPredicate(isoDate) {
+      return function predicate(filenameobject) {
+        //transactionLogsDeletionStartDate defaults to current timestamp as part of isoDate function.
+        return _.lt(filenameobject.lastModified, isoDate);
+      };
+    }
+
+    return Promise
+      .try(() => this.filename.isoDate(dateOlderThan))
+      .then(isoDate => this.listFilenames(prefix, getPredicate(isoDate), iteratees, true));
   }
 }
 
--- a/data-access-layer/iaas/BackupStore.js
+++ b/data-access-layer/iaas/BackupStore.js
@@ -277,7 +277,7 @@ class BackupStore {
       .then(isoDate => this.listBackupFiles(options, getPredicate(isoDate), iteratees));
   }
 
-  listTransactionLogsOlderThan(options, dateOlderThan, container) {
+  listFilesOlderThan(options, dateOlderThan, container) {
     const iteratees = ['lastModified'];
     let prefix = options.prefix;
 
--- a/data-access-layer/iaas/BackupStore.js
+++ b/data-access-layer/iaas/BackupStore.js
@@ -275,9 +275,10 @@ class BackupStore {
       .then(isoDate => this.listBackupFiles(options, getPredicate(isoDate), iteratees));
   }
 
-  listFilesOlderThan(options, dateOlderThan, container) {
+  listTransactionLogsOlderThan(options, dateOlderThan) {
     const iteratees = ['lastModified'];
-    let prefix = options.prefix;
+    const prefix = `${options.service_name}/logs/${options.instance_id}`;
+    const container = `${this.containerPrefix}-${options.service_name}`;
 
     function getPredicate(isoDate) {
       return function predicate(filenameobject) {
@@ -285,10 +286,23 @@ class BackupStore {
         return _.lt(new Date(filenameobject.lastModified).toISOString(), isoDate);
       };
     }
-
     return Promise
       .try(() => this.filename.isoDate(dateOlderThan))
       .then(isoDate => this.listFilenames(prefix, getPredicate(isoDate), iteratees, container, true));
+  }
+
+  deleteTransactionLogsOlderThan(options, dateOlderThan) {
+    let deletedTransactionLogFilesCount = 0;
+    const container = `${this.containerPrefix}-${options.service_name}`;
+
+    return this.listTransactionLogsOlderThan(options, dateOlderThan)
+      .tap(transactionLogsFiles => {
+        if (!_.isEmpty(transactionLogsFiles)) {
+          deletedTransactionLogFilesCount = transactionLogsFiles.length;
+        }
+      })
+      .map(transactionLogsFile => this.cloudProvider.remove(container, transactionLogsFile.name))
+      .then(() => deletedTransactionLogFilesCount);
   }
 }
 
--- a/data-access-layer/iaas/BackupStore.js
+++ b/data-access-layer/iaas/BackupStore.js
@@ -32,9 +32,10 @@ class BackupStore {
     return _.nth(/^(.+)-broker$/.exec(this.containerName), 1);
   }
 
-  listFilenames(prefix, predicate, iteratees, dontParseFilename) {
+  listFilenames(prefix, predicate, iteratees, container, dontParseFilename) {
     const options = {
-      prefix: prefix
+      prefix: prefix,
+      container: container
     };
 
     const self = this;
@@ -276,20 +277,20 @@ class BackupStore {
       .then(isoDate => this.listBackupFiles(options, getPredicate(isoDate), iteratees));
   }
 
-  listTransactionLogsOlderThan(options, dateOlderThan) {
+  listTransactionLogsOlderThan(options, dateOlderThan, container) {
     const iteratees = ['lastModified'];
     let prefix = options.prefix;
 
     function getPredicate(isoDate) {
       return function predicate(filenameobject) {
         //transactionLogsDeletionStartDate defaults to current timestamp as part of isoDate function.
-        return _.lt(filenameobject.lastModified, isoDate);
+        return _.lt(new Date(filenameobject.lastModified).toISOString(), isoDate);
       };
     }
 
     return Promise
       .try(() => this.filename.isoDate(dateOlderThan))
-      .then(isoDate => this.listFilenames(prefix, getPredicate(isoDate), iteratees, true));
+      .then(isoDate => this.listFilenames(prefix, getPredicate(isoDate), iteratees, container, true));
   }
 }
 
--- a/data-access-layer/iaas/CloudProviderClient.js
+++ b/data-access-layer/iaas/CloudProviderClient.js
@@ -61,10 +61,10 @@ class CloudProviderClient extends BaseCloudClient {
   list(container, options) {
     if (arguments.length < 2) {
       options = container;
-      container = _.get(options, 'container') || this.containerName;
+      container = this.containerName;
     }
     return this.storage
-      .getFilesAsync(container, _.omit(options, 'container'))
+      .getFilesAsync(container, options)
       .then(listOfFiles => {
         let list = [];
         let isTruncated = false;
--- a/data-access-layer/iaas/CloudProviderClient.js
+++ b/data-access-layer/iaas/CloudProviderClient.js
@@ -61,10 +61,10 @@ class CloudProviderClient extends BaseCloudClient {
   list(container, options) {
     if (arguments.length < 2) {
       options = container;
-      container = this.containerName;
+      container = _.get(options, 'container') || this.containerName;
     }
     return this.storage
-      .getFilesAsync(container, options)
+      .getFilesAsync(container, _.omit(options, 'container'))
       .then(listOfFiles => {
         let list = [];
         let isTruncated = false;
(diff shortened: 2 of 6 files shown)


jobs/ScheduleBackupJob.js
```diff
@@ -99,11 +124,33 @@ class ScheduleBackupJob extends BaseJob {
             .deleteBackup(deleteOptions)
             .then(() => backup.backup_guid);
         }
-      }).then(deletedBackupGuids => {
-        logger.info(`Successfully deleted backup guids : ${deletedBackupGuids} - instance deleted : ${instanceDeleted}`);
+      })
+      .then(deletedBackupGuids => {
+        const serviceName = catalog.getService(job.attrs.data.service_id).name;
```
sauravmndl: Put a comment // Deleting transaction logs from service-container.

jobs/ScheduleBackupJob.js
```diff
@@ -80,10 +80,35 @@ class ScheduleBackupJob extends BaseJob {
...
+        }
+      }
+      return filteredOldBackups;
+    }
+
     const options = _.omit(job.attrs.data, 'trigger', 'type');
     return backupStore
       .listBackupsOlderThan(options, config.backup.retention_period_in_days)
-      .then(oldBackups => this.filterOldBackups(oldBackups))
+      .then(oldBackups => filterOldBackups(oldBackups))
```
sauravmndl: Put a comment //Deleting base backup/ backup guids.

jobs/ScheduleBackupJob.js
```diff
@@ -80,10 +80,35 @@ class ScheduleBackupJob extends BaseJob {
...
+      // is 15 days it would include all backups in (15th, 16th, 17th ...) */
+      if (typeof oldBackups !== 'undefined' && oldBackups.length > 0) {
+        // Older backups are sorted as latest at first
+        let sortedBackups = _.sortBy(oldBackups, ['started_at']).reverse();
+        const latestSuccessIndex = _.findIndex(sortedBackups,
+          backup => backup.state === CONST.OPERATION.SUCCEEDED);
+        if (latestSuccessIndex === -1) {
+          //No successful backup beyond retention period.
+          filteredOldBackups = sortedBackups;
+          transactionLogsDeleteBefore = new Date();
```
sauravmndl: //Not current date. Consider retention period days older date. config.backup.retention_period_in_days + 1 can be good enough.

test/test_broker/mocks/cloudProvider.js
```diff
@@ -124,4 +126,27 @@ function list(containerName, prefix, filenames, responseStatusCode, times) {
     .reply(responseStatusCode || 200, files, {
       'X-Container-Object-Count': '42'
     });
+}
+
+function listTransactionLogs(containerName, prefix, fileObjects) {
```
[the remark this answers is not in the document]
pritishmishra: Done!

data-access-layer/iaas/BackupStore.js
```diff
@@ -275,6 +276,22 @@ class BackupStore {
       .try(() => this.filename.isoDate(backupStartedBefore))
       .then(isoDate => this.listBackupFiles(options, getPredicate(isoDate), iteratees));
   }
+
+  listTransactionLogsOlderThan(options, dateOlderThan, container) {
```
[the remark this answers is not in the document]
pritishmishra: Done!

jobs/ScheduleBackupJob.js
pritishmishra: Done!

jobs/ScheduleBackupJob.js
pritishmishra: Done!

jobs/ScheduleBackupJob.js
pritishmishra: Done!

pritishmishra:
@sauravmndl Implemented Review comments. Please review again.

jobs/ScheduleBackupJob.js
```diff
@@ -99,22 +130,43 @@ class ScheduleBackupJob extends BaseJob {
             .deleteBackup(deleteOptions)
             .then(() => backup.backup_guid);
         }
-      }).then(deletedBackupGuids => {
-        logger.info(`Successfully deleted backup guids : ${deletedBackupGuids} - instance deleted : ${instanceDeleted}`);
+      })
+      .then(deletedBackupGuids => {
+        // Deleting transaction logs from service-container.
+        let deletedTransactionLogFilesCount = 0;
+        return backupStore.listFilesOlderThan(listOptions, transactionLogsDeleteBefore, container)
```
kamath-prasad: is this list function scalable for all IAASes ? Like for Azure we have limits of 5000 blobs , correct ?

data-access-layer/iaas/CloudProviderClient.js
```diff
@@ -61,10 +61,10 @@ class CloudProviderClient extends BaseCloudClient {
   list(container, options) {
     if (arguments.length < 2) {
       options = container;
-      container = this.containerName;
+      container = _.get(options, 'container') || this.containerName;
```
sauravmndl: May be not required this. call **list(container, options)** method with container [here](https://github.com/cloudfoundry-incubator/service-fabrik-broker/pull/355/files#diff-c79d1f6f368d710e95c0626775567787R50)

jobs/ScheduleBackupJob.js
```diff
@@ -80,11 +80,42 @@ class ScheduleBackupJob extends BaseJob {
...
+      // is 15 days it would include all backups in (15th, 16th, 17th ...) */
+      if (typeof oldBackups !== 'undefined' && oldBackups.length > 0) {
+        // Older backups are sorted as latest at first
+        let sortedBackups = _.sortBy(oldBackups, ['started_at']).reverse();
+        const latestSuccessIndex = _.findIndex(sortedBackups,
+          backup => backup.state === CONST.OPERATION.SUCCEEDED);
+        if (latestSuccessIndex === -1) {
+          //No successful backup beyond retention period.
+          filteredOldBackups = sortedBackups;
+          transactionLogsDeleteBefore = new Date(Date.now() - (config.backup.retention_period_in_days + 1) * 24 * 60 * 60 * 1000).toISOString();
```
sauravmndl: Also handle the case when instance deleted. In that case after retention period post instance deletion, all backups and transaction logs should be cleaned up.

data-access-layer/iaas/BackupStore.js
```diff
@@ -275,6 +276,22 @@ class BackupStore {
       .try(() => this.filename.isoDate(backupStartedBefore))
       .then(isoDate => this.listBackupFiles(options, getPredicate(isoDate), iteratees));
   }
+
+  listFilesOlderThan(options, dateOlderThan, container) {
+    const iteratees = ['lastModified'];
```
sauravmndl: Can we pass **iteratees**  as function, so that it becomes generic.

data-access-layer/iaas/BackupStore.js
```diff
@@ -275,6 +276,22 @@ class BackupStore {
...
+    let prefix = options.prefix;
```
sauravmndl: Let's harden as **const** instead of **let**

data-access-layer/iaas/BackupStore.js
```diff
@@ -275,6 +276,22 @@ class BackupStore {
...
+    function getPredicate(isoDate) {
+      return function predicate(filenameobject) {
+        //transactionLogsDeletionStartDate defaults to current timestamp as part of isoDate function.
+        return _.lt(new Date(filenameobject.lastModified).toISOString(), isoDate);
+      };
+    }
+
+    return Promise
+      .try(() => this.filename.isoDate(dateOlderThan))
+      .then(isoDate => this.listFilenames(prefix, getPredicate(isoDate), iteratees, container, true));
```
ketakig12: If you see, we have defined separate BackupStore for OOB and Service Instance where we do specific tasks like finding the prefix. I would recommend to have the same approach here (create a new BackupStore class for Service) and keep all iaas prefix related logic in iaas layer instead of caller. This whole function maybe then can be moved to separate BackupStore file, as I think it will be only used for transaction logs.

data-access-layer/iaas/CloudProviderClient.js
ketakig12: Yes, container is already passed as param, why do we need it in options again?

data-access-layer/iaas/CloudProviderClient.js
```diff
@@ -61,10 +61,10 @@ class CloudProviderClient extends BaseCloudClient {
...
     }
     return this.storage
-      .getFilesAsync(container, options)
+      .getFilesAsync(container, _.omit(options, 'container'))
```
ketakig12: Then I guess, this will also be removed if we are not passing container in options.

jobs/ScheduleBackupJob.js
```diff
@@ -80,11 +80,42 @@ class ScheduleBackupJob extends BaseJob {
...
+      }
+      return filteredOldBackups;
+    }
+
     const options = _.omit(job.attrs.data, 'trigger', 'type');
+    const serviceName = catalog.getService(job.attrs.data.service_id).name;
+    const listOptions = {
+      prefix: `${serviceName}/logs/${job.attrs.data.instance_id}`
+    };
+    const container = `${backupStore.containerPrefix}-${serviceName}`;
```
ketakig12: As commented above, both prefix and container formation can be / should be part of BackupStore.

jobs/ScheduleBackupJob.js
```diff
@@ -80,11 +80,42 @@ class ScheduleBackupJob extends BaseJob {
   }
 
   static deleteOldBackup(job, instanceDeleted) {
+    let transactionLogsDeleteBefore;
```
ketakig12: Maybe rename the variable as `transactionLogsBefore` as this is the time before which you get all logs.

jobs/ScheduleBackupJob.js
```diff
@@ -80,11 +80,42 @@ class ScheduleBackupJob extends BaseJob {
...
+    function filterOldBackups(oldBackups) {
+      let filteredOldBackups = [];
+      /* oldBackups : This aray should contain all older backups
+      // including last retenion day's backup. E.g. if retention period
+      // is 15 days it would include all backups in (15th, 16th, 17th ...) */
+      if (typeof oldBackups !== 'undefined' && oldBackups.length > 0) {
+        // Older backups are sorted as latest at first
+        let sortedBackups = _.sortBy(oldBackups, ['started_at']).reverse();
+        const latestSuccessIndex = _.findIndex(sortedBackups,
+          backup => backup.state === CONST.OPERATION.SUCCEEDED);
```
ketakig12: Should we also check trigger here? If the succeeded backup is of on-demand, user can delete it.

jobs/ScheduleBackupJob.js
```diff
@@ -99,22 +130,43 @@ class ScheduleBackupJob extends BaseJob {
...
-          .listBackupFilenames(backupStartedBefore, options)
-          .then(listOfFiles => {
-            if (listOfFiles.length === 0) {
+        transactionLogsDeleteBefore = new Date().toISOString();
+        return Promise.all([
+            backupStore.listBackupFilenames(backupStartedBefore, options),
+            backupStore.listFilesOlderThan(listOptions, transactionLogsDeleteBefore, container)
+          ])
+          .spread((listOfBackups, listOfTransactionLogs) => {
+            if (listOfBackups.length === 0 && listOfTransactionLogs.length === 0) {
```
ketakig12: We are **always** keeping 1 successful backup beyond retention period (even for deleted instance). So, how will this condition be satisfied?

jobs/ScheduleBackupJob.js
pritishmishra: Won't that be automatically taken care by above logic? Since, after the retention period, the backups will be deleted and upon not finding any backup, the above logic kicks in and deletes transaction logs older than 15 days.

data-access-layer/iaas/BackupStore.js
```diff
@@ -32,9 +32,10 @@ class BackupStore {
     return _.nth(/^(.+)-broker$/.exec(this.containerName), 1);
   }
 
-  listFilenames(prefix, predicate, iteratees) {
+  listFilenames(prefix, predicate, iteratees, container, dontParseFilename) {
```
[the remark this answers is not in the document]
pritishmishra: `parseFilename` reduces backward compatibility as now all other functions have to pass `true` and the new function will have to pass `false`. 
Here, using the opposite ensures that the original function signatures are left unchanged and only the new ones have to pass `true`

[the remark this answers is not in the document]
ketakig12: The containerName in this case is generic, I suppose --> it is used for service specific containers as in [here](https://github.com/cloudfoundry-incubator/service-fabrik-broker/blob/0c87df6471cfd9fd5be8063608b086acc6623a55/data-access-layer/iaas/BackupStore.js#L215). And as I have written in another comment below, I feel container name / prefix formation should be inside BackupStore.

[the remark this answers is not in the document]
ketakig12: parseFilename reduces backward compatibility --> Yes, I understand that, but is it not possible to keep default for it if not passed? dontParseFilename just does not look good as a var name. But I am fine if you decide not to change.

data-access-layer/iaas/CloudProviderClient.js
ketakig12: Not sure if I understand this, you can pass container as parameter instead if passing in options.

pritishmishra:
I had received quite a few review comments regarding creating `container` name inside the BackupStore file rather than passing the `container` param in function calls.

I had a discussion w/ @sauravmndl and we concluded that this could reduce the use-case of the `ListFilenames` function. This function was not only introduced for listing the Transaction Logs but also could be extended for any use-case which involves listing files in a container, without downloading the contents of the file.

So, restricting the creation of `container` name in the BackupStore file itself, will involve passing some extra param each time, we want to create `container` with different convention, which could become cumbersome. Hence, passing the `container` param as part of function call.

data-access-layer/iaas/CloudProviderClient.js
pritishmishra: Agreed. I think we can get rid of this logic and retrieve the `container` from the function param.
I have made the changes accordingly.

data-access-layer/iaas/CloudProviderClient.js
pritishmishra: Yes. Removed these changes.

data-access-layer/iaas/BackupStore.js
pritishmishra: The logic of introducing these functions is not limited to only Transaction Logs. I have explained this in a comment below.

jobs/ScheduleBackupJob.js
pritishmishra: Yes, however, I think the understanding is that we don't have an use-case right now, for exceeding the number of blobs above 5000, as per the estimated current-usage.

jobs/ScheduleBackupJob.js
pritishmishra: Done!

pritishmishra:
Regarding 
https://github.com/cloudfoundry-incubator/service-fabrik-broker/pull/355#pullrequestreview-144951868, yes, we need these changes for BackupReaperJob too, but I think, we have parked this for later.

Regarding 
https://github.com/cloudfoundry-incubator/service-fabrik-broker/pull/355#issuecomment-412449319, we had a discussion post my comment on this and after discussion with @kamath-prasad  and @sauravmndl , we have introduced transaction logs specific functions in BackupStore library code, and avoided the prefix/container creation call outside of BackupStore. Please revisit the code changes.

We haven’t however created a new BackupStore file specifically for Transaction Logs, as was one of the suggestions.

PR #358, opened 2018-08-08
Cross Instance Point In Time Recovery support

### LINKED ISSUE #357: Point In Time Recovery Cross Instance Restore
**Description**
Service Fabrik broker needs to support cross instance restore. Two scenarios need to cover

1. Source instance present in system
2. Source instance deleted

sauravmndl:
Fixes #357 and #360

### DIFF
--- a/api-controllers/FabrikBaseController.js
+++ b/api-controllers/FabrikBaseController.js
@@ -127,11 +127,6 @@ class FabrikBaseController extends BaseController {
   }
 
   validateRestoreQuota(options) {
-    const isPitrEnabled = this.getService(options.service_id).pitr;
-    if (!isPitrEnabled) {
-      logger.debug('Non pitr service', options);
-      return true;
-    }
     return this.backupStore
       .getRestoreFile(options)
       .then(metdata => {
--- a/api-controllers/FabrikBaseController.js
+++ b/api-controllers/FabrikBaseController.js
@@ -127,6 +127,12 @@ class FabrikBaseController extends BaseController {
   }
 
   validateRestoreQuota(options) {
+    const isPitrEnabled = this.getService(options.service_id).pitr;
+    if (!isPitrEnabled) {
+      // For non pitr supported services skipping quota check
+      logger.debug('Non pitr service', options);
+      return true;
+    }
     return this.backupStore
       .getRestoreFile(options)
       .then(metdata => {
@@ -217,6 +223,10 @@ class FabrikBaseController extends BaseController {
 
   createInstance(instance_id, service_id, plan_id, context) {
     return this.fabrik.createInstance(instance_id, service_id, plan_id, context);
+  }
+
+  getService(service_id) {
+    return catalog.getService(service_id);
   }
 
   getPlan(plan_id) {
--- a/api-controllers/FabrikBaseController.js
+++ b/api-controllers/FabrikBaseController.js
@@ -115,7 +115,7 @@ class FabrikBaseController extends BaseController {
 
   validateRestoreTimeStamp(epochDateString) {
     // Here validating 
-    // 1. Requested time stamp should be an epoch.
+    // 1. Requested time stamp should be epoch millisecond.
     // 2. Requested time should not be older than 14(retention period) days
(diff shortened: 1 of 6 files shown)


api-controllers/ServiceFabrikApiController.js
```diff
@@ -450,7 +450,7 @@ class ServiceFabrikApiController extends FabrikBaseController {
         }
```
ketakig12: What does it mean to send timestamp in a request for non-pitr service? Should that be also validated before proceeding?

api-controllers/ServiceFabrikApiController.js
```diff
@@ -450,7 +450,7 @@ class ServiceFabrikApiController extends FabrikBaseController {
...
       })
       .then(() => this.validateRestoreQuota({
```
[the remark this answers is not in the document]
ketakig12: Yes, I think should be validated for the target instance.

api-controllers/FabrikBaseController.js
```diff
@@ -127,6 +127,12 @@ class FabrikBaseController extends BaseController {
   }
 
   validateRestoreQuota(options) {
+    const isPitrEnabled = this.getService(options.service_id).pitr;
+    if (!isPitrEnabled) {
+      // For non pitr supported services skipping quota check
```
jagadish-kb: remove comments which are self-explanatory from looking at the code directly & straight forward.

api-controllers/ServiceFabrikApiController.js
```diff
@@ -482,7 +482,7 @@ class ServiceFabrikApiController extends FabrikBaseController {
           return this.backupStore.getBackupFile(backupFileOptions);
         }
       })
-      .catchThrow(NotFound, new UnprocessableEntity(`Cannot restore for guid/timeStamp '${timeStamp || backupGuid}' as no successful backup found in this space`))
+      .catchThrow(NotFound, new UnprocessableEntity(`Cannot restore for guid/timeStamp '${timeStamp || backupGuid}' as no successful backup found in this space.`))
```
jagadish-kb: comment is related to LN 490 - shouldn't we also pass source instance id and ensure they belong to the same plan as the instance id being restored or  check the restore predecessors & see if its allowed?

sauravmndl: The restore predecessor is check in LN 490. As the input is plan id of metadata of backup, and checked with current instance's plan or whether it is predecessor of current plan. So, I think by that it is already covered.

api-controllers/FabrikBaseController.js
sauravmndl: done

api-controllers/ServiceFabrikApiController.js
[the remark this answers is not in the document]
sauravmndl: Yes, it should be checked for current instance.

api-controllers/ServiceFabrikApiController.js
sauravmndl: Fabrik searches for a base backup before the requested timestamp. while triggering the backup via agent api backup_guid also sent in this case. Transaction log replaying happens in case time_stamp is also present. 

The check need not to be implemented at server end. If required can be introduced in client (cf cli/ UI based).
@jagadish-kb  @kamath-prasad  What is your opinion?

kamath-prasad: I think this validation is not necessary since anyways agent will not consume the timestamp even if passed.

api-controllers/FabrikBaseController.js
```diff
@@ -127,6 +127,11 @@ class FabrikBaseController extends BaseController {
   }
 
   validateRestoreQuota(options) {
+    const isPitrEnabled = this.getService(options.service_id).pitr;
```
abh1kg: Can we come up with a better option than "pitr"? Not sure if we can though :)

api-controllers/ServiceFabrikApiController.js
sauravmndl: added a check for non pitr

PR #363, opened 2018-08-14
Fix lockedResourceDetails check

amitmalav:
In 187b release, values of `CONST.APISERVER.RESOURCE_GROUPS` were changed. Hence there was some disparity in expected resourceGroup for lock resource, [before](https://github.com/cloudfoundry-incubator/service-fabrik-broker/blob/rel-2018.T07a/common/constants.js#L245) and [after](https://github.com/cloudfoundry-incubator/service-fabrik-broker/blob/rel-2018.T07b/common/constants.js#L246) 187b code.


amitmalav:
Tested change manually.
Created one lock resource with 7a code.
Deployed 7b code with above change.
Lock was released by poller.

PR #366, opened 2018-08-16
Add functions for stopping and starting BOSH deployments

abh1kg:
- This is required for the upcoming restore hooks flow where these functions would be invoked as part of the recovery process to speed up recovery time. As of now, these functions are not invoked. The start and stop functions start and stop all jobs of the provided BOSH deployment and returns a composite task ID for the operation. Callers must poll for the task operation status.

### DIFF
--- a/data-access-layer/bosh/BoshDirectorClient.js
+++ b/data-access-layer/bosh/BoshDirectorClient.js
@@ -771,6 +771,41 @@ class BoshDirectorClient extends HttpClient {
       );
   }
 
+  startDeployment(deploymentName) {
+    return Promise.try(() => {
+        return this
+          .getDirectorConfig(deploymentName);
+      })
+      .then(config => this.invokeDeploymentJobAction(config, deploymentName, CONST.BOSH_DEPLOYMENT_ACTIONS.STARTED));
+  }
+
+  stopDeployment(deploymentName) {
+    return Promise.try(() => {
+        return this
+          .getDirectorConfig(deploymentName);
+      })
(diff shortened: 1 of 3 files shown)


data-access-layer/bosh/BoshDirectorClient.js
```diff
@@ -771,6 +771,41 @@ class BoshDirectorClient extends HttpClient {
       );
   }
```
kamath-prasad: just a question , we dont wait for the task to complete ? Or it will be part of the script which invoke these functions ?

abh1kg: Yes--it would be the job of the client that calls these functions

subhankarc:
@abh1kg can you take care of @sauravmndl comments?

data-access-layer/bosh/BoshDirectorClient.js
```diff
@@ -771,6 +771,41 @@ class BoshDirectorClient extends HttpClient {
...
+      .makeRequestWithConfig({
+        method: 'PUT',
+        url: `/deployments/${deploymentName}/jobs/*`,
+        headers: {
+          'content-type': 'text/yaml'
+        },
+        qs: {
+          'state': expectedState
+        }
+      }, 302, directorConfig)
```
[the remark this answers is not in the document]
abh1kg: In the same file, other status codes are hardcoded. Will refrain from doing this to avoid discrepancy in code.

abh1kg:
Closing this PR since the branch is outdated. Created a new pull #392 
@sauravmndl , @subhankarc , @kamath-prasad

PR #368, opened 2018-08-20
Changing the constant for the apiserver sync resource state timeout

subhankarc:
* Changing the constant to make this function generic
* Increasing the value to 175, as default http timeout is 180. Kepping it little lower than that.

The same will be done as part of BOSH manager PR in master branch

### DIFF
--- a/common/constants.js
+++ b/common/constants.js
@@ -233,7 +233,7 @@ module.exports = Object.freeze({
     }
   },
   APISERVER: {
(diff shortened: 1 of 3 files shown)


subhankarc:
@kamath-prasad This needs to be sent as part of 8a. Can you please review and merge this?

subhankarc:
corresponding master PR https://github.com/cloudfoundry-incubator/service-fabrik-broker/pull/369

PR #369, opened 2018-08-20
Test fix on master and change of constant

subhankarc:
* Changing the constant to make this function generic
* Increasing the value to 175, as default http timeout is 180. Kepping it little lower than that.


subhankarc:
corresponding rel PR https://github.com/cloudfoundry-incubator/service-fabrik-broker/pull/368

PR #372, opened 2018-08-24
Removing request.instance references

subhankarc:
* Assign instance will ultimately be removed. 

* In this PR cleaning up the references of request.instance from middleware so that assigninstance can be later removed from the lifecycle route.

* Removing couple of stale comments


subhankarc:
@kamath-prasad 

Removed the unneeded methods. Coverage is now increased. Can you please merge.

PR #377, opened 2018-08-28
Handle Migration for Docker Bind

amitmalav:
Issue:
Before 2.0 code migarion => binding exists (b_id1)
After Migraion =>
In delete bind call
code is changed from
```javascript
In controller:
createBindResource (as it does not exist) with 
status = {
state: CONST.APISERVER.RESOURCE_STATE.DELETE
}
```
to
```javascript
createBindResource with 
status = {
state: CONST.APISERVER.RESOURCE_STATE.DELETE,
response: {}
}
```

### DIFF
--- a/api-controllers/ServiceBrokerApiController.js
+++ b/api-controllers/ServiceBrokerApiController.js
@@ -404,7 +404,8 @@ class ServiceBrokerApiController extends FabrikBaseController {
               },
               options: params,
               status: {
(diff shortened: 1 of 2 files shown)


kamath-prasad:
@amitmalav : this can be merged ?

subhankarc:
Yes it can be merged.

PR #381, opened 2018-08-30
Sent event logs

muralisc:
- Sent events logs to Reimann
- Clear the pollers only if the state is updated successfully
- Add catch block

### DIFF
--- a/managers/restore-manager/RestoreTaskPoller.js
+++ b/managers/restore-manager/RestoreTaskPoller.js
@@ -51,7 +51,7 @@ class RestoreTaskPoller {
               })
               .then(() => {
                 if (utils.isServiceFabrikOperationFinished(operationStatusResponse.state)) {
-                  _.set(restore_opts,'user.name', changedOptions.username);
+                  _.set(restore_opts, 'user.name', changedOptions.username);
                   return RestoreTaskPoller._logEvent(restore_opts, CONST.OPERATION_TYPE.RESTORE, operationStatusResponse);
                 }
               });
--- a/managers/restore-manager/RestoreTaskPoller.js
+++ b/managers/restore-manager/RestoreTaskPoller.js
@@ -47,10 +47,6 @@ class RestoreTaskPoller {
                   logger.debug('Clearing Restore Task Poller:', object.metadata.name);
                   clearInterval(interval);
                   RestoreTaskPoller.pollers[object.metadata.name] = false;
-                }
-              })
-              .then(() => {
-                if (utils.isServiceFabrikOperationFinished(operationStatusResponse.state)) {
                   _.set(restore_opts, 'user.name', changedOptions.username);
                   return RestoreTaskPoller._logEvent(restore_opts, CONST.OPERATION_TYPE.RESTORE, operationStatusResponse);
                 }
--- a/managers/restore-manager/RestoreTaskPoller.js
+++ b/managers/restore-manager/RestoreTaskPoller.js
@@ -2,6 +2,7 @@
 // TODO abstract to a base poller
 
 const _ = require('lodash');
+const Promise = require('bluebird');
 const eventmesh = require('../../data-access-layer/eventmesh');
 const CONST = require('../../common/constants');
 const logger = require('../../common/logger');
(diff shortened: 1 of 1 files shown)


managers/restore-manager/RestoreTaskPoller.js
```diff
@@ -29,30 +30,33 @@ class RestoreTaskPoller {
...
                 resourceId: changedOptions.restore_guid,
                 status: {
                   'state': operationStatusResponse.state,
                   response: operationStatusResponse
                 }
-              }));
+              }))
+              .then(() => {
+                if (_.includes([CONST.APISERVER.RESOURCE_STATE.SUCCEEDED, CONST.APISERVER.RESOURCE_STATE.ABORTED], operationStatusResponse.state)) {
+                  logger.debug('Cancel the poller:', object.metadata.name);
```
amitmalav: Please include `Clearing Restore Task Poller`, `Starting Restore Task Poller` etc in logger statements

managers/restore-manager/RestoreTaskPoller.js
```diff
@@ -29,30 +30,33 @@ class RestoreTaskPoller {
...
+                }
+              })
+              .then(() => {
+                if (utils.isServiceFabrikOperationFinished(operationStatusResponse.state)) {
+                  _.set(restore_opts, 'user.name', changedOptions.username);
+                  return RestoreTaskPoller._logEvent(restore_opts, CONST.OPERATION_TYPE.RESTORE, operationStatusResponse);
+                }
+              });
           })
+          .catch(e => logger.error('Caught error in poller', e))
```
amitmalav: Same here.

amitmalav: On second thought, we can cancel the poller here. and whenever watch refreshes we will anyway get the event again hence new poller will start. We will avoid running erroneous poll every 3 sec rather it will be done every 1 minute when watch refreshes

amitmalav: What happens if this error comes in [`_logEvent`](https://github.com/cloudfoundry-incubator/service-fabrik-broker/pull/381/files#diff-8c64b223c0a29cc710c7f01ce1d89d8dR55)
State is succeeded but poller won't be cancelled, makes sense to cancel the poller => next event won't come.

managers/restore-manager/RestoreTaskPoller.js
```diff
@@ -29,30 +30,33 @@ class RestoreTaskPoller {
...
                 resourceType: CONST.APISERVER.RESOURCE_TYPES.DEFAULT_RESTORE,
                 resourceId: changedOptions.restore_guid,
                 status: {
                   'state': operationStatusResponse.state,
                   response: operationStatusResponse
                 }
-              }));
+              }))
+              .then(() => {
+                if (_.includes([CONST.APISERVER.RESOURCE_STATE.SUCCEEDED, CONST.APISERVER.RESOURCE_STATE.ABORTED], operationStatusResponse.state)) {
```
amitmalav: Why not use `utils.isServiceFabrikOperationFinished(operationStatusResponse.state)`

amitmalav: What about `FAILED` state?

muralisc: Using `utils.isServiceFabrikOperationFinished(operationStatusResponse.state)`

managers/restore-manager/RestoreTaskPoller.js
muralisc: Updated as `Clearing Restore Task Poller`
`Starting poller for restore` is already at [line 66](
https://github.com/cloudfoundry-incubator/service-fabrik-broker/pull/381/files/c55a27eb80f29ae1b8ac009adf04785485c81e42#diff-8c64b223c0a29cc710c7f01ce1d89d8dR66)

managers/restore-manager/RestoreTaskPoller.js
muralisc: The poller is already cancelled at line 48

PR #383, opened 2018-08-31
Fix unbounded runs of update jobs even into next cycle

jagadish-kb:
If while a job is in flight and before it can commit its state to DB,
if the job crashes due to DB down or any other issue, then its next
run would not be null and agenda re-runs this job. Now this re-run
is a dupe of the same attempt which could create additional retry jobs.

In this fix, putting in a unique constraint on the jobdata which contains
the attempt count. So even if by chance a job reruns when it schedules
a retry job it will only overwrite existing retry job with the next attempt
count. Unique constraint on job is added.

Additionally have not kept retries to run infinitely. If the number of
attempts exceeds the number of attempts in a update window
(ex.reschedule delay of 30 mins -  48 runs per day * 7 = 336),
then next window run anyway runs. So will stop rescheduling in such
cases.


kamath-prasad:
Closing the PR as this fix is not required on 8B branch,

PR #384, opened 2018-08-31
Fix unbounded runs of update jobs even into next cycle

jagadish-kb:
If while a job is in flight and before it can commit its state to DB,
if the job crashes due to DB down or any other issue, then its next
run would not be null and agenda re-runs this job. Now this re-run
is a dupe of the same attempt which could create additional retry jobs.

In this fix, putting in a unique constraint on the jobdata which contains
the attempt count. So even if by chance a job reruns when it schedules
a retry job it will only overwrite existing retry job with the next attempt
count. Unique constraint on job is added.

Additionally have not kept retries to run infinitely. If the number of
attempts exceeds the number of attempts in a update window
(ex.reschedule delay of 30 mins -  48 runs per day * 7 = 336),
then next window run anyway runs. So will stop rescheduling in such
cases.

### DIFF
--- a/jobs/Scheduler.js
+++ b/jobs/Scheduler.js
@@ -253,11 +253,14 @@ class Scheduler {
         data = data ? _.omit(data, CONST.JOB_NAME_ATTRIB) : {};
         const uniqueCriteria = {};
         if (avoidDupJobWithSameData) {
-          _.each(data, (attrValue, attr) => uniqueCriteria[`data.${attr}`] = attrValue);
+          _.each(data, (attrValue, attr) => {
+            uniqueCriteria[`data.${attr}`] = attrValue;
+            return true;
+          });
         } else {
           uniqueCriteria['data._n_a_m_e_'] = jobName;
         }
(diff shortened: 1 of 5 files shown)


jobs/Scheduler.js
```diff
@@ -382,8 +387,7 @@ class Scheduler {
         .runAt(jobName,
           type,
           config.scheduler.jobs.reschedule_delay,
-          jobData,
-          CONST.SYSTEM_USER
+          jobData
```
amitmalav: Any particular reason why we are not passing `user` param here?

jagadish-kb: The method `runAt` never had user param. Was passing it wrongly here, so removed it.

subhankarc:
@jagadish-kb Shall I merge it?

jagadish-kb:
@subhankarc - yeah please do.
@sauravmndl @pritishmishra -- There was a fix pending from your end on the scheduled backup, please fix that as well on the similar lines.

PR #385, opened 2018-09-01
Stopping schedule update from going through old flow

subhankarc:
Schedule update also send the parameter `service-fabrik-operation` which was being checked in the patchInstance to support restore via old flow. For that, even schedule update was also going through old flow. This fix removes the bifurcation as restore manager is already in place and also fixes the schedule update problem.


subhankarc:
@abh1kg UTs which were pointing to old flow were removed earlier. Current UTs are covering the new flow. This was a temporary code till restoremanager in place which we forgot to remove. :(

PR #390, opened 2018-09-03
Removing assignInstance from V2 route

subhankarc:
As the factory pattern is no more used for lifecycle operations,

* Removed the assignInstance middleware from V2
* changed the corresponding code in the EventLogIntercepter and FabrikBaseController
* Added a test for Director Dashboard
* Removed unneeded code from `DirectorManager` and `DirectorInstance`.


api-controllers/routes/broker/v2.js
```diff
@@ -31,7 +31,7 @@ router.use(commonMiddleware.error({
 
 /* Service Instance Router */
 instanceRouter.use(controller.handler('ensurePlatformContext'));
-instanceRouter.use(controller.handler('assignInstance'));
+//instanceRouter.use(controller.handler('assignInstance'));
```
kamath-prasad: remove commented code.

broker/lib/fabrik/DirectorInstance.js
```diff
@@ -23,28 +23,6 @@ class DirectorInstance extends BaseInstance {
     this.networkSegmentIndex = undefined;
```
kamath-prasad: similar changes not required for docker - https://github.com/cloudfoundry-incubator/service-fabrik-broker/blob/master/broker/lib/fabrik/DockerInstance.js#L36-L58 ?

subhankarc: Oops! I made a mistake. Actually this cannot be removed. DashboardController uses it still!

subhankarc:
@kamath-prasad The factory pattern in the V2 route is removed with this. But it cannot be immediately removed as Dashboard and extension routes still uses them.

subhankarc:
@kamath-prasad There were no test for director dashboard access. Added a test which catches accidental removal of code like this.

kamath-prasad:
@subhankarc : ok , can you please update the tests ? coverage reduced ..

PR #393, opened 2018-09-07
Add HTTP Status Code constants

abh1kg:
- Migrate hardcoded numerals for status codes to constants from NPM library

### DIFF
--- a/package.json
+++ b/package.json
@@ -32,7 +32,10 @@
     "service-fabrik"
   ],
   "dependencies": {
+    "@google-cloud/compute": "0.9.1",
+    "@google-cloud/storage": "1.6.0",
     "agenda": "0.9.0",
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -45,11 +48,12 @@
     "express-interceptor": "1.2.0",
     "express-jwt": "5.0.0",
     "express-session": "1.13.0",
+    "http-status-codes": "^1.3.0",
```
sauravmndl: Just for understanding ? What if we want to support exact version ?

abh1kg: We can use the actual version as well.. I used `npm install --save` which inherently uses the carat-based nearest version. Also, I think the cloud delivery principles would prefer this. Would check this and update the pull if required

PR #395, opened 2018-09-10
[Fix] Event Logging not working for Async operation

subhankarc:
* Sending back operation for async operation
* EventLogInterceptor was not working for the same as it expects the operation as part of last operation.


amitmalav:
If possible, we can write integration test for async services as well, that will really help

PR #398, opened 2018-09-10
Fixing delete event logging for 410 Gone response

subhankarc:
* Asynchronous Delete response with 410 response were ignored and no event was logged for the same.
* Checking explicitly for event type as delete and statuscode as 410 and sending event info for the same.

### DIFF
--- a/api-controllers/ServiceBrokerApiController.js
+++ b/api-controllers/ServiceBrokerApiController.js
@@ -269,7 +269,7 @@ class ServiceBrokerApiController extends FabrikBaseController {
     }
 
     function notFound(err) {
-      if (!operation || (operation && operation.type === 'delete')) {
+      if (_.get(operation, 'type') === 'delete') {
         return gone();
       }
       failed(err);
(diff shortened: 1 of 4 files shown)


subhankarc:
@jagadish-kb Added the description. This PR is to handle the 410 scenario only.

https://github.com/cloudfoundry-incubator/service-fabrik-broker/pull/395 was created to solve the event logging issue for all the other async events.

api-controllers/ServiceBrokerApiController.js
```diff
@@ -265,11 +265,13 @@ class ServiceBrokerApiController extends FabrikBaseController {
     }
 
     function gone() {
-      res.status(CONST.HTTP_STATUS_CODE.GONE).send({});
+      res.status(CONST.HTTP_STATUS_CODE.GONE).send({
+        state: CONST.OPERATION.SUCCEEDED
```
amitmalav: According to [service broker api](https://github.com/openservicebrokerapi/servicebroker/blob/v2.12/spec.md#response-1), `GONE` should be sent with empty body. Any particular reason why we are sending body.

subhankarc: You are right. I missed this spec. Will change accordingly.

subhankarc:
@amitmalav @jagadish-kb Updated the PR based on comment from @amitmalav 

Please check.

PR #401, opened 2018-09-11
Added support for decoupling scheduled updated from cf

sauravmndl:
1. Decoupled update by calling broker api
2. Change in Agent bind/unbind apis
3. Introduced config driven (from service plan) lock operations


jobs/ServiceInstanceUpdateJob.js
```diff
@@ -72,9 +73,18 @@ class ServiceInstanceUpdateJob extends BaseJob {
 
   static updateInstanceIfOutdated(instanceDetails, planId, operationResponse) {
     const plan = catalog.getPlan(planId);
+    let context;
     logger.info(`Instance Id: ${instanceDetails.instance_id} - manager : ${plan.manager.name} - Force Update: ${plan.service.force_update}`);
-    return this
-      .getOutdatedDiff(instanceDetails.deployment_name, instanceDetails.instance_id, plan)
+    return cloudController.getOrgAndSpaceGuid(instanceDetails.instance_id)
```
subhankarc: why is this CF call needed? If we need to fetch the org and space, fetch it from apiserver resource.

jobs/ServiceInstanceUpdateJob.js
```diff
@@ -134,41 +146,49 @@ class ServiceInstanceUpdateJob extends BaseJob {
...
-            platform: CONST.PLATFORM.CF,
-            organization_guid: opts.organization_guid,
-            space_guid: opts.space_guid
-          };
-          opts.context = context;
-          return directorManager
-            .diffManifest(deploymentName, opts);
-        })
-      );
+      .then(directorManager => directorManager.diffManifest(deploymentName, tenantInfo));
```
subhankarc: Please use `DirectorService` instead. DirectorManager and usage of it is deprecated.

jobs/ServiceInstanceUpdateJob.js
```diff
@@ -134,41 +146,49 @@ class ServiceInstanceUpdateJob extends BaseJob {
       });
   }
 
-  static getOutdatedDiff(deploymentName, instanceId, plan) {
+  static deploymentStaggered(err) {
```
subhankarc: Why do we have to do this response description parsing logic? Does not look good.

jobs/ServiceInstanceUpdateJob.js
```diff
@@ -134,41 +146,49 @@ class ServiceInstanceUpdateJob extends BaseJob {
...
+    const response = _.get(err, 'error', {});
+    const description = _.get(response, 'description', '');
+    return description.indexOf(CONST.FABRIK_OPERATION_STAGGERED) > 0 && description.indexOf(CONST.FABRIK_OPERATION_COUNT_EXCEEDED) > 0;
+  }
+
+  static deploymentLocked(err) {
```
subhankarc: same as above.

jobs/ServiceInstanceUpdateJob.js
```diff
@@ -134,41 +146,49 @@ class ServiceInstanceUpdateJob extends BaseJob {
...
+          service_id: plan.service.id,
+          plan_id: plan.id,
+          previous_values: {
+            service_id: plan.service.id,
+            plan_id: plan.id,
+            organization_id: context.organization_id,
+            space_id: context.space_guid
+          },
+          parameters: {
+            scheduled: true,
```
subhankarc: We should pass and preserve the `service-fabrik-operation` parameter. `EventlogInterceptor` has dependency on the same.

managers/bosh-manager/DirectorService.js
```diff
@@ -778,7 +769,14 @@ class DirectorService extends BaseDirectorService {
     return Promise.join(
         this.executeActions(CONST.SERVICE_LIFE_CYCLE.PRE_BIND, actionContext),
         this.getDeploymentIps(deploymentName),
-        (preBindResponse, ips) => this.agent.createCredentials(ips, binding.parameters, preBindResponse)
+        (preBindResponse, ips) => utils.retry(() => this.agent.createCredentials(ips, binding.parameters, preBindResponse), {
+          maxAttempts: 3,
+          timeout: Math.floor(config.http_timeout * 2 / 3)
```
subhankarc: Ok so the timeout is configured here. The problem could be if the first IP is not reachable and taking time, always it would time out. I was thinking of timeout per request.

jobs/ServiceInstanceUpdateJob.js
sauravmndl: I was present in FabrikOperation class. I moved it here.

managers/bosh-manager/DirectorService.js
sauravmndl: Sure. Will do it.

api-controllers/FabrikBaseController.js
```diff
@@ -63,7 +63,7 @@ class FabrikBaseController extends BaseController {
             resourceId: req.params.instance_id,
             operation: operationType
           }
-        });
+        }, _.get(plan, 'supported_operations'));
```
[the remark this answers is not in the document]
sauravmndl: done. passing plan

common/utils/ServiceBrokerClient.js
```diff
@@ -72,6 +72,26 @@ class ServiceBrokerClient extends HttpClient {
       }, 200)
       .then(res => res.body);
   }
+
+  updateServiceInstance(options) {
+    logger.info(`-> Updating instance -  name: ${options.instance_id}`);
```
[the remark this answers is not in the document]
sauravmndl: done

data-access-layer/eventmesh/LockManager.js
```diff
@@ -94,14 +94,14 @@ class LockManager {
    * @param {string} [lockDetails.lockedResourceDetails.operation] - Operation type who is acquiring the lock. ex: backup
    */
 
-  lock(resourceId, lockDetails) {
+  lock(resourceId, lockDetails, supportedOperations) {
```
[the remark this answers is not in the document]
sauravmndl: done

data-access-layer/service-agent/Agent.js
```diff
@@ -204,19 +204,23 @@ class Agent extends HttpClient {
...
 
   deleteCredentials(ips, credentials, preUnbindResponse) {
     const body = {
       credentials: credentials,
       actions: preUnbindResponse
     };
-    return this
-      .getHost(ips, 'credentials')
-      .then(ip => this.post(ip, 'credentials/delete', body, 200));
+    return Promise
```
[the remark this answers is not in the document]
sauravmndl: have put timeout 30 sec for each operation.

jobs/ServiceInstanceUpdateJob.js
sauravmndl: done

data-access-layer/eventmesh/LockManager.js
```diff
@@ -214,8 +214,16 @@ class LockManager {
     });
   }
 
-  _getLockType(operation) {
-    if (_.includes(CONST.APISERVER.WRITE_OPERATIONS, operation)) {
+  _getLockType(requestedOperation, supportedOperations) {
```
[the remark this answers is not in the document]
sauravmndl: passing plan now

broker/config/settings.yml
```diff
@@ -703,6 +703,10 @@ defaults: &defaults
       id: 'bc158c9a-7934-401e-94ab-057082a5073f'
       name: 'v1.0-xsmall'
       free: false
+      supported_operations:
```
subhankarc: `async_ops_supporting_parallel_sync_ops` is what I could think of as a name.

Also, I think use of & and * is better so that u declare it once and use elsewhere.

subhankarc: Also, flat list instead of having read, as we discussed.

data-access-layer/service-agent/Agent.js
```diff
@@ -204,19 +204,37 @@ class Agent extends HttpClient {
...
     };
-    return this
-      .getHost(ips, 'credentials')
-      .then(ip => this.post(ip, 'credentials/create', body, 200));
+    return Promise
+      .any(_.map(ips, ip => utils.retry(() => this
+          .getHost([ip], 'credentials')
+          .then(ip => this.post(ip, 'credentials/create', body, 200)), {
+            maxAttempts: 2,
+            timeout: 30000
```
subhankarc: Put a constant may be!

jobs/ServiceInstanceUpdateJob.js
```diff
@@ -39,25 +40,23 @@ class ServiceInstanceUpdateJob extends BaseJob {
...
-        .getServicePlanIdForInstanceId(instanceDetails.instance_id)
-        .then((planId) => (planId === undefined ?
+      return eventmesh.apiServerClient.getResource({
+          resourceGroup: CONST.APISERVER.RESOURCE_GROUPS.DEPLOYMENT,
+          resourceType: CONST.APISERVER.RESOURCE_TYPES.DIRECTOR,
+          resourceId: instanceDetails.instance_id
+        })
+        .catch(errors.NotFound, () => undefined)
+        .then(resource => _.get(resource, 'spec.options'))
+        .then((resourceDetails) => (resourceDetails === undefined ?
```
subhankarc: Ideally all the resource would be created once a scheduled update is triggered on live, still deletion is little risky, can we log erro to start with?

managers/bosh-manager/DirectorService.js
```diff
@@ -232,22 +231,14 @@ class DirectorService extends BaseDirectorService {
...
-    }
     return this
       .initialize(operation)
-      .then(() => token ? jwt.verify(token, config.password) : null)
-      .then(serviceFabrikOperation => {
-        logger.info('SF Operation input:', serviceFabrikOperation);
-        this.operation = _.get(serviceFabrikOperation, 'name', 'update');
+      .then(() => {
+        logger.info('Parameters for update operation:', _.get(params, 'parameters'));
         // normal update operation
```
subhankarc: Please remove this comment now. I should have only removed it in my PR.

subhankarc:
@amitmalav please review once.

data-access-layer/service-agent/Agent.js
sauravmndl: done

broker/config/settings.yml
sauravmndl: done

jobs/ServiceInstanceUpdateJob.js
sauravmndl: as discussed leaving, as would be taken care by update.

managers/bosh-manager/DirectorService.js
sauravmndl: done

jobs/ServiceInstanceUpdateJob.js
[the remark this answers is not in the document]
sauravmndl: inherited from FabrikOperation

broker/config/settings.yml
```diff
@@ -703,6 +703,9 @@ defaults: &defaults
       id: 'bc158c9a-7934-401e-94ab-057082a5073f'
       name: 'v1.0-xsmall'
       free: false
+      async_ops_supporting_parallel_sync_ops: &async_ops_supporting_parallel_sync_ops
```
subhankarc: just add one liner with a description in case it is not understood.

data-access-layer/eventmesh/LockManager.js
```diff
@@ -94,14 +94,14 @@ class LockManager {
    * @param {string} [lockDetails.lockedResourceDetails.operation] - Operation type who is acquiring the lock. ex: backup
```
amitmalav: Can you please add `plan` also in jsdoc as optional param

data-access-layer/eventmesh/LockManager.js
```diff
@@ -214,8 +214,15 @@ class LockManager {
     });
   }
 
-  _getLockType(operation) {
-    if (_.includes(CONST.APISERVER.WRITE_OPERATIONS, operation)) {
+  _getLockType(requestedOperation, plan) {
+    const supportedOperations = _.get(plan, 'async_ops_supporting_parallel_sync_ops');
```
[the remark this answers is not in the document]
jagadish-kb: Rather i will check for `supported_operations` for the plan and if the `supported_operation` includes `bind_during_update` then in the method `middleware.checkBlockingOperationInProgress`, if i see if its write locked and the operation is update, then if the plan supports bind_during_update, and if the requested operation is bind, i allow it through.

api-controllers/FabrikBaseController.js
```diff
@@ -63,7 +63,7 @@ class FabrikBaseController extends BaseController {
             resourceId: req.params.instance_id,
             operation: operationType
           }
-        });
+        }, plan);
```
jagadish-kb: not related to this PR but in general for SF2.0, during Bind we dont acquire any READ locks. Shouldn't we be acquiring it? look here `middleware.checkBlockingOperationInProgress` - if we dont see a write lock we just proceed, but what if someone fires update while a bind is in progress? or say fires a delete ?

broker/config/settings.yml
jagadish-kb: +1. Too complicated name. Also I feel this should not be a generic name as above. Rather use `supported_features` and specifically call it `bind_during_update`. Also how do you know if the agent is in normal update Vs Major version upgrade? Can agents support bind during both scenarios or should we even distinguish between them? Somehow i am inclined towards leveraging `supported_features` as it fits better.

common/utils/ServiceBrokerClient.js
```diff
@@ -72,6 +73,31 @@ class ServiceBrokerClient extends HttpClient {
       }, 200)
       .then(res => res.body);
   }
+
+  updateServiceInstance(options) {
```
jagadish-kb: The intention of this PR was to decouple update from CF and if i understood it right allow for bind during update. If the /v2/service_instances patch api still returns back 202, then during normal updates bind wont go through because CF itself would error out. So the scope of this PR is to directly hit broker end point for update?

data-access-layer/service-agent/Agent.js
```diff
@@ -196,27 +197,45 @@ class Agent extends HttpClient {
...
   createCredentials(ips, parameters, preBindResponse) {
     const body = {
       parameters: parameters,
       actions: preBindResponse
     };
-    return this
-      .getHost(ips, 'credentials')
-      .then(ip => this.post(ip, 'credentials/create', body, 200));
+    return Promise
+      .any(_.map(ips, ip => utils.retry(() => this
```
jagadish-kb: Promise.any wrapping retry not a good combination. Since retry happens in background even though one of the promises are resolved retry on unsupported agent also will continue to happen.

Why not wrap the existing code in util.retry like the below?
```
 return utils.retry(() => { 
      return this
      .getHost(ips, 'credentials')
      .then(ip => this.post(ip, 'credentials/create', body, 200))},
            maxAttempts: 2,
            timeout: CONST.AGENT.OPERATION_TIMEOUT_IN_MILLIS,
            predicate: (err) => (err instanceof errors.Timeout)
          }); 
;
```
The above code does a retry of the entire stuff again in case of timeout error. In all other cases you should not retry. The below code will end up retrying even when there is legitigmate errors being thrown from agent.

data-access-layer/service-agent/Agent.js
```diff
@@ -196,27 +197,45 @@ class Agent extends HttpClient {
...
   deleteCredentials(ips, credentials, preUnbindResponse) {
     const body = {
       credentials: credentials,
       actions: preUnbindResponse
     };
-    return this
-      .getHost(ips, 'credentials')
-      .then(ip => this.post(ip, 'credentials/delete', body, 200));
+    return Promise
+      .any(_.map(ips, ip => utils.retry(() => this
```
jagadish-kb: Same comment as above.

jobs/ServiceInstanceUpdateJob.js
```diff
@@ -39,25 +40,23 @@ class ServiceInstanceUpdateJob extends BaseJob {
         logger.error(msg);
         return this.runFailed(new errors.ServiceUnavailable(msg), operationResponse, job, done);
       }
-      return this
-        .getServicePlanIdForInstanceId(instanceDetails.instance_id)
-        .then((planId) => (planId === undefined ?
+      return eventmesh.apiServerClient.getResource({
```
jagadish-kb: Resource group and types are all plan related. Instead wouldnt it be better if we hit the info end point of the service instance by servicefabrik api controller and consume that? Because the instance could also be K8S down the line and this code is typing up the code more to director. So i would recommend enhancing FabrikAPI to return instance info if required (if it is already not returning it). -- i.e. in case if agents are down of if agents cannot be down if the info endpoint is currently erroring, instead of erroinng they can include the state and return back other info from etcd as part of that api.

jobs/ServiceInstanceUpdateJob.js
```diff
@@ -86,34 +87,36 @@ class ServiceInstanceUpdateJob extends BaseJob {
...
-            if (err instanceof errors.DeploymentAttemptRejected) {
+            if (this.deploymentStaggered(err)) {
               //If deployment was staggered due to exhaustion of workers, reschedule update job
               //Retry attempts do not count when deployment is staggered
               //TODO: Need to check if the next run for scheduled update causes problems if the earlier deployment did not go through
               trackAttempts = false;
               err.statusMessage = 'Deployment attempt rejected due to BOSH overload. Update cannot be initiated';
             }
-            if (err instanceof errors.DeploymentAlreadyLocked) {
+            if (this.deploymentLocked(err)) {
```
jagadish-kb: can you also move the lines 119-122 into the method `deploymentLocked` and rename this method into `deploymentLockedForBackup`  and return the boolean accordingly?

managers/bosh-manager/DirectorService.js
```diff
@@ -232,22 +231,13 @@ class DirectorService extends BaseDirectorService {
...
     return this
       .initialize(operation)
-      .then(() => token ? jwt.verify(token, config.password) : null)
-      .then(serviceFabrikOperation => {
-        logger.info('SF Operation input:', serviceFabrikOperation);
-        this.operation = _.get(serviceFabrikOperation, 'name', 'update');
-        // normal update operation
+      .then(() => {
+        logger.info('Parameters for update operation:', _.get(params, 'parameters'));
+        this.operation = this.operation || 'update';
```
jagadish-kb: Is this line required?

managers/bosh-manager/DirectorService.js
```diff
@@ -232,22 +231,13 @@ class DirectorService extends BaseDirectorService {
...
         if (this.operation === 'update') {
```
jagadish-kb: Is this line required? now that you have removed all other code? So its always update right?

managers/bosh-manager/DirectorService.js
```diff
@@ -778,7 +768,13 @@ class DirectorService extends BaseDirectorService {
     return Promise.join(
         this.executeActions(CONST.SERVICE_LIFE_CYCLE.PRE_BIND, actionContext),
         this.getDeploymentIps(deploymentName),
-        (preBindResponse, ips) => this.agent.createCredentials(ips, binding.parameters, preBindResponse)
+        (preBindResponse, ips) => utils.retry(() => this.agent.createCredentials(ips, binding.parameters, preBindResponse), {
```
jagadish-kb: Agent.js does a retry and here you are retrying thrice? so in effect you have max 6 attempts at recreating creds. Can you add retry only at one place? Just having it here as well is fine.

managers/bosh-manager/DirectorService.js
```diff
@@ -778,7 +768,13 @@ class DirectorService extends BaseDirectorService {
...
+          maxAttempts: 3
+        })
+        .catch(errors.Timeout, err => {
```
jagadish-kb: why is this error logged specifically? This would have anyways got logged in LN790. Seems unnecessary.

managers/bosh-manager/DirectorService.js
```diff
@@ -816,7 +812,14 @@ class DirectorService extends BaseDirectorService {
           this.getDeploymentIps(deploymentName),
           this.getBindingProperty(deploymentName, id)
         ]))
-      .spread((preUnbindResponse, ips, binding) => this.agent.deleteCredentials(ips, binding.credentials, preUnbindResponse))
+      .spread((preUnbindResponse, ips, binding) => utils.retry(() => this.agent.deleteCredentials(ips, binding.credentials, preUnbindResponse), {
```
jagadish-kb: Same comment as above. Retry at multiple places unnecessary.

data-access-layer/eventmesh/LockManager.js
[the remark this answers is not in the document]
jagadish-kb: Also service instance update job fires update again as part of retry. So at any point in time two updates will happen from the job, so having the read lock would lead to pushing two updates downstream.

data-access-layer/eventmesh/LockManager.js
sauravmndl: done

broker/config/settings.yml
sauravmndl: put the comment

common/utils/ServiceBrokerClient.js
sauravmndl: only for schedule update.

data-access-layer/service-agent/Agent.js
sauravmndl: removed retry from here.

managers/bosh-manager/DirectorService.js
sauravmndl: yes. for directly invoking instance and call. Breaks UT otherwise.

jobs/ServiceInstanceUpdateJob.js
sauravmndl: This is the generic method for 'deploymentLocked'. for backup or not checked in caller

managers/bosh-manager/DirectorService.js
sauravmndl: removed

managers/bosh-manager/DirectorService.js
sauravmndl: removed from agent.

jobs/ServiceInstanceUpdateJob.js
jagadish-kb: Agree but is this deploymentLocked used anywhere else?  If this is common method, then move this to utils so that it can be reused and also move the LN119-120 into another util method `isDeploymentLockedFor(operationName, err)`
```
function isDeploymentLockedFor(operationName, err){
    if(isDeploymentLocked(err)){
          const errMessage = _.get(err, 'error.description') || '';
         const operation = errMessage.split(' ').splice(-1)[0];	              
          if (operation === operationName) {    
                  return true;
          }
       }
   return false;
}
```
Just recommending the above for better readability. 
Also better to name functions with boolean response with prefix `is` or `has`. In this case `is` suits better.

managers/bosh-manager/DirectorService.js
jagadish-kb: Ok.

managers/bosh-manager/DirectorService.js
```diff
@@ -778,7 +768,13 @@ class DirectorService extends BaseDirectorService {
     return Promise.join(
         this.executeActions(CONST.SERVICE_LIFE_CYCLE.PRE_BIND, actionContext),
         this.getDeploymentIps(deploymentName),
-        (preBindResponse, ips) => this.agent.createCredentials(ips, binding.parameters, preBindResponse)
+        (preBindResponse, ips) => utils.retry(() => this.agent.createCredentials(ips, binding.parameters, preBindResponse), {
+          maxAttempts: 3,
+          timeout: CONST.AGENT.OPERATION_TIMEOUT_IN_MILLIS
```
jagadish-kb: Do you want to add a predicate here to retry to only retry in case of TimeOut? You seem to be retrying even for other generic errors which Agent might be throwing.

sauravmndl: Can be better to retry as scenarios might be getinfo successful but createcreds not. Or there might be other unknown scenarios. Also, the timeout happens in 35 sec much lower than cf timeout. Is that fine? In all other retries in code base we don't check only for timeout. So, shall we keep the same for this as well ?

api-controllers/FabrikBaseController.js
sauravmndl: As discussed 'async_ops_supporting_parallel_sync_ops' parameter in plan having the info.

data-access-layer/eventmesh/LockManager.js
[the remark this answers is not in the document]
sauravmndl: As discussed 'async_ops_supporting_parallel_sync_ops' parameter in plan having the info.

data-access-layer/service-agent/Agent.js
sauravmndl: removed retry from here.

jobs/ServiceInstanceUpdateJob.js
sauravmndl: As no provisioner is present apart from bosh director, hence hard coded. Ideal would be to save plan id in job data. In plan the resource group and type are present.
So, have put comment
```
      /* In case of instance provisioned by other than bosh director.
       * Need to chnage resource group / type. Better to save plan name in job data.
       * Otherwise need to look for instance_id with all possible provisioners.
       */
```

jobs/ServiceInstanceUpdateJob.js
sauravmndl: moved to utils.

managers/bosh-manager/DirectorService.js
sauravmndl: removed

api-controllers/FabrikBaseController.js
kamath-prasad: Maybe I didn't get this correctly , someone firing update or delete while bind is in progress cannot happen as CF would not allow it , correct ? Because for user triggered update/delete , the request will come via CF ?!

I see that now since we have decoupled outdated instance update from CF , there is a chance that while bind is in progress an outdated instance update can go through.

Is this understanding correct ?

jagadish-kb: @kamath-prasad - Your understanding is correct. In summary this is what we are doing:
1. For SYNCH operations no locks are acquired by SF.
2. Only for ASYCNH operations we acquire lock and we flag that lock as READ/WRITE. If a LOCK is flagged as READ, then it would allow for parallel SYNCH requests to go through.

So we still rely on the platform for locking SYNCH operations as you have rightly pointed out and this is sth if K8s does not provide similar to CF, then we must revisit this model and have our locking completely independent of platform. Hope this clarifies any further doubts that you had.

data-access-layer/eventmesh/LockManager.js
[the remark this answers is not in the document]
jagadish-kb: Ok.

managers/bosh-manager/DirectorService.js
jagadish-kb: You are retrying on the method 'createCredentials' - this method internally does two things - get ip (via /info) & create creds. Now if either of these fails, the method 'createCredentials' fails. Also when you are retrying in scenarios like above (especially creating sth), its better to retry on known scenarios. If unknown scenarios occur, which also qualifies for retry then they must be added in subsequently as known scenarios. Should not retry for anything and everything.

jagadish-kb: With regards to this timeout value, better get this clarified once from service team if this SLA is acceptable to services teams. Also i thought you were a big proponent of configs over constant :-), if this is sth thats configurable wouldnt you want it in your configs ?

api-controllers/FabrikBaseController.js
subhankarc: Adding to what @jagadish-kb said, with scheduled update not going through CF, even if it happens that update comes while bind in progress and update brings the VM down, the retry would ensure that the credential is created.

managers/bosh-manager/DirectorService.js
subhankarc: +1 for keeping it a config with a default in the spec.

sauravmndl: sure, will put in config :)

jobs/ServiceInstanceUpdateJob.js
```diff
@@ -86,34 +87,36 @@ class ServiceInstanceUpdateJob extends BaseJob {
...
               //If deployment locked then backup is in progress. So reschedule update job,
               //Retry attempts dont count when deployment is locked for backup.
               // Only if locked for backup operation
               // Message example: 
               // Service Instance abcdefgh-abcd-abcd-abcd-abcdefghijkl __Locked__ at Mon Sep 10 2018 11:17:01 GMT+0000 (UTC) for backup
-              const errMessage = err.message || '';
+              const errMessage = _.get(err, 'error.description') || '';
```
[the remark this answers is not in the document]
sauravmndl: Introduction of error parsing is wrong approach itself. As it is a legacy code, had to do this way.

managers/bosh-manager/DirectorService.js
subhankarc: @jagadish-kb @sauravmndl  I feel retrying for all error scenarios is not going to create any issue, if not make it better. Finding the set of errors for which we need to retry is tricky, as errors like "credentials cannot be created as cluster topology changed in between" may also be types of error for which retry is necessary. I do not see any error for which retry should not be done, do you see any? As it is only be retried for finite amount of time, I guess it is better to keep the retry for all errors. Let me know what you think.

sauravmndl: @jagadish-kb As discussed keep as it is. Better to find out all possible errors and do retry only for those errors. So, not retrying only for timeout error. Anyway it will timeout in 35 secs (configurable now). In future if required we will find out list of all possible errors and do retry for only those for all agent operations.

jobs/ServiceInstanceUpdateJob.js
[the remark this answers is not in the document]
sauravmndl: done

PR #404, opened 2018-09-16
[DRAFT] Make bind/unbind/backup operations resilient to BOSH unavailability

abh1kg:
- Cache all static IP addresses of all managed BOSH deployments by inspecting the deployment manifest
- Use the deployment IP cache for all operations instead of requesting BOSH
- Use concurrency limits for parallel loading of the manifests [to prevent overloading BOSH]
- Load IPs if initial cache is unavailable [Cache Busting]


abh1kg:
Will add test cases in some time.. 
@subhankarc , @jagadish-kb : can you do a cursory review for this ?

PR #408, opened 2018-09-18
Dashboard rendering: Fetching current plan of service instance while validation

jayeshjk:
* Currently, while dashboard rendering, we fetch current plan of the service instance from CC in the getInfo() call. This fix was introduced in an earlier PR https://github.com/cloudfoundry-incubator/service-fabrik-broker/pull/277 to handle the case where instance plan was updated. (dashboard URL remains same after plan update)
* Before call to getInfo(), validation process happens in which plan id present in dashboard url is validated against plan ids in service catalog. This validation will therefore fail if plan id in url happens to be so outdated that it got removed from catalog.
* This PR addresses this issue by making call to CC to fetch current plan in the validation itself instead of later point of time, i.e., in getInfo() call.

### DIFF
--- a/test/test_broker/mocks/cloudController.js
+++ b/test/test_broker/mocks/cloudController.js
@@ -311,7 +311,7 @@ function getServiceInstancesInSpaceWithName(instance_name, space_guid, present) 
     });
 }
 
-function findServicePlanByInstanceId(instance_id, plan_guid, plan_unique_id, resources, times = 1) {
+function findServicePlanByInstanceId(instance_id, plan_guid, plan_unique_id, resources, times) {
   const defaultResources = [{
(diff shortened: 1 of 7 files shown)


test/test_broker/mocks/cloudController.js
```diff
@@ -325,6 +325,7 @@ function findServicePlanByInstanceId(instance_id, plan_guid, plan_unique_id, res
     .query({
       q: `service_instance_guid:${instance_id}`
     })
+    .times(times)
```
subhankarc: do it like this instead `.times(times || 1)`

jayeshjk: okay .. will update.

PR #415, opened 2018-09-25
Added support to decouple cf update in admin update api

sauravmndl:
1. Decoupled cf update
2. For instance space guid, org guid calling api server instead of cf

### DIFF
--- a/api-controllers/ServiceFabrikAdminController.js
+++ b/api-controllers/ServiceFabrikAdminController.js
@@ -17,6 +17,9 @@ const ScheduleManager = require('../jobs');
 const BackupReportManager = require('../reports');
 const CONST = require('../common/constants');
 const maintenanceManager = require('../maintenance').maintenanceManager;
+const serviceBrokerClient = require('../common/utils/ServiceBrokerClient');
+const eventmesh = require('../data-access-layer/eventmesh');
+const DirectorService = require('../managers/bosh-manager/DirectorService');
 const Conflict = errors.Conflict;
 const Forbidden = errors.Forbidden;
 const BadRequest = errors.BadRequest;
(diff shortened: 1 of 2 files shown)


api-controllers/ServiceFabrikAdminController.js
```diff
@@ -52,19 +68,41 @@ class ServiceFabrikAdminController extends FabrikBaseController {
...
+          resourceGroup: CONST.APISERVER.RESOURCE_GROUPS.DEPLOYMENT,
+          resourceType: CONST.APISERVER.RESOURCE_TYPES.DIRECTOR,
+          resourceId: instanceId
+        })
+        .catch(errors.NotFound, () => undefined)
+        .then(resource => _.get(resource, 'spec.options'))
+        .then(resource => {
+          resourceDetails = resource;
+          if (resourceDetails === undefined) {
+            throw new errors.NotFound(`Resource details of service instance ${instanceId} not found in api server.`);
```
subhankarc: Can we have this method segregated to find Plan_ID and organization guid and space guid.

api-controllers/ServiceFabrikAdminController.js
```diff
@@ -52,19 +68,41 @@ class ServiceFabrikAdminController extends FabrikBaseController {
...
+          } else {
+            const planId = _.get(resourceDetails, 'plan_id');
+            plan = catalog.getPlan(planId);
+            context = _.get(resourceDetails, 'context');
+            if (allowForbiddenManifestChanges === false) {
+              const tenantInfo = _.pick(resourceDetails, ['context', 'organization_guid', 'space_guid']);
```
subhankarc: getting only context is not enough? Do we need to get organization_guid and space_guid also?

sauravmndl: Legacy code :) Looking at DirectorService.generateManifest() seems not required.

subhankarc: lets remove then

api-controllers/ServiceFabrikAdminController.js
sauravmndl: Well, can be. How to return? as composite obj ? Then use like obj.plan, obj.organization_guid, obj.space_guid?
Might be required to fetch some other parameters. So as *resourceDetails* having all, ad-hoc facing would be good? or should I change ?

subhankarc: recource details should be fine.

api-controllers/ServiceFabrikAdminController.js
subhankarc: Checked the code, should be removed. As you said, it's a legacy code.

sauravmndl: done

Issue #414, opened 2018-09-25
APIServerClient must implement exponential backoff when trying to reconnect


amitmalav:
[Source Ref](https://github.com/cloudfoundry-incubator/service-fabrik-broker/blob/master/data-access-layer/eventmesh/ApiServerClient.js#L164)
```javascript
     stream.on('error', err => {
          logger.error(`Error occured during watching for resource ${resourceGroup}, ${resourceType}`, err);
          this.registerWatcher(resourceGroup, resourceType, callback, queryString);
          //throw err;
        });
```

In case of an error, retry is fired immediately and in case etcd happens to be down then this retry goes on continuously. General best practice is to have an exponential backoff when doing retries to underlying systems.

Take a look at reconnect implementation of DBManager, which uses utils.retry which has exponential backoff implemented internally.

PFB excerpt of the log from one of the operators & you can see even within the same millisecond several retries happening.:
```
2018-09-21T02:38:57.200Z - error: Error occured during watching for resource deployment.servicefabrik.io, directors Error: connect EHOSTUNREACH 10.244.14.252:8443 - Local (192.168.50.1:63821)
    at Object.exports._errnoException (util.js:1018:11)
    at exports._exceptionWithHostPort (util.js:1041:20)
    at connect (net.js:880:16)
    at net.js:969:9
    at _combinedTickCallback (internal/process/next_tick.js:73:7)
    at process._tickDomainCallback (internal/process/next_tick.js:128:9)
2018-09-21T02:38:57.200Z - error: Error occured during watching for resource bind.servicefabrik.io, directorbinds Error: connect EHOSTUNREACH 10.244.14.252:8443 - Local (192.168.50.1:63822)
    at Object.exports._errnoException (util.js:1018:11)
    at exports._exceptionWithHostPort (util.js:1041:20)
    at connect (net.js:880:16)
    at net.js:969:9
    at _combinedTickCallback (internal/process/next_tick.js:73:7)
    at process._tickDomainCallback (internal/process/next_tick.js:128:9)
2018-09-21T02:38:57.200Z - error: Error occured during watching for resource deployment.servicefabrik.io, directors Error: connect EHOSTUNREACH 10.244.14.252:8443 - Local (192.168.50.1:63823)
    at Object.exports._errnoException (util.js:1018:11)
    at exports._exceptionWithHostPort (util.js:1041:20)
    at connect (net.js:880:16)
    at net.js:969:9
[...]

PR #416, opened 2018-09-25
Remove error event on stream

amitmalav:
Workaround for #414 
As we are already refreshing watch after some time again and again, we don't really need `stream.on(error)`.
But if some fine day we are able to avoid refreshing watch or decide to increase refresh logic.
`error` event on stream will have to be handled intelligently.

### DIFF
--- a/data-access-layer/eventmesh/ApiServerClient.js
+++ b/data-access-layer/eventmesh/ApiServerClient.js
@@ -159,12 +159,6 @@ class ApiServerClient {
               timeoutSeconds: CONST.APISERVER.WATCH_TIMEOUT
             }
           });
-        // Removed this piece of code as watch is being refreshed every WATCHER_REFRESH_INTERVAL: 60000 (1 minute)
-        // stream.on('error', err => {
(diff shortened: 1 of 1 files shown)


data-access-layer/eventmesh/ApiServerClient.js
```diff
@@ -159,11 +159,12 @@ class ApiServerClient {
               timeoutSeconds: CONST.APISERVER.WATCH_TIMEOUT
             }
           });
-        stream.on('error', err => {
-          logger.error(`Error occured during watching for resource ${resourceGroup}, ${resourceType}`, err);
-          this.registerWatcher(resourceGroup, resourceType, callback, queryString);
-          //throw err;
-        });
+        // Removed this piece of code as watch is being refreshed every WATCHER_REFRESH_INTERVAL: 60000 (1 minute)
```
kamath-prasad: pls remove code all-together , no commented code pls.

amitmalav: Done!

PR #423, opened 2018-10-04
Add response in case of error in backup trigger

### LINKED ISSUE #422: Backup State should show some response in case of error
Currently if some error happens in backup trigger, response in apiserver defaultbackup resource is empty.
```
 "status": {
        "error": "{\"message\":\"openstack Error (413): Over Limit\"}",
        "lastOperation": "{}",
        "response": "{}",
        "state": "failed"
    }
```
Same error message can be put in response field also
```
"status": {
        "error": "{\"message\":\"openstack Error (413): Over Limit\"}",
        "lastOperation": "{}",
        "response": "{\"message\":\"openstack Error (413): Over Limit\"}",
        "state": "failed"
    }
```

amitmalav:
Fixes #422

### DIFF
--- a/managers/backup-manager/BackupService.js
+++ b/managers/backup-manager/BackupService.js
@@ -132,6 +132,10 @@ class BackupService extends BaseDirectorService {
             resourceId: result.backup_guid,
             status: {
               state: CONST.APISERVER.RESOURCE_STATE.FAILED,
+              response: {
+                state: CONST.APISERVER.RESOURCE_STATE.FAILED,
+                description: err.message
+              },
               error: utils.buildErrorJson(err)
             }
           }))


managers/backup-manager/BackupService.js
```diff
@@ -132,6 +132,7 @@ class BackupService extends BaseDirectorService {
             resourceId: result.backup_guid,
             status: {
               state: CONST.APISERVER.RESOURCE_STATE.FAILED,
+              response: utils.buildErrorJson(err),
```
subhankarc: Will this have state attribute as part of response?

amitmalav: Now added state explicitly.

PR #426, opened 2018-10-08
Add Base Poller

amitmalav:
Currently there are 3 different apiserver based pollers which have similar working logic.
With this change there will be one base poller which can be extended by all apiserver based pollers.

`BaseStatusPoller.js`
What it does??
> It also has erstwhile boilerplate code which handles polling logic in case of HA (Only one process should poll at a time)

Actual `getStatus` method is to be implemented by classes extending BasePoller mentioned next.
`BoshTaskStatusPoller`
`BackupStatusPoller`
`RestoreStatusPoller`


amitmalav:
Unit tests are WIP

managers/restore-manager/RestoreStatusPoller.js
```diff
@@ -0,0 +1,76 @@
...
+const utils = require('../../common/utils');
+const EventLogInterceptor = require('../../common/EventLogInterceptor');
+const BaseStatusPoller = require('../BaseStatusPoller');
+
+class RestoreStatusPoller extends BaseStatusPoller {
+  constructor() {
+    super({
+      resourceGroup: CONST.APISERVER.RESOURCE_GROUPS.RESTORE,
+      resourceType: CONST.APISERVER.RESOURCE_TYPES.DEFAULT_RESTORE,
+      validStateList: [CONST.APISERVER.RESOURCE_STATE.IN_PROGRESS],
```
kamath-prasad: just like backup , shouldn't we check for aborting state here as well ?

amitmalav: For only `in_progress` state we are starting pollers. @muralisc can you please confirm

amitmalav:
@jagadish-kb As discussed let's have one `BaseStatusPoller` class which can extend `BaseManager`.
This change we can send in another PR.

managers/BaseStatusPoller.js
```diff
@@ -0,0 +1,134 @@
...
+    const resourceDetails = eventmesh.apiServerClient.parseResourceDetailsFromSelfLink(object.metadata.selfLink);
+    // If no lockedByPoller annotation then set annotation  with time
+    // Else check timestamp if more than specific time than start polling and change lockedByPoller Ip
+    return eventmesh.apiServerClient.getResource({
+        resourceGroup: resourceDetails.resourceGroup,
+        resourceType: resourceDetails.resourceType,
+        resourceId: object.metadata.name,
+      })
+      .then(resourceBody => {
+        const options = _.get(resourceBody, 'spec.options');
```
muralisc: Can we refactor this into a separate function so the code flow is clear.
I guess the code flow in this function is similar:
```
pollStatus()
      getResourceDeatails() // already present
      getLock()
      getStatus()  // already present
      stopIfDone() 
```

managers/restore-manager/RestoreStatusPoller.js
muralisc: We need to add polling support for restore abort similar to backup.

managers/BaseStatusPoller.js
```diff
@@ -0,0 +1,134 @@
...
+  init() {
+    const queryString = `state in (${_.join(this.validStateList, ',')})`;
+    return eventmesh.apiServerClient.registerWatcher(this.resourceGroup, this.resourceType, this.startPoller.bind(this), queryString)
+      .then(stream => {
+        logger.debug(`Successfully set watcher on ${this.resourceType} resource types for status polling with query string:`, queryString);
+        return Promise
+          .delay(CONST.APISERVER.POLLER_WATCHER_REFRESH_INTERVAL)
+          .then(() => {
+            logger.debug(`Refreshing stream after ${CONST.APISERVER.POLLER_WATCHER_REFRESH_INTERVAL}`);
+            stream.abort();
```
sauravmndl: why stream abort required? just for understanding.

amitmalav: We have to close the old stream (if it hasn't died) in order to avoid multiple connections

managers/BaseStatusPoller.js
amitmalav: Done!

PR #427, opened 2018-10-08
Event logging for sf internal update operations

sauravmndl:
Earlier sf triggered update used to happen via cloud controller. Hence event logging for update used to happen via last operation api. Now broker triggered updates (schedule/admin update) happens via calling broker's api directly. so, there is no one call last operation api. Internally apiserver entity is updated with operation status for that instance e.g. update is in_progress, succeeded etc. So event logging is now done when for update and parameter 'service-fabrik-operation' is present then only log event.

### DIFF
--- a/managers/bosh-manager/BoshTaskPoller.js
+++ b/managers/bosh-manager/BoshTaskPoller.js
@@ -132,7 +132,9 @@ class BoshTaskPoller {
                   if (_.get(resourceBody.status.response, 'type') === CONST.OPERATION_TYPE.UPDATE &&
                     _.get(resourceBody.status.response, 'parameters.service-fabrik-operation') === true &&
                     _.includes([CONST.APISERVER.RESOURCE_STATE.FAILED, CONST.APISERVER.RESOURCE_STATE.FAILED], lastOperationOfInstance.state)) {
-                    return _logEvent(options, lastOperationOfInstance, CONST.HTTP_METHOD.PATCH);
+                    return _logEvent(_.assign(options, {
+                      instance_id: metadata.name
+                    }), lastOperationOfInstance, CONST.HTTP_METHOD.PATCH);
                   }
                 });
             }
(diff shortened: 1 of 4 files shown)


managers/bosh-manager/BoshTaskPoller.js
```diff
@@ -119,6 +127,14 @@ class BoshTaskPoller {
                 })
                 .catch(Conflict, () => {
                   logger.debug(`Not able to acquire bosh task poller processing lock for instance ${object.metadata.name}, Request is probably picked by other worker`);
+                })
+                .finally(() => {
+                  if (_.get(resourceBody.status.response, 'type') === 'update' &&
```
kamath-prasad: pls use constants for `update` .

managers/bosh-manager/BoshTaskPoller.js
```diff
@@ -29,6 +30,10 @@ class BoshTaskPoller {
           const options = _.get(resourceBody, 'spec.options');
           const pollerAnnotation = _.get(resourceBody, 'metadata.annotations.lockedByTaskPoller');
           logger.debug(`pollerAnnotation is ${pollerAnnotation} current time is: ${new Date()}`);
+          let lastOperationOfInstance = {
+            state: 'in progress',
```
kamath-prasad: isn't there a constant for `in progress` ?

jagadish-kb:
Also, please follow PR guidelines and do not leave it blank.  Title for this PR is misleading, does not provide right information for other reviewers who dont have this context.

managers/bosh-manager/BoshTaskPoller.js
sauravmndl: Actually it is state of osb api. So state should be 'in progress'.

kamath-prasad: ok , fine.

managers/bosh-manager/BoshTaskPoller.js
sauravmndl: done

PR #428, opened 2018-10-08
Modifying flow of getDeploymentIps to fetch IPs from ApiServer

jayeshjk:
* To introduce resiliency in bind/unbind during bosh outage, we store deployment IPs (which are needed during bind/unbind) in etcd. (Separate PR for this coming once https://github.com/cloudfoundry-incubator/service-fabrik-broker/pull/426 merges).

* This PR adds modifications in getDeploymentIps flow to utilize entries in etcd. Now the flow in getDeploymentIps will be local_cache --> etcd --> bosh.

* Failures while getting/updating resource from/to etcd can be ignored to not alter the flow of parent operation as IPs could still be obtained from bosh.

### DIFF
--- a/data-access-layer/bosh/BoshDirectorClient.js
+++ b/data-access-layer/bosh/BoshDirectorClient.js
@@ -567,7 +567,7 @@ class BoshDirectorClient extends HttpClient {
     logger.info(`[getDeploymentIps] making request to ApiServer for deployment - ${deploymentName}`);
     let resourceId = utils.parseServiceInstanceIdFromDeployment(deploymentName);
     if (deploymentName === resourceId) {
-      //don't contact to ApiServer if not a service-fabrik deployment
+      //don't contact to ApiServer if it is out-of-band deployment
       return Promise.resolve();
     }
     return eventmesh.apiServerClient.getResource({
@@ -585,7 +585,7 @@ class BoshDirectorClient extends HttpClient {
   putDeploymentIpsInResource(deploymentName, ips) {
     let resourceId = utils.parseServiceInstanceIdFromDeployment(deploymentName);
     if (deploymentName === resourceId) {
-      //don't contact to ApiServer if not a service-fabrik deployment
+      //don't contact to ApiServer if it is out-of-band deployment
       return Promise.resolve();
     }
     return eventmesh.apiServerClient.updateResource({
--- a/data-access-layer/bosh/BoshDirectorClient.js
+++ b/data-access-layer/bosh/BoshDirectorClient.js
@@ -563,20 +563,12 @@ class BoshDirectorClient extends HttpClient {
     });
   }
 
-  parseServiceInstanceIdFromDeployment(deploymentName) {
-    const deploymentNameArray = utils.deploymentNameRegExp().exec(deploymentName);
-    if (Array.isArray(deploymentNameArray) && deploymentNameArray.length === 4) {
-      return deploymentNameArray[3];
-    }
-    return deploymentName;
-  }
-
   getDeploymentIpsFromResource(deploymentName) {
     logger.info(`[getDeploymentIps] making request to ApiServer for deployment - ${deploymentName}`);
-    let resourceId = this.parseServiceInstanceIdFromDeployment(deploymentName);
+    let resourceId = utils.parseServiceInstanceIdFromDeployment(deploymentName);
     if (deploymentName === resourceId) {
       //don't contact to ApiServer if not a service-fabrik deployment
-      return Promise.try(() => {});
+      return Promise.resolve();
     }
     return eventmesh.apiServerClient.getResource({
         resourceGroup: CONST.APISERVER.RESOURCE_GROUPS.DEPLOYMENT,
@@ -586,15 +578,15 @@ class BoshDirectorClient extends HttpClient {
       .then(resource => JSON.parse(_.get(resource, 'metadata.annotations.deploymentIps', '{}')))
       .catch(err => {
         logger.error(`[getDeploymentIps] Error occurred while getting deployment Ips for ${deploymentName} from ApiServer`, err);
-        return {};
+        return;
       });
   }
 
   putDeploymentIpsInResource(deploymentName, ips) {
-    let resourceId = this.parseServiceInstanceIdFromDeployment(deploymentName);
+    let resourceId = utils.parseServiceInstanceIdFromDeployment(deploymentName);
     if (deploymentName === resourceId) {
       //don't contact to ApiServer if not a service-fabrik deployment
-      return Promise.try(() => {});
+      return Promise.resolve();
     }
     return eventmesh.apiServerClient.updateResource({
         resourceGroup: CONST.APISERVER.RESOURCE_GROUPS.DEPLOYMENT,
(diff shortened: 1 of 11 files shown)


jayeshjk:
Added "DO_NOT_MERGE" because PR for adding deployment IPs in deployment resource while create/update is not raised. Its unsafe to include flow without that information.

data-access-layer/bosh/BoshDirectorClient.js
```diff
@@ -542,16 +543,84 @@ class BoshDirectorClient extends HttpClient {
...
+        logger.error(`[getDeploymentIps] Error occurred while getting deployment Ips for ${deploymentName} from ApiServer`, err);
+        return {};
+      });
+  }
+
+  putDeploymentIpsInResource(deploymentName, ips) {
+    let resourceId = this.parseServiceInstanceIdFromDeployment(deploymentName);
+    if (deploymentName === resourceId) {
+      //don't contact to ApiServer if not a service-fabrik deployment
+      return Promise.try(() => {});
```
kamath-prasad: generally what is a good practise - `Promise.resolve(true);` or `Promise.try(() => {});` ?

data-access-layer/bosh/BoshDirectorClient.js
```diff
@@ -542,16 +543,84 @@ class BoshDirectorClient extends HttpClient {
...
+                  return this.putDeploymentIpsInResource(deploymentName, ips)
+                    .then(() => ips);
+                });
+            }
           });
       }
     });
+  }
+
+  parseServiceInstanceIdFromDeployment(deploymentName) {
```
kamath-prasad: This function is present in [ServiceFabrikAdminController](https://github.com/cloudfoundry-incubator/service-fabrik-broker/blob/master/api-controllers/ServiceFabrikAdminController.js#L143-L149) as well , better to keep it in utils ?

data-access-layer/bosh/BoshDirectorClient.js
```diff
@@ -542,16 +543,84 @@ class BoshDirectorClient extends HttpClient {
...
+    }
+    return deploymentName;
+  }
+
+  getDeploymentIpsFromResource(deploymentName) {
+    logger.info(`[getDeploymentIps] making request to ApiServer for deployment - ${deploymentName}`);
+    let resourceId = this.parseServiceInstanceIdFromDeployment(deploymentName);
+    if (deploymentName === resourceId) {
+      //don't contact to ApiServer if not a service-fabrik deployment
+      return Promise.try(() => {});
```
kamath-prasad: same question as below.

data-access-layer/bosh/BoshDirectorClient.js
jayeshjk: Will move this function to utils. Initially I had more specific code under this, so made a copy here.

data-access-layer/bosh/BoshDirectorClient.js
```diff
@@ -542,16 +543,76 @@ class BoshDirectorClient extends HttpClient {
...
       }
     });
+  }
 
+  getDeploymentIpsFromResource(deploymentName) {
+    logger.info(`[getDeploymentIps] making request to ApiServer for deployment - ${deploymentName}`);
+    let resourceId = utils.parseServiceInstanceIdFromDeployment(deploymentName);
+    if (deploymentName === resourceId) {
+      //don't contact to ApiServer if not a service-fabrik deployment
+      return Promise.resolve();
```
jagadish-kb: shouldnt this instead throw an error? Here you are supressing it as it might lead to some other failure in caller. Better to throw exception.

data-access-layer/bosh/BoshDirectorClient.js
```diff
@@ -542,16 +543,76 @@ class BoshDirectorClient extends HttpClient {
...
+        logger.error(`[getDeploymentIps] Error occurred while getting deployment Ips for ${deploymentName} from ApiServer`, err);
+        return;
+      });
+  }
+
+  putDeploymentIpsInResource(deploymentName, ips) {
+    let resourceId = utils.parseServiceInstanceIdFromDeployment(deploymentName);
+    if (deploymentName === resourceId) {
+      //don't contact to ApiServer if not a service-fabrik deployment
+      return Promise.resolve();
```
jagadish-kb: same as above with regards to throwing error.

data-access-layer/bosh/BoshDirectorClient.js
```diff
@@ -542,16 +543,76 @@ class BoshDirectorClient extends HttpClient {
...
+    }
+    return eventmesh.apiServerClient.updateResource({
+        resourceGroup: CONST.APISERVER.RESOURCE_GROUPS.DEPLOYMENT,
+        resourceType: CONST.APISERVER.RESOURCE_TYPES.DIRECTOR,
+        resourceId: resourceId,
+        metadata: {
+          annotations: {
```
jagadish-kb: metadata even has labels. can you verify if this was overwriting the entire metadata or does it patch metadata and other objects like labels, name, etc.. remain intact? Hope this is tested :-)

data-access-layer/bosh/BoshDirectorClient.js
```diff
@@ -542,16 +543,76 @@ class BoshDirectorClient extends HttpClient {
...
+          }
+        }
+      })
+      .catch(err => {
+        logger.error(`[getDeploymentIps] Error occured while updating resource for ${deploymentName} on ApiServer`, err);
+        return;
+      });
+  }
+
+  getDeploymentIpsFromDirector(deploymentName) {
```
jagadish-kb: why is this wrapper method required? why not directly invoke existing method `getDeploymentInstances`?

data-access-layer/bosh/BoshDirectorClient.js
jagadish-kb: Just change the comment to include OOB deployment scenario.

data-access-layer/bosh/BoshDirectorClient.js
jagadish-kb: Ok.

data-access-layer/bosh/BoshDirectorClient.js
jagadish-kb: Got it. Thanks.!

data-access-layer/bosh/BoshDirectorClient.js
```diff
@@ -542,16 +543,76 @@ class BoshDirectorClient extends HttpClient {
...
+    if (deploymentName === resourceId) {
+      //don't contact to ApiServer if not a service-fabrik deployment
+      return Promise.resolve();
+    }
+    return eventmesh.apiServerClient.getResource({
+        resourceGroup: CONST.APISERVER.RESOURCE_GROUPS.DEPLOYMENT,
+        resourceType: CONST.APISERVER.RESOURCE_TYPES.DIRECTOR,
+        resourceId: resourceId
+      })
+      .then(resource => JSON.parse(_.get(resource, 'metadata.annotations.deploymentIps', '{}')))
```
amitmalav: Should the default value be empty object or empty list?
Wouldn't it be better to send `undefined` as we are doing in error case?

jayeshjk: JSON.parse gives error with 'undefined', that's why I had to include empty object in the first place. I agree that it looks a tad bit ugly, should we do some refactoring?

PR #429, opened 2018-10-09
Use weekly cron for scheduled update

muralisc:
Add function `getRandomCronForOnceEveryXDaysWeekly` to get weekly cron.


muralisc:
@kamath-prasad As suggested i have added comments

PR #432, opened 2018-10-10
Adding Indexes to JobDetails/JobRunDetails schema MongoDB

mayanktiwari09:
added to field: createdAt


kamath-prasad:
LGTM ... As discussed , please check if the indexes created this way get used by the queries in [backup report generation](https://github.com/cloudfoundry-incubator/service-fabrik-broker/blob/master/reports/BackupReportManager.js#L76-L79).

PR #434, opened 2018-10-11
Modifying BoshTaskStatusPoller to update deployment IPs in deployment resources

jayeshjk:
* To introduce bind unbind resiliency towards bosh outages, corresponding flows have been altered to fetch deployment IPs from etcd first. ( See this PR https://github.com/cloudfoundry-incubator/service-fabrik-broker/pull/428)
* In this PR we modify BoshTaskStatusPoller to update deployment IPs information on the deployment resource at the time of deployment creation and update to ensure consistency.

### DIFF
--- a/managers/bosh-manager/BoshTaskStatusPoller.js
+++ b/managers/bosh-manager/BoshTaskStatusPoller.js
@@ -38,37 +38,7 @@ class BoshTaskStatusPoller extends BaseStatusPoller {
         .tap(lastOperationValue => logger.debug('last operation value is ', lastOperationValue))
         .tap(lastOperationValue => lastOperationOfInstance = lastOperationValue)
         .then(lastOperationValue => Promise.all([
-          Promise.try(() => {
-            if (lastOperationValue.resourceState === CONST.APISERVER.RESOURCE_STATE.SUCCEEDED &&
-              (lastOperationValue.type === 'create' || lastOperationValue.type === 'update')) {
-              return directorService.director.getDeploymentNameForInstanceId(directorService.guid)
-                .then(deploymentName => directorService.director.getDeploymentIpsFromDirector(deploymentName))
-                .then(ips => eventmesh.apiServerClient.updateResource({
-                  resourceGroup: CONST.APISERVER.RESOURCE_GROUPS.DEPLOYMENT,
-                  resourceType: CONST.APISERVER.RESOURCE_TYPES.DIRECTOR,
-                  resourceId: instanceId,
-                  status: {
-                    lastOperation: lastOperationValue,
-                    state: lastOperationValue.resourceState
-                  },
-                  metadata: {
-                    annotations: {
-                      deploymentIps: JSON.stringify(ips)
-                    }
-                  }
-                }));
-            } else {
-              return eventmesh.apiServerClient.updateResource({
-                resourceGroup: CONST.APISERVER.RESOURCE_GROUPS.DEPLOYMENT,
-                resourceType: CONST.APISERVER.RESOURCE_TYPES.DIRECTOR,
-                resourceId: instanceId,
-                status: {
-                  lastOperation: lastOperationValue,
-                  state: lastOperationValue.resourceState
-                }
-              });
-            }
-          }),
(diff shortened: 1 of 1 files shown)


managers/bosh-manager/BoshTaskStatusPoller.js
```diff
@@ -34,33 +34,77 @@ class BoshTaskStatusPoller extends BaseStatusPoller {
...
-      }), Promise.try(() => {
-        if (_.includes([CONST.APISERVER.RESOURCE_STATE.SUCCEEDED, CONST.APISERVER.RESOURCE_STATE.FAILED], lastOperationValue.resourceState)) {
-          // cancel the poller and clear the array
+      .then(directorService => directorService.lastOperation(_.get(resourceBody, 'status.response'))
+        .tap(lastOperationValue => logger.debug('last operation value is ', lastOperationValue))
+        .tap(lastOperationValue => lastOperationOfInstance = lastOperationValue)
+        .then(lastOperationValue => Promise.all([
+          Promise.try(() => {
+            if (lastOperationValue.resourceState === CONST.APISERVER.RESOURCE_STATE.SUCCEEDED &&
+              (lastOperationValue.type === 'create' || lastOperationValue.type === 'update')) {
```
kamath-prasad: can you add `create` and `update` to constants ? I guess @sauravmndl  has already added update to constants.

managers/bosh-manager/BoshTaskStatusPoller.js
```diff
@@ -34,33 +34,77 @@ class BoshTaskStatusPoller extends BaseStatusPoller {
...
+              return directorService.director.getDeploymentNameForInstanceId(directorService.guid)
+                .then(deploymentName => directorService.director.getDeploymentIpsFromDirector(deploymentName))
```
kamath-prasad: question : what would happen if BOSH director is unavailable during this call ? Shouldn't the exception be caught here and we proceed with rest of the flow ?

managers/bosh-manager/BoshTaskStatusPoller.js
```diff
@@ -75,26 +119,8 @@ class BoshTaskStatusPoller extends BaseStatusPoller {
...
-          status: {
-            lastOperation: lastOperationOfInstance,
-            state: CONST.APISERVER.RESOURCE_STATE.FAILED,
-            error: utils.buildErrorJson(err)
-          }
-        });
-      })
+        })
+      )
       .finally(() => {
```
amitmalav: `finally` block can also be put in same promise chain of `directorService`

managers/bosh-manager/BoshTaskStatusPoller.js
```diff
@@ -34,33 +34,77 @@ class BoshTaskStatusPoller extends BaseStatusPoller {
...
-          state: lastOperationValue.resourceState
-        }
-      }), Promise.try(() => {
-        if (_.includes([CONST.APISERVER.RESOURCE_STATE.SUCCEEDED, CONST.APISERVER.RESOURCE_STATE.FAILED], lastOperationValue.resourceState)) {
-          // cancel the poller and clear the array
+      .then(directorService => directorService.lastOperation(_.get(resourceBody, 'status.response'))
+        .tap(lastOperationValue => logger.debug('last operation value is ', lastOperationValue))
+        .tap(lastOperationValue => lastOperationOfInstance = lastOperationValue)
+        .then(lastOperationValue => Promise.all([
+          Promise.try(() => {
```
sauravmndl: As this part became bigger, can be put inside a function.

managers/bosh-manager/BoshTaskStatusPoller.js
amitmalav: If BOSH director is unavailable `lastOperation` itself would fail.
But event then I think we should catch the exception and then go through with operation.

managers/bosh-manager/BoshTaskStatusPoller.js
jayeshjk: By previous code flow, `finally` block will still be executed if `directorService.lastOperation` fails. If `finally` is moved within same promise chain, it's execution will depend `directorService.lastOperation` being successfully resolved. @sauravmndl could you pls confirm this?

PR #437, opened 2018-10-18
Removing createBindingProperty on Bosh from bind flow and deleteBindingProperty from unbind flow

jayeshjk:
* Now that bind credentials are being stored on etcd we don't need to create binding property on bosh, so that operation is removed from bind flow.

* Also, operation deleteBindingProperty in unbind flow is also made conditional based on whether bind resource was found in etcd or not. In other words, we don't need to delete binding properties on bosh for new binds as they won't be created there in the first place.

* Modified UTs to accommodate new bind/unbind flow.

### DIFF
--- a/operators/bosh-operator/DirectorService.js
+++ b/operators/bosh-operator/DirectorService.js
@@ -834,8 +834,7 @@ class DirectorService extends BaseDirectorService {
         }
         logger.info(`[getCredentials] Fetching property from bosh for binding ${id}`);
         return this.getBindingProperty(deploymentName, id)
-          .then(binding => binding.credentials)
-          .tap(() => this.deleteBindingProperty(deploymentName, id));
+          .then(binding => binding.credentials);
       });
   }
 
@@ -863,11 +862,6 @@ class DirectorService extends BaseDirectorService {
       .getDeploymentProperty(deploymentName, `binding-${id}`)
       .then(result => JSON.parse(result))
       .catchThrow(NotFound, new ServiceBindingNotFound(id));
-  }
-
-  deleteBindingProperty(deploymentName, id) {
-    return this.director
-      .deleteDeploymentProperty(deploymentName, `binding-${id}`);
   }
 
   diffManifest(deploymentName, opts) {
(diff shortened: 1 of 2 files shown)


operators/bosh-operator/DirectorService.js
```diff
@@ -836,7 +834,8 @@ class DirectorService extends BaseDirectorService {
         }
         logger.info(`[getCredentials] Fetching property from bosh for binding ${id}`);
         return this.getBindingProperty(deploymentName, id)
-          .then(binding => binding.credentials);
+          .then(binding => binding.credentials)
+          .tap(() => this.deleteBindingProperty(deploymentName, id));
```
amitmalav: This is done to clear all old bindings from bosh? Any particular reason why it was moved here from `deleteBinding ` method?

operators/bosh-operator/DirectorService.js
```diff
@@ -849,7 +848,7 @@ class DirectorService extends BaseDirectorService {
       })
       .then(resource => {
         let response = _.get(resource, 'status.response', undefined);
```
amitmalav: No need to set default as `undefined` as resolved value itself will be `undefined` in that case

operators/bosh-operator/DirectorService.js
```diff
@@ -849,7 +848,7 @@ class DirectorService extends BaseDirectorService {
...
-        if (response) {
+        if (!_.isEmpty(response)) {
```
amitmalav: Just curious, why did we add this `_.isEmpty` here. In what situations we will have empty `response` object?

operators/bosh-operator/DirectorService.js
jayeshjk: This was moved here to avoid Bosh call in normal flow. Also as we won't be creating new binding properties, deleteBindingProperty should not happen by default.

amitmalav:
On a second though, I'd suggest we remove `deleteBindingProperty` call to bosh all together, as it seems inconsistent with the flow that we are not creating property on bosh but deleting it. It would also invite some error scenarios in corner cases.

jayeshjk:
@amitmalav  @kamath-prasad  Based on discussions, it seems that removing deleteBindingProperty call all together seems safer. Will make the appropriate changes.

operators/bosh-operator/DirectorService.js
subhankarc: I agree. Let us not have the deleteBindingProperty call at all.

PR #424, opened 2018-10-05
Remove assignManager

muralisc:
Remove `assignManager` call from `api-controllers/routes/api/v1.js`
The functionality is now available in `setPlan` in `api-controllers/ServiceFabrikApiController.js`
Remove `req.manager.isAutoUpdatePossible`

PR #444, opened 2018-10-23
Removing ExecuteActions method from DirectorManager

subhankarc:
* ExecuteActions method is no more used in DirectorManager, removing it.
* Removed the corresponding test-cases.


subhankarc:
closing this. added this as part of https://github.com/cloudfoundry-incubator/service-fabrik-broker/pull/424

PR #446, opened 2018-10-24
Checkmarx Fix : Missing HSTS Header for dashboard controller

### LINKED ISSUE #445: Missing HSTS Header for dashboard controller
### Missing HSTS Header
**Risk**
Failure to set an HSTS header and provide it with a reasonable "max-age" value of at least one year may leave users vulnerable to Man-in-the-Middle attacks.
**Cause**
Many users browse to websites by simply typing the domain name into the address bar, without the protocol prefix. The browser will automatically assume that the user's intended protocol is HTTP, instead of
the encrypted HTTPS protocol.
When this initial request is made, an attacker can perform a Man-in-the-Middle attack and manipulate it to redirect users to a malicious web-site of the attacker's choosing. To protect the user from such an occurence, the HTTP Strict Transport Security (HSTS) header instructs the user's browser to disallow use of an unsecure HTTP connection to the the domain associated with the HSTS header.
Once a browser that supports the HSTS feature has visited a web-site and the header was set, it will no longer allow communicating with the domain over an HTTP connection.

Once an HSTS header was issued for a specific website, the browser is also instructed to prevent users from manually overriding and accepting an untrusted SSL certificate for as long as the "max-age" value still applies. The recommended "max-age" value is for at least one year in seconds, or 31536000.
**General Recommendations**

- Before setting the HSTS header - consider the implications it may have:
  - Forcing HTTPS will prevent any future use of HTTP, which could hinder some testing
  - Disabling HSTS is not trivial, as once it is disabled on the site, it must also be disabled on the
browser
- Set the HSTS header either explicitly within application code, or using web-server configurations.
- Ensure the "max-age" value for HSTS headers is set to 31536000 to ensure HSTS is strictly enforced
for at least one year.
- Once HSTS has been enforced, submit the web-application's address to an HSTS preload list - this will ensure that, even if a client is accessing the web-application for the first time (implying HSTS has not yet been set by the web-application), a browser that respects the HSTS preload list would still treat the web-application as if it had already issued an HSTS header. Note that this requires the server to have a trusted SSL certificate, and issue an HSTS header with a maxAge of 1 year (31536000)
- Note that this query is designed to return one result per application. This means that if more than one vulnerable response without an HSTS header is identified, only the first identified instance of this issue will be highlighted as a result. Since HSTS is required to be enforced across the entire application to be considered a secure deployment of HSTS functionality, fixing this issue only where the query highlights this result is likely to produce subsequent results in other sections of the application; therefore, when adding this header via code, ensure it is uniformly deployed across the entire application. If this header is added via configuration, ensure that this configuration applies to the entire application.

**Source Code Examples**
**JavaScript**
**Using Helmet with Express**
```
var express = require('express')
var helmet = require('helmet') // Helmet includes HSTS as a built-in header
var app = express() app.use(helmet())
```
**Using Explicit HSTS Package (Built into Helmet, Both Are Not Required)**
```
var hsts = require('hsts')
app.use(hsts({
 maxAge: 31536000
}))
```
**Explicitly Setting HSTS Header in Code**
`res.setHeader("Strict-Transport-Security", "max-age=31536000");`

subhankarc:
Closes #445

### DIFF
--- a/common/ExpressApp.js
+++ b/common/ExpressApp.js
@@ -9,6 +9,7 @@ const morgan = require('morgan');
 const bodyParser = require('body-parser');
 const config = require('./config');
 const logger = require('./logger');
+const helmet = require('helmet'); // Helmet includes HSTS as a built-in header
 const middleware = require('./middleware');
 const connectTimeout = require('connect-timeout');
 
@@ -29,6 +30,7 @@ class ExpressApp {
     app.set('title', cfg.title || `Service Fabrik Broker (${type})`);
     if (cfg.ssl) {
       app.set('ssl', cfg.ssl);
+      app.use(helmet());
     }
     app.disable('etag');
     app.set('views', path.join(__dirname, 'views'));
--- a/package.json
+++ b/package.json
@@ -45,6 +45,7 @@
     "express-interceptor": "1.2.0",
     "express-jwt": "5.0.0",
     "express-session": "1.13.0",
+    "helmet": "3.14.0",
     "hystrixjs": "0.2.0",
     "ip": "1.1.3",
     "js-yaml": "3.6.1",


common/ExpressApp.js
```diff
@@ -29,6 +30,7 @@ class ExpressApp {
     app.set('title', cfg.title || `Service Fabrik Broker (${type})`);
     if (cfg.ssl) {
       app.set('ssl', cfg.ssl);
+      app.use(helmet());
```
abh1kg: @subhankarc ,  @kamath-prasad : I looked at this late. AFAIK, adding the helmet middleware is not enough to prevent HSTS attacks. One needs to also add code like the following as a peer dependency to `helmet` using the child `hsts` module:

```
app.use(helmet.hsts({
  maxAge: <time for expiry>
}))
```

Of course, this is from experience but from past experience with the `helmet` module, the idea was always to keep the module as a core and add dependencies as needed. Otherwise, the `helmet` module itself would become really heavyweight

abh1kg: To add to this, from previous experience, adding the helmet middleware by default can also lead to side effects because a bunch of headers and processing happens on each request. This has led to issues like `max http header size exceeded` in some cases- just food for thought

subhankarc: @abh1kg As far as I read, helmet by default adds maxAge as 1 year so there is no need to explicitly set it.

subhankarc: By default, it adds the following

```
X-DNS-Prefetch-Control: off
X-Frame-Options: SAMEORIGIN
Strict-Transport-Security: max-age=15552000; includeSubDomains
X-Download-Options: noopen
X-Content-Type-Options: nosniff
X-XSS-Protection: 1; mode=block
```

PR #449, opened 2018-10-24
Move platform managers from broker/lib/fabrik to platform-managers

amitmalav:


kamath-prasad:
@amitmalav : is this mergable ?

amitmalav:
yes. It can be merged

PR #450, opened 2018-10-24
Added code for bosh resilience in create, update and delete flow.

sauravmndl:
This is to address bosh resilience. In create and update flow if bosh is not reachable during acquire or find network segment index respectively, the deployments should be staggered.  The staggered deployments would tried out later in BoshStaggeredDeploymentPoller. The same thing is done for delete deployment as well.

### DIFF
--- a/common/constants.js
+++ b/common/constants.js
@@ -385,5 +385,6 @@ module.exports = Object.freeze({
   },
   DOCKER_HOST_CONFIG: {
     PIDS_LIMIT: 150
-  }
+  },
+  SYSTEM_ERRORS: ['ECONNREFUSED', 'ECONNRESET', 'EPIPE', 'ETIMEDOUT', 'ESOCKETTIMEDOUT']
 });
--- a/operators/bosh-operator/BoshOperator.js
+++ b/operators/bosh-operator/BoshOperator.js
@@ -109,7 +109,7 @@ class BoshOperator extends BaseOperator {
         resourceId: _.get(changeObjectBody, 'metadata.name'),
         status: {
           response: response,
-          state: CONST.APISERVER.RESOURCE_STATE.IN_PROGRESS
+          state: _.get(response, 'task_id') ? CONST.APISERVER.RESOURCE_STATE.IN_PROGRESS : CONST.APISERVER.RESOURCE_STATE.WAITING
         }
       }))
       .catch(ServiceInstanceNotFound, () => eventmesh.apiServerClient.deleteResource({
--- a/operators/bosh-operator/BoshStaggeredDeploymentPoller.js
+++ b/operators/bosh-operator/BoshStaggeredDeploymentPoller.js
@@ -24,18 +24,16 @@ class BoshStaggeredDeploymentPoller extends BaseStatusPoller {
     const instanceId = resourceBody.metadata.name;
     const resourceOptions = _.get(resourceBody, 'spec.options');
     const deploymentName = _.get(resourceBody, 'status.response.deployment_name');
+    const operationType = _.get(resourceBody, 'status.response.type');
     return DirectorService
       .createInstance(instanceId, resourceOptions)
       .then(directorService => {
-        if (deploymentName) {
-          return directorService.createOrUpdateDeployment(deploymentName, resourceOptions);
-        } else {
-          const type = _.get(resourceBody, 'status.response.type');
-          if (type === CONST.OPERATION_TYPE.CREATE) {
-            return directorService.create(resourceOptions);
-          } else if (type === CONST.OPERATION_TYPE.UPDATE) {
-            return directorService.update(resourceOptions);
-          }
+        if (operationType === CONST.OPERATION_TYPE.CREATE) {
+          return directorService.create(resourceOptions, deploymentName);
+        } else if (operationType === CONST.OPERATION_TYPE.UPDATE) {
+          return directorService.update(resourceOptions, deploymentName);
+        } else if (operationType === CONST.OPERATION_TYPE.DELETE) {
+          return directorService.delete(resourceOptions, deploymentName);
         }
       })
       .then(directorResponse => {
--- a/operators/bosh-operator/BoshStaggeredDeploymentPoller.js
+++ b/operators/bosh-operator/BoshStaggeredDeploymentPoller.js
@@ -26,7 +26,18 @@ class BoshStaggeredDeploymentPoller extends BaseStatusPoller {
     const deploymentName = _.get(resourceBody, 'status.response.deployment_name');
     return DirectorService
       .createInstance(instanceId, resourceOptions)
-      .then(directorService => directorService.createOrUpdateDeployment(deploymentName, resourceOptions))
+      .then(directorService => {
+        if (deploymentName) {
+          return directorService.createOrUpdateDeployment(deploymentName, resourceOptions);
+        } else {
+          const type = _.get(resourceBody, 'status.response.type');
+          if (type === CONST.OPERATION_TYPE.CREATE) {
+            return directorService.create(resourceOptions);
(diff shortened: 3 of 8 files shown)


operators/bosh-operator/DirectorService.js
```diff
@@ -237,18 +247,22 @@ class DirectorService extends BaseDirectorService {
     };
     return this
       .initialize(operation)
+      .catch(err => logger.error(`Error occurred while finding network index for update
```
kamath-prasad: Just a question , in case of update why do we need to fail in case network segment index determination fails via BOSH ? Since during update the deployment name is already present , we can get it from the same , right ?

sauravmndl: Here we are not failing, staggered. Deployment name is figured out from local cache. If cache miss happens we go to bosh and figure out. So, if bosh calls fails here, staggered.

But one error handling missed here, ServiceInstanceNotFound, thanks for giving a comment here 👍 . would update.

sauravmndl:
@kamath-prasad 
Is dev pipeline run ?
==> Triggered, would update here once completed.

operators/bosh-operator/BoshStaggeredDeploymentPoller.js
```diff
@@ -26,7 +26,18 @@ class BoshStaggeredDeploymentPoller extends BaseStatusPoller {
     const deploymentName = _.get(resourceBody, 'status.response.deployment_name');
     return DirectorService
       .createInstance(instanceId, resourceOptions)
-      .then(directorService => directorService.createOrUpdateDeployment(deploymentName, resourceOptions))
+      .then(directorService => {
+        if (deploymentName) {
```
jagadish-kb: This multiple if/else seems a bit confusing at first, only to later figure out that if deploymentName exists then you dont have to initialize in case of create/update. How about changing the lines 30 - 38 like this:
```
resourceOptions._deploymentName = deploymentName;
directorService.update(resourceOptions);
```
You just update deployment name in resourceOptions and invoke just update and Changing DirectoService.initialize method like the below. Initialize will additionally take in second arg (payload), whatever is sent to create/update and if deployment exists, it straightaway initialises network segment index.
```
initialize(operation, params) {
    return Promise
      .try(() => {
        if(params._deploymentName) {
           return this.getNetworkSegmentIndex(deploymentName);
        }
        this.operation = operation.type;
        if (operation.type === CONST.OPERATION_TYPE.CREATE) {
          return this.acquireNetworkSegmentIndex(this.guid);
        }
        return this.findNetworkSegmentIndex(this.guid);
      })
      .tap(networkSegmentIndex => {
        assert.ok(_.isInteger(networkSegmentIndex), `Network segment index '${networkSegmentIndex}' must be an integer`);
        this.networkSegmentIndex = networkSegmentIndex;
      })
      .tap(() => {
        if (operation.type === 'delete') {
          return Promise
            .all([
              this.platformManager.preInstanceDeleteOperations({
                guid: this.guid
              }),
              this.deleteRestoreFile()
            ]);
        }
      });
  }
```

operators/bosh-operator/DirectorService.js
```diff
@@ -219,14 +220,23 @@ class DirectorService extends BaseDirectorService {
     };
     return this
       .initialize(operation)
+      .catch(err => {
+        logger.error(`Error occurred while acquiring network index for create
```
jagadish-kb: Do you want to push this error statement to LN 226 & add a logger.warn at LN229 with warn saying error occurred which is potentially BOSH being unavailable and deployment will continue to remain stagerred?

operators/bosh-operator/DirectorService.js
```diff
@@ -237,18 +247,27 @@ class DirectorService extends BaseDirectorService {
     };
     return this
       .initialize(operation)
+      .catch(err => {
+        logger.error(`Error occurred while finding network index for update
```
jagadish-kb: same comment as above with regards to logger.

operators/bosh-operator/DirectorService.js
sauravmndl: done

operators/bosh-operator/DirectorService.js
sauravmndl: done

operators/bosh-operator/BoshStaggeredDeploymentPoller.js
sauravmndl: did the same thing bit different way. But can discuss if needs to be changed.

common/constants.js
```diff
@@ -385,5 +385,6 @@ module.exports = Object.freeze({
   },
   DOCKER_HOST_CONFIG: {
     PIDS_LIMIT: 150
-  }
+  },
+  SYSTEM_ERRORS: ['ECONNREFUSED', 'ECONNRESET', 'EPIPE', 'ETIMEDOUT', 'ESOCKETTIMEDOUT']
```
abh1kg: Just a general question: in any case, the BOSH API status code is the driving factor for such errors. Do we need to handle these system errors as well? I know it's safer but just trying to understand-- if the response does not have a status code, that should be sufficient for a failure, isn't it?

operators/bosh-operator/BoshStaggeredDeploymentPoller.js
```diff
@@ -24,9 +24,18 @@ class BoshStaggeredDeploymentPoller extends BaseStatusPoller {
     const instanceId = resourceBody.metadata.name;
     const resourceOptions = _.get(resourceBody, 'spec.options');
     const deploymentName = _.get(resourceBody, 'status.response.deployment_name');
+    const operationType = _.get(resourceBody, 'status.response.type');
     return DirectorService
       .createInstance(instanceId, resourceOptions)
-      .then(directorService => directorService.createOrUpdateDeployment(deploymentName, resourceOptions))
+      .then(directorService => {
+        if (operationType === CONST.OPERATION_TYPE.CREATE) {
```
subhankarc: Use switch case instead.

sauravmndl: thought once :) will do.

common/constants.js
sauravmndl: want to staggered on connection failure or down. For other cases wanted to fail, e.g. EADDRINUSE, EMFILE (source https://nodejs.org/api/errors.html#errors_class_systemerror). So, generic status code not found will lead to staggered in those cases also. Is that fine ?

abh1kg: Sounds good...

operators/bosh-operator/BoshOperator.js
```diff
@@ -109,7 +109,7 @@ class BoshOperator extends BaseOperator {
         resourceId: _.get(changeObjectBody, 'metadata.name'),
         status: {
           response: response,
-          state: CONST.APISERVER.RESOURCE_STATE.IN_PROGRESS
+          state: _.get(response, 'task_id') ? CONST.APISERVER.RESOURCE_STATE.IN_PROGRESS : CONST.APISERVER.RESOURCE_STATE.WAITING
```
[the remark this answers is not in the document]
sauravmndl: Changed the title.

operators/bosh-operator/DirectorService.js
```diff
@@ -534,19 +547,26 @@ class DirectorService extends BaseDirectorService {
...
+  delete(params, deploymentName) {
     const operation = {
       type: 'delete'
     };
     return this
-      .initialize(operation)
-      .then(() => this.deleteDeployment(this.deploymentName, params))
+      .initialize(operation, deploymentName)
+      .then(() => {
+        if (this.networkSegmentIndex) {
```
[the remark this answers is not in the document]
sauravmndl: Just for the extra caution have kept it.

PR #456, opened 2018-10-31
Update Lodhash version

kamath-prasad:
Updating `lodhash`  library version.


kamath-prasad:
dev pipeline green on azure.

PR #459, opened 2018-11-02
Keep creating binding property for sf-mongodb bind

jayeshjk:
* In previous PR https://github.com/cloudfoundry-incubator/service-fabrik-broker/pull/437, code for creating bind properties on Bosh was removed.
* However, DbManager still fetches bind property of service-fabrik-mongodb from Bosh, which will create problem while recreating landscape. Hence as a temporary fix, allowing creation of bind property for service-fabrik-mongodb.


jayeshjk:
Dev pipeline # 246, relevant pipelines green.

jayeshjk:
@jagadish-kb, I agree that creating binding property in DbManager is the right way to proceed. As discussed I will raise another fix targeting this.

PR #460, opened 2018-11-03
Adding indexes for JobDetails and JobRunetails

mayanktiwari09:
Adding for attribute - 'type'

### DIFF
--- a/data-access-layer/db/JobDetailSchema.js
+++ b/data-access-layer/db/JobDetailSchema.js
@@ -48,15 +48,22 @@ const JobSchema = new Mongoose.Schema({
 
 JobSchema.index({
   type: 1,
-  createdAt: 1
-});
-JobSchema.index({
-  name: 1,
+  'data.instance_id': 1,
   createdAt: 1
 });
 JobSchema.index({
   type: 1,
+  'data.instance_id': 1
+});
+JobSchema.index({
+  type: 1,
   'data.instance_id': 1,
+  'response.delete_backup_status.instance_deleted': 1
+});
+JobSchema.index({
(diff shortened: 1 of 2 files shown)


data-access-layer/db/JobDetailSchema.js
```diff
@@ -45,5 +46,18 @@ const JobSchema = new Mongoose.Schema({
   }
 });
```
kamath-prasad: Add index for https://github.com/cloudfoundry-incubator/service-fabrik-broker/blob/master/reports/BackupReportManager.js#L164-L167

data-access-layer/db/JobRunDetailSchema.js
```diff
@@ -51,5 +52,18 @@ const JobRunDetailSchema = new Mongoose.Schema({
...
 
+JobRunDetailSchema.index({
+  type: 1,
+  createdAt: 1
+});
+JobRunDetailSchema.index({
+  name: 1,
+  createdAt: 1
+});
+JobRunDetailSchema.index({
```
kamath-prasad: which query does this index cover ?

data-access-layer/db/JobDetailSchema.js
```diff
@@ -45,5 +46,18 @@ const JobSchema = new Mongoose.Schema({
...
 
+JobSchema.index({
+  type: 1,
+  createdAt: 1
+});
+JobSchema.index({
+  name: 1,
+  createdAt: 1
+});
+JobSchema.index({
```
kamath-prasad: which query does this index qu

PR #461, opened 2018-11-05
[rel-10b]Keep creating binding property for sf-mongodb bind (#459)

jayeshjk:
* In previous PR #437, code for creating bind properties on Bosh was removed.
* However, DbManager still fetches bind property of service-fabrik-mongodb from Bosh, which will create problem while recreating landscape. Hence as a temporary fix, allowing creation of bind property for service-fabrik-mongodb.

### DIFF
--- a/operators/bosh-operator/DirectorService.js
+++ b/operators/bosh-operator/DirectorService.js
@@ -13,6 +13,7 @@ const NotFound = errors.NotFound;
 const ServiceInstanceNotFound = errors.ServiceInstanceNotFound;
 const ScheduleManager = require('../../jobs');
 const CONST = require('../../common/constants');
+const ServiceBindingAlreadyExists = errors.ServiceBindingAlreadyExists;
 const bosh = require('../../data-access-layer/bosh');
 const eventmesh = require('../../data-access-layer/eventmesh');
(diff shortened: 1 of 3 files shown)


operators/bosh-operator/DirectorService.js
```diff
@@ -725,6 +732,12 @@ class DirectorService extends BaseDirectorService {
         })
       )
       .tap(() => logger.info('+-> Deleted service credentials'))
+      .tap(() => {
```
sauravmndl: should be part of then() block.

operators/bosh-operator/DirectorService.js
```diff
@@ -686,6 +687,12 @@ class DirectorService extends BaseDirectorService {
         utils.maskSensitiveInfo(bindCreds);
         logger.info(`+-> Created binding:${JSON.stringify(bindCreds)}`);
       })
+      .tap(() => {
```
sauravmndl: should be part of then() block.

abh1kg: +1- the entire thing should be in the then block

operators/bosh-operator/DirectorService.js
abh1kg: +1

PR #462, opened 2018-11-05
implemented logic to enable multi-az only for internal users

jagadish-kb:
There is a need to enable multi-az only for internal customers only. In order to achieve this, have transformed the existing config variable `config.multi_az_enabled` from boolean to enum with values `internal`, `all` or `disabled`.
`internal` - multi-az enabled only for internal users
`all` - multi-az enabled for all users. [equivalent to set boolean `true`]
`disabled` - feature is disabled to all users [equivalent to set boolean `false`]


kamath-prasad:
@sauravmndl / @amitmalav : can you guys review this PR ?

PR #467, opened 2018-11-16
Add the login_hint parameter for password grant flow of accessing oauth token from cf uaa

kamath-prasad:
With this change we always force the identity provider to be UAA which will ensure that even if some other default identity providers are set oauth would be generated against UAA.

### DIFF
--- a/data-access-layer/cf/UaaClient.js
+++ b/data-access-layer/cf/UaaClient.js
@@ -117,7 +117,8 @@ class UaaClient extends HttpClient {
         grant_type: 'password',
         client_id: this.clientId,
         username: username,
-        password: password
+        password: password,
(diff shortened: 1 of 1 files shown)


data-access-layer/cf/UaaClient.js
```diff
@@ -117,7 +117,8 @@ class UaaClient extends HttpClient {
         grant_type: 'password',
         client_id: this.clientId,
         username: username,
-        password: password
+        password: password,
+        login_hint: '{"origin":"uaa"}'
```
abh1kg: This is fine actually- no need for a parameter because we can recommend creating users at the UAA for someone wanting to deploy the broker release. This means that an LDAP integration for these "technical users" will not be needed

PR #468, opened 2018-11-19
Modifying DbManager init flow to store/fetch credentials on ApiServer

jayeshjk:
* In continuation to previous changes towards bosh resiliency, we need to reduce bosh dependency in  case of service-fabrik-mongodb credentials also.
* With this PR, any new attempts to bind to service-fabrik-mongodb (e.g., landscape recreations, fresh setup etc) will result into storing credentials on ApiServer instead of bosh deployment property.
* In case of existing bind credentials, once they are fetched from bosh binding property they will also be populated on ApiServer and thenceforth will be fetched from ApiServer itself.
* (For previous workflow regarding this issue please see:  https://github.com/cloudfoundry-incubator/service-fabrik-broker/pull/437, https://github.com/cloudfoundry-incubator/service-fabrik-broker/pull/459)

### DIFF
--- a/broker/lib/fabrik/DBManager.js
+++ b/broker/lib/fabrik/DBManager.js
@@ -14,6 +14,7 @@ const NotFound = errors.NotFound;
 const CONST = require('../../../common/constants');
 const dbConnectionManager = require('../../../data-access-layer/db/DbConnectionManager');
 const BasePlatformManager = require('../../../platform-managers/BasePlatformManager');
+const eventmesh = require('../../../data-access-layer/eventmesh');
 
 /**
  * DB can be configured into ServiceFabrik by either providing the URL of already provisioned mongodb via 'config.mongodb.url'
@@ -101,14 +102,22 @@ class DBManager {
       if (this.bindInfo) {
         return this.bindInfo;
       }
-      return utils
-        .retry(() => this
-          .directorService
-          .getBindingProperty(config.mongodb.deployment_name, CONST.FABRIK_INTERNAL_MONGO_DB.BINDING_ID), {
-            maxAttempts: 5,
-            minDelay: 5000,
-            predicate: (err) => !(err instanceof ServiceBindingNotFound)
-          });
+      return eventmesh.apiServerClient.getResource({
+        resourceGroup: CONST.APISERVER.RESOURCE_GROUPS.BIND,
+        resourceType: CONST.APISERVER.RESOURCE_TYPES.DIRECTOR_BIND,
+        resourceId: _.toLower(CONST.FABRIK_INTERNAL_MONGO_DB.BINDING_ID)
+      })
+      .then(resource => _.get(resource, 'status.response'))
+      .catch(NotFound, () => {
+        return utils
+          .retry(() => this
+            .directorService
+            .getBindingProperty(config.mongodb.deployment_name, CONST.FABRIK_INTERNAL_MONGO_DB.BINDING_ID), {
+              maxAttempts: 5,
+              minDelay: 5000,
+              predicate: (err) => !(err instanceof ServiceBindingNotFound)
+            });
+        });
     });
   }
 
@@ -214,7 +223,21 @@ class DBManager {
           logger.info(`MongoDB ${operation} request is complete. Check status for task id - ${taskId}`);
           this.director
             .pollTaskStatusTillComplete(taskId)
-            .then(response => this.dbCreateUpdateSucceeded(response, createIfNotPresent))
+            .then(response => {
+              return Promise.try(() => {
+                if(createIfNotPresent) {
+                  //This is precaution to ensure that no previous bind resource exists in the event that
+                  //service-fabrik-mongodb is recreated but ApiServer is not.
+                  return eventmesh.apiServerClient.deleteResource({
+                    resourceGroup: CONST.APISERVER.RESOURCE_GROUPS.BIND,
+                    resourceType: CONST.APISERVER.RESOURCE_TYPES.DIRECTOR_BIND,
+                    resourceId: _.toLower(CONST.FABRIK_INTERNAL_MONGO_DB.BINDING_ID)
+                  });
+                }
+              })
+              .then(() => this.dbCreateUpdateSucceeded(response, createIfNotPresent))
+              .catch(NotFound, () => this.dbCreateUpdateSucceeded(response, createIfNotPresent))
+            })
             .catch(err => this.dbCreateUpdateFailed(err, operation));
         })
         .catch(err => this.dbCreateUpdateFailed(err, operation));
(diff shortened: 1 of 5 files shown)


kamath-prasad:
@jagadish-kb / @amitmalav : can you pls review/merge this PR ?

broker/lib/fabrik/DBManager.js
```diff
@@ -251,6 +287,21 @@ class DBManager {
...
+    return eventmesh.apiServerClient.createResource({
+      resourceGroup: CONST.APISERVER.RESOURCE_GROUPS.BIND,
+      resourceType: CONST.APISERVER.RESOURCE_TYPES.DIRECTOR_BIND,
+      resourceId: bindProperty.id,
+      options: {
+        binding_id: bindProperty.id
+      },
+      status: {
+        state: CONST.APISERVER.RESOURCE_STATE.SUCCEEDED,
+        response: bindProperty
```
amitmalav: We usually store [base64 encoded](https://github.com/cloudfoundry-incubator/service-fabrik-broker/blob/master/operators/bosh-operator/BoshBindOperator.js#L60) credentials in binding resources.
Is it possible to do same here?

operators/bosh-operator/DirectorService.js
```diff
@@ -779,12 +772,6 @@ class DirectorService extends BaseDirectorService {
         })
       )
       .tap(() => logger.info('+-> Deleted service credentials'))
-      .tap(() => {
-        //TODO: Temporary fix. Binding info should always be fetched from etcd.
-        if (deploymentName === config.mongodb.deployment_name) {
```
amitmalav: In current flow, there is no provision for deleting mongodb binding property??

broker/lib/fabrik/DBManager.js
jayeshjk: Yes, will modify. I had missed this. Thanks for pointing out.

operators/bosh-operator/DirectorService.js
jayeshjk: Yes, AFAIK I didn't see any unbind call for service-fabrik-mongodb.

kamath-prasad:
@jayeshjk  : can you pls incorporate the review comments ?

broker/lib/fabrik/DBManager.js
```diff
@@ -214,7 +231,21 @@ class DBManager {
           logger.info(`MongoDB ${operation} request is complete. Check status for task id - ${taskId}`);
           this.director
             .pollTaskStatusTillComplete(taskId)
-            .then(response => this.dbCreateUpdateSucceeded(response, createIfNotPresent))
+            .then(response => {
+              return Promise.try(() => {
```
jagadish-kb: unnecessary `Promise.try` can remove this from with a then block.

broker/lib/fabrik/DBManager.js
```diff
@@ -214,7 +231,21 @@ class DBManager {
...
+                  if (createIfNotPresent) {
+                    //This is precaution to ensure that no previous bind resource exists in the event that
```
jagadish-kb: better to update the resource if it exists. Currently you are handling only NotFound exception here, but due to some other issue if delete fails, operation will result in failure. Also since bind is being fired without checking, you need to handle NotFound, specifically. However if you had this as part of getResource then either create/update as part of storeBindingProperty it would make it more readable. 
Even this approach is fine, if you want to continue this, move this check at LN228. So that if this fails, unnecessarily you dont end up provisioning DB and you fail fast.

broker/lib/fabrik/DBManager.js
```diff
@@ -240,8 +271,13 @@ class DBManager {
             this.bindInfo = {
               credentials: credentials
             };
-            this.initialize();
-          });
+            return this.storeBindPropertyOnApiServer({
```
jagadish-kb: put this in utils.retry, so that in case of etcd errors we at-least retry.

broker/lib/fabrik/DBManager.js
```diff
@@ -214,7 +231,21 @@ class DBManager {
...
+                    //service-fabrik-mongodb is recreated but ApiServer is not.
+                    return eventmesh.apiServerClient.deleteResource({
+                      resourceGroup: CONST.APISERVER.RESOURCE_GROUPS.BIND,
+                      resourceType: CONST.APISERVER.RESOURCE_TYPES.DIRECTOR_BIND,
+                      resourceId: _.toLower(CONST.FABRIK_INTERNAL_MONGO_DB.BINDING_ID)
+                    });
+                  }
+                })
+                .then(() => this.dbCreateUpdateSucceeded(response, createIfNotPresent))
+                .catch(NotFound, () => this.dbCreateUpdateSucceeded(response, createIfNotPresent));
```
jagadish-kb: can move this `catch(NotFound), ()=> logger.debug('DB binding not present. ignoring it'))` to LN243 to avoid dupe `dbCreateUpdateSucceeded`

operators/bosh-operator/DirectorService.js
jagadish-kb: @jayeshjk you are right. There is no need for unbind of internal mongodb.

broker/lib/fabrik/DBManager.js
jayeshjk: I agree that readability will improve if this logic moves to storeBindingPropery. So if I understand correctly: In storeBindingProperty, based on output of getResource call (NotFound or successful)we will either create or update the bind resource. Can you please confirm this understanding?

broker/lib/fabrik/DBManager.js
jayeshjk: This will be resolved if this logic moves to storeBindingProperty method.

broker/lib/fabrik/DBManager.js
```diff
@@ -101,14 +102,30 @@ class DBManager {
...
-          });
+      return eventmesh.apiServerClient.getResource({
+          resourceGroup: CONST.APISERVER.RESOURCE_GROUPS.BIND,
+          resourceType: CONST.APISERVER.RESOURCE_TYPES.DIRECTOR_BIND,
+          resourceId: _.toLower(CONST.FABRIK_INTERNAL_MONGO_DB.BINDING_ID)
+        })
+        .then(resource => utils.decodeBase64(_.get(resource, 'status.response')))
+        .catch(NotFound, () => {
+          return utils
+            .retry(() => this
```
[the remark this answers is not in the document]
jayeshjk: Done.

broker/lib/fabrik/DBManager.js
jayeshjk: Done.

kamath-prasad:
@amitmalav : can you pls check and approve/merge this PR ?

PR #469, opened 2018-11-19
Throw exception when deployment name is not present in task id

jayeshjk:
* This PR addresses code in getTask and other methods of BoshDirectorClient, which by default communicate to landscape bosh (not bosh_sf) when taskId passed is not in required format.
* This is not a desired behaviour because:
1. we no longer actively create deployments on landscape bosh
2. taskId should always be in required format (see here: https://github.com/cloudfoundry-incubator/service-fabrik-broker/blob/rel-2018.T10b/data-access-layer/bosh/BoshDirectorClient.js#L498 and https://github.com/cloudfoundry-incubator/service-fabrik-broker/blob/rel-2018.T10b/data-access-layer/bosh/BoshDirectorClient.js#L509)
* Therefore replacing this behaviour with exception.

### DIFF
--- a/data-access-layer/bosh/BoshDirectorClient.js
+++ b/data-access-layer/bosh/BoshDirectorClient.js
@@ -829,12 +829,7 @@ class BoshDirectorClient extends HttpClient {
   getTask(taskId) {
     const splitArray = this.parseTaskid(taskId);
     if (splitArray === null) {
-      return this
-        .makeRequestWithConfig({
-          method: 'GET',
-          url: `/tasks/${taskId}`
-        }, 200, this.getConfigByName(CONST.BOSH_DIRECTORS.BOSH))
-        .then(res => JSON.parse(res.body));
+      throw new UnprocessableEntity(`not able to query correct bosh as taskId is not in required format: ${taskId}`);
     }
     const deploymentName = splitArray[1];
     const taskIdAlone = splitArray[2];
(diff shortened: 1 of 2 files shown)


data-access-layer/bosh/BoshDirectorClient.js
```diff
@@ -829,12 +829,7 @@ class BoshDirectorClient extends HttpClient {
   getTask(taskId) {
     const splitArray = this.parseTaskid(taskId);
     if (splitArray === null) {
-      return this
-        .makeRequestWithConfig({
-          method: 'GET',
-          url: `/tasks/${taskId}`
-        }, 200, this.getConfigByName(CONST.BOSH_DIRECTORS.BOSH))
-        .then(res => JSON.parse(res.body));
+      throw new errors.UnprocessableEntity(`not able to query correct bosh as taskId is not in required format: ${taskId}`);
```
sauravmndl: Just start the statement in upper case character and add UT. Rest looks fine.

jayeshjk: Will do. Thanks

data-access-layer/bosh/BoshDirectorClient.js
```diff
@@ -829,12 +829,7 @@ class BoshDirectorClient extends HttpClient {
   getTask(taskId) {
     const splitArray = this.parseTaskid(taskId);
     if (splitArray === null) {
-      return this
-        .makeRequestWithConfig({
-          method: 'GET',
-          url: `/tasks/${taskId}`
-        }, 200, this.getConfigByName(CONST.BOSH_DIRECTORS.BOSH))
-        .then(res => JSON.parse(res.body));
+      throw new errors.UnprocessableEntity(`Not able to query correct bosh as taskId is not in required format: ${taskId}`);
```
amitmalav: Putting example of `required format` in comment above will help

PR #473, opened 2018-11-22
Remove dependency on director

amitmalav:
Changes:
1) Make populateCache() dependency configurable
2) Configuration to pass kubeconfig for apiserver client

### DIFF
--- a/common/utils/HttpClient.js
+++ b/common/utils/HttpClient.js
@@ -30,7 +30,7 @@ class HttpClient {
       if (httpCircuitConfig && httpCircuitConfig.apis) {
         this.buildCommandFactory(this.baseUrl);
       } else {
-        logger.warn('Circuit breaker config not found for HTTP. Hystrix will not be configured for ${options.baseUrl}');
+        logger.warn(`Circuit breaker config not found for HTTP. Hystrix will not be configured for ${options.baseUrl}`);
       }
     }
   }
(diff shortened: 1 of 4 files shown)


common/utils/HttpClient.js
```diff
@@ -30,7 +30,7 @@ class HttpClient {
       if (httpCircuitConfig && httpCircuitConfig.apis) {
         this.buildCommandFactory(this.baseUrl);
       } else {
-        logger.warn('Circuit breaker config not found for HTTP. Hystrix will not be configured for ${options.baseUrl}');
+        logger.warn(`Circuit breaker config not found for HTTP. Hystrix will not be configured for ${options.baseUrl}`);
```
subhankarc: +1

data-access-layer/eventmesh/ApiServerClient.js
```diff
@@ -16,13 +16,15 @@ const NotFound = errors.NotFound;
 const Conflict = errors.Conflict;
 const InternalServerError = errors.InternalServerError;
 
-const apiserver = new kc.Client({
-  config: {
+const apiserverConfig = config.apiserver.pathToKubeConfig ?
```
subhankarc: check how the kubeconfig can be found from within the pod itself.

Can you try this? https://github.com/godaddy/kubernetes-client/blob/d17bb81cdf4c45289b10aba84c586f07bfdf0b75/test/config.test.js#L44

subhankarc: let's stick to the current approach for now.

PR #476, opened 2018-11-23
fix: isMultiAzDeploymentEnabled caller to evaluate promise response

jagadish-kb:
PlatformManager.isMultiAzDeploymentEnabled is promisified function, whose value was being assigned as is, without the evaluation of the promise to a boolean attrib. So this made multi-az true by default. Fixing it by evaluating the promise and then proceeding with manifest generation.

### DIFF
--- a/operators/bosh-operator/DirectorService.js
+++ b/operators/bosh-operator/DirectorService.js
@@ -860,46 +860,53 @@ class DirectorService extends BaseDirectorService {
   }
 
   generateManifest(deploymentName, opts, preDeployResponse, preUpdateAgentResponse) {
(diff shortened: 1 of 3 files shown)


kamath-prasad:
Again we got the same error : 

```
 1 failing
  1) #DirectorService instances director #bind no context : returns 201 Created:
     Uncaught AssertionError: expected false to be true
      at Function.verify (test/test_broker/mocks/index.js:48:3)
      at Timeout.setTimeout (test/test_broker/operators.DirectorService.spec.js:973:23)
``` 

i retrigger travis for now.

PR #480, opened 2018-11-27
Fix: Upgrade workflow is running with READ Lock type

jagadish-kb:
Fixes - HCPCFS-2122
1. Operation Type for workflow was set as update and this operation is flagged as an asncyh_operation that can run in parallel. So this leads to workflow getting read lock. Added new operation type 'update_workflow' and assocaited it with WRITE_LOCK.
2.  During some race conditions (for ex., StreamRefresh & broker restart) where in state of task could not be updated, there were cases of same task being created again. As part of fixing the dupe task issue, made task id as a function of workflow_id and task_order. So this way even if race condition leads to a creation of dupe task scenario it will lead to an error (name of resource is to be unique in etcd) which will be ignored.

### DIFF
--- a/operators/BaseOperator.js
+++ b/operators/BaseOperator.js
@@ -128,7 +128,7 @@ class BaseOperator {
       } else {
         processingLockStatus.conflict = true;
         logger.debug(`Resource ${objectBody.metadata.name}  ${resourceGroup}, ${resourceType}, ${queryString} -` +
-          ` is picked by process with ip ${lockedByManager} at ${processingStartedAt} will be held for another ${elapsedTimeInMs}(ms)`);
+          ` is picked by process with ip ${lockedByManager} at ${processingStartedAt} will be held for another ${elapsedTimeInMs - CONST.PROCESSING_REQUEST_BY_MANAGER_TIMEOUT}(ms)`);
       }
     });
   }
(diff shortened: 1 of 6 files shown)


jagadish-kb:
@kamath-prasad - Not sure why the coverage shows as reduced. I have atleast manually verified and all modified lines are covered. Since the drop is 0.002% its pretty hard to notice where exactly is this drop :-)

operators/BaseOperator.js
```diff
@@ -127,7 +127,8 @@ class BaseOperator {
           });
       } else {
         processingLockStatus.conflict = true;
-        logger.debug(`Resource ${objectBody.metadata.name}  ${resourceGroup}, ${resourceType}, ${queryString} - is picked by process with ip ${lockedByManager} at ${processingStartedAt}`);
+        logger.debug(`Resource ${objectBody.metadata.name}  ${resourceGroup}, ${resourceType}, ${queryString} -` +
+          ` is picked by process with ip ${lockedByManager} at ${processingStartedAt} will be held for another ${elapsedTimeInMs}(ms)`);
```
amitmalav: Shouldn't it be => Will be held for another `elapsedTimeInMs  - CONST.PROCESSING_REQUEST_BY_MANAGER_TIMEOUT`?

amitmalav:
Travis build was failing, probably that's why coverall is also messed up. I have re-triggered the build

PR #489, opened 2018-11-30
Adding support for query parameter specifying login_hint in dashboard URL

jayeshjk:
* With this PR dashboard URL can be optionally appended with a query parameter specifying login_hint.

### DIFF
--- a/data-access-layer/cf/UaaClient.js
+++ b/data-access-layer/cf/UaaClient.js
@@ -22,12 +22,12 @@ class UaaClient extends HttpClient {
     this.clientId = config.cf.client_id || 'cf';
   }
 
-  authorizationUrl(options, addLoginHint) {
+  authorizationUrl(options, loginHint) {
     options = _.assign({
       response_type: 'code'
     }, options);
-    if (addLoginHint) {
-      options.login_hint = '{"origin":"uaa"}';
+    if (loginHint && loginHint !== '') {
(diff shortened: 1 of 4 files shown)


data-access-layer/cf/UaaClient.js
```diff
@@ -22,11 +22,13 @@ class UaaClient extends HttpClient {
     this.clientId = config.cf.client_id || 'cf';
   }
 
-  authorizationUrl(options) {
+  authorizationUrl(options, addLoginHint) {
```
kamath-prasad: here why is this a boolean ? Ideally it will have a value as "uaa" right ? So if its a non null string we can directly assign it to the origin value below .. So that it becomes extensible in future.

```
if (addLoginHint!==null && addLoginHint !== undefined) {
      options.login_hint = '{"origin":addLoginHint}';
    }
```

something like this.

jayeshjk: My understanding was that login_hint will be needed for UAA case only, hence the hardcoding here. However, in the view of extensibility will make this change.

kamath-prasad: Yes , that is correct . But in future if origin is different then we would not need changes in the broker.

jayeshjk: Understood. Will push the change shortly

PR #492, opened 2018-11-30
Restructure code to get rid of optional dependencies

amitmalav:
This PR contains changes which are required to run `broker v2 apis` only with minimal settings.
Hence Unnecessary dependencies on modules like (cf, fabrik) were delegated.
Dependency on `fabrik` is moved from `FabrikBaseController` to individual controllers which need fabrik module.

Sample minimal settings file can be found [here](https://gist.github.com/amitmalav/452a32d9c7c2c3fea2a32f5e7de744dc)

### DIFF
--- a/common/utils/index.js
+++ b/common/utils/index.js
@@ -17,7 +17,6 @@ const EventLogDomainSocketClient = require('./EventLogDomainSocketClient');
 const EventLogDBClient = require('./EventLogDBClient');
 const EventLogInterceptor = require('../EventLogInterceptor');
 const errors = require('../errors');
-const BasePlatformManager = require('../../platform-managers/BasePlatformManager');
 const NotImplemented = errors.NotImplemented;
 exports.HttpClient = HttpClient;
 exports.RetryOperation = RetryOperation;
@@ -595,6 +594,7 @@ function deploymentStaggered(err) {
 }
 
 function getPlatformManager(context) {
+  const BasePlatformManager = require('../../platform-managers/BasePlatformManager');
   let platform = context.platform;
   if (platform === CONST.PLATFORM.SM) {
     platform = context.origin;


broker/lib/index.js
```diff
@@ -1,16 +1,15 @@
...
 exports.middleware = require('./middleware');
-exports.fabrik = require('./fabrik');
 exports.bootstrap = bootstrap;
 const config = require('../../common/config');
 if (config.enable_swarm_manager) {
   exports.docker = require('../../data-access-layer/docker');
 }
 const logger = require('../../common/logger');
 const docker = exports.docker;
+require('./fabrik/UnlockResourcePoller');
```
subhankarc: move this UnlockResourcePoller out of fabrik.

subhankarc: put it in common.

common/utils/index.js
```diff
@@ -590,4 +592,17 @@ function deploymentStaggered(err) {
   const response = _.get(err, 'error', {});
   const description = _.get(response, 'description', '');
   return description.indexOf(CONST.FABRIK_OPERATION_STAGGERED) > 0 && description.indexOf(CONST.FABRIK_OPERATION_COUNT_EXCEEDED) > 0;
+}
+
+function getPlatformManager(context) {
```
subhankarc: Use it in DirectorService as well.

subhankarc:
@kamath-prasad  Can you please review and merge this? We want to merge it to master as this does not have any feature change, only dependency reduction.

broker/lib/index.js
amitmalav: Done!

common/utils/index.js
amitmalav: Done!

PR #495, opened 2018-12-05
Introducing config for schedule-update

pritishmishra:
The idea here is to have an easy way to enable/disable the `schedule_update` feature for service.
Thus, the approach here is to expose an admin API endpoint to pass a config parameter with pre-defined key (`disable_scheduled_update_` + service-name). This config is read at the run-time of the job and if the update is disabled, then the update doesn't run.
To make things generic, introducing a configMap resource to store the configs. This is to toggle any other feature in the future.
There are basically 2 endpoints:
1. Creating/Updating a config
2. Getting the config (this can be used for internal feature functions)

### DIFF
--- a/api-controllers/ServiceFabrikAdminController.js
+++ b/api-controllers/ServiceFabrikAdminController.js
@@ -72,10 +72,10 @@ class ServiceFabrikAdminController extends FabrikBaseController {
     return Promise.try(() => {
       logger.info(`Forbidden Manifest flag set to ${allowForbiddenManifestChanges}`);
       return eventmesh.apiServerClient.getResource({
-        resourceGroup: CONST.APISERVER.RESOURCE_GROUPS.DEPLOYMENT,
-        resourceType: CONST.APISERVER.RESOURCE_TYPES.DIRECTOR,
-        resourceId: instanceId
-      })
+          resourceGroup: CONST.APISERVER.RESOURCE_GROUPS.DEPLOYMENT,
+          resourceType: CONST.APISERVER.RESOURCE_TYPES.DIRECTOR,
+          resourceId: instanceId
+        })
         .catch(errors.NotFound, () => undefined)
         .then(resource => _.get(resource, 'spec.options'))
         .then(resource => {
@@ -109,9 +109,9 @@ class ServiceFabrikAdminController extends FabrikBaseController {
     const deploymentName = instanceDetails.deployment_name;
     logger.debug(`Getting outdated diff for  :  ${deploymentName}`);
     return DirectorService.createInstance(instanceDetails.instance_id, {
-      plan_id: plan.id,
-      context: tenantInfo.context
-    })
+        plan_id: plan.id,
+        context: tenantInfo.context
+      })
       .then(directorInstance => directorInstance.diffManifest(deploymentName, tenantInfo))
       .tap(result => logger.debug(`Diff of manifest for ${deploymentName} is ${result.diff}`))
       .then(result => result.diff);
@@ -217,24 +217,24 @@ class ServiceFabrikAdminController extends FabrikBaseController {
           resourceType: plan.resourceType,
           resourceId: this.getInstanceId(deploymentName)
         })
-          .then(context => {
-            const opts = {};
-            opts.context = context;
-            return Promise
-              .all([
-                this.director.getDeploymentVmsVitals(deploymentName),
-                this.director.getTasks({
-                  deployment: deploymentName
-                }),
-                manager.diffManifest(deploymentName, opts).then(utils.unifyDiffResult)
-              ])
-              .spread((vms, tasks, diff) => ({
-                name: deploymentName,
-                diff: diff,
-                tasks: _.filter(tasks, task => !_.startsWith(task.description, 'snapshot')),
-                vms: _.filter(vms, vm => !_.isNil(vm.vitals))
-              }));
-          })
+        .then(context => {
+          const opts = {};
+          opts.context = context;
+          return Promise
+            .all([
+              this.director.getDeploymentVmsVitals(deploymentName),
+              this.director.getTasks({
+                deployment: deploymentName
+              }),
+              manager.diffManifest(deploymentName, opts).then(utils.unifyDiffResult)
+            ])
+            .spread((vms, tasks, diff) => ({
+              name: deploymentName,
+              diff: diff,
+              tasks: _.filter(tasks, task => !_.startsWith(task.description, 'snapshot')),
+              vms: _.filter(vms, vm => !_.isNil(vm.vitals))
+            }));
+        })
       )
       .then(locals => {
         res.format({
@@ -333,10 +333,10 @@ class ServiceFabrikAdminController extends FabrikBaseController {
           return false;
         }
         return eventmesh.apiServerClient.getPlatformContext({
-          resourceGroup: CONST.APISERVER.RESOURCE_GROUPS.DEPLOYMENT,
-          resourceType: CONST.APISERVER.RESOURCE_TYPES.DIRECTOR,
-          resourceId: this.getInstanceId(deployment.name)
-        })
+            resourceGroup: CONST.APISERVER.RESOURCE_GROUPS.DEPLOYMENT,
+            resourceType: CONST.APISERVER.RESOURCE_TYPES.DIRECTOR,
+            resourceId: this.getInstanceId(deployment.name)
+          })
           .then(context => {
             const opts = {};
             opts.context = context;
@@ -594,11 +594,11 @@ class ServiceFabrikAdminController extends FabrikBaseController {
           .value();
 
         return ScheduleManager.schedule(
-          req.params.name,
-          CONST.JOB.SCHEDULED_OOB_DEPLOYMENT_BACKUP,
-          req.body.repeatInterval,
-          data,
-          req.user)
+            req.params.name,
+            CONST.JOB.SCHEDULED_OOB_DEPLOYMENT_BACKUP,
+            req.body.repeatInterval,
+            data,
+            req.user)
           .then(body => res
             .status(201)
             .send(body));
--- a/api-controllers/ServiceFabrikAdminController.js
+++ b/api-controllers/ServiceFabrikAdminController.js
@@ -72,10 +72,10 @@ class ServiceFabrikAdminController extends FabrikBaseController {
     return Promise.try(() => {
       logger.info(`Forbidden Manifest flag set to ${allowForbiddenManifestChanges}`);
       return eventmesh.apiServerClient.getResource({
-          resourceGroup: CONST.APISERVER.RESOURCE_GROUPS.DEPLOYMENT,
-          resourceType: CONST.APISERVER.RESOURCE_TYPES.DIRECTOR,
-          resourceId: instanceId
-        })
+        resourceGroup: CONST.APISERVER.RESOURCE_GROUPS.DEPLOYMENT,
+        resourceType: CONST.APISERVER.RESOURCE_TYPES.DIRECTOR,
+        resourceId: instanceId
+      })
         .catch(errors.NotFound, () => undefined)
         .then(resource => _.get(resource, 'spec.options'))
         .then(resource => {
@@ -109,9 +109,9 @@ class ServiceFabrikAdminController extends FabrikBaseController {
     const deploymentName = instanceDetails.deployment_name;
     logger.debug(`Getting outdated diff for  :  ${deploymentName}`);
     return DirectorService.createInstance(instanceDetails.instance_id, {
-        plan_id: plan.id,
-        context: tenantInfo.context
-      })
+      plan_id: plan.id,
+      context: tenantInfo.context
+    })
       .then(directorInstance => directorInstance.diffManifest(deploymentName, tenantInfo))
       .tap(result => logger.debug(`Diff of manifest for ${deploymentName} is ${result.diff}`))
       .then(result => result.diff);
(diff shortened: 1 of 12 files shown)


api-controllers/ServiceFabrikAdminController.js
```diff
@@ -735,6 +736,33 @@ class ServiceFabrikAdminController extends FabrikBaseController {
         .send(body));
   }
 
+  createUpdateConfig(req, res) {
```
kamath-prasad: why is create and update clubbed into 1 ? For all other entities we have separate functions right ?

PUT - representing update.
POST - representing create.

api-controllers/routes/admin.js
```diff
@@ -32,6 +32,12 @@ router.route('/backups')
 router.route('/backups/:backup_guid/delete')
   .post(controller.handler('deleteBackup'))
   .all(commonMiddleware.methodNotAllowed(['POST']));
+router.route('/config/create')
```
kamath-prasad: same comment as above..

common/constants.js
```diff
@@ -148,6 +148,9 @@ module.exports = Object.freeze({
     HUMAN_INTERVAL: 'human-readable-interval',
     CRON_EXPRESSION: 'cron_expression'
   },
+  CONFIG: {
```
kamath-prasad: better to rename as `CONFIG_KEYS`

common/constants.js
```diff
@@ -148,6 +148,9 @@ module.exports = Object.freeze({
...
+    SCHEDULED_UPDATE_CONFIG_PREFIX: 'disable_scheduled_update_'
```
kamath-prasad: `SCHEDULED_UPDATE_CONFIG_PREFIX` --> `DISABLE_SCHEDULED_UPDATE_PREFIX` ?

jobs/ServiceInstanceUpdateJob.js
```diff
@@ -51,9 +51,20 @@ class ServiceInstanceUpdateJob extends BaseJob {
...
-          this.handleInstanceDeletion(instanceDetails, operationResponse, done) :
-          this.updateInstanceIfOutdated(instanceDetails, resourceDetails, operationResponse)))
+        .then(resourceDetails => {
+          if (resourceDetails) {
+            return this.isInstanceUpdateDisabled(resourceDetails)
+              .then(disabled => {
+                if (!disabled) {
+                  return this.updateInstanceIfOutdated(instanceDetails, resourceDetails, operationResponse);
+                }
+                operationResponse.update_init = 'disabled';
```
kamath-prasad: @vinaybheri : can you pls check once we disable scheduled update for any service type can it have update on the scheduled update IT test which runs during deploy ?

@pritishmishra - maybe you can check once with Vinay and make necessary adjustments to that script too to handle this case.

common/constants.js
pritishmishra: okay.

jobs/ServiceInstanceUpdateJob.js
pritishmishra: Correct me if I am wrong here. But if the Scheduled Update IT triggers an update on the service having the feature disabled, it should anyway succeed. Will that still fail the test?

api-controllers/ServiceFabrikAdminController.js
kamath-prasad: It is the REST standard we have followed in all API's in the broker. Hence even in this case the same should hold good.

@jagadish-kb : please share your thoughts here.

jagadish-kb: @kamath-prasad especially for this config use case i would recommend merging create/update into one because generally if someone calls a PUT and resource does not exist, one must throw error saying Resource Not Found. Similarly for create if it already exists, we must throw error. Since the configs are something we might toggle, if we go with strict Create & Update, then invokers must first fetch and then do POST or PUT. Hence to keep things simple, it will always be PUT from consumer perspective and merging both Create/Update into one.

jobs/ServiceInstanceUpdateJob.js
jagadish-kb: @kamath-prasad -- Its fine for the tests to fail right? We would have disabled it for some reason and we can fail them. Or alternatively if you want to ensure your pipelines are green, IT Test before running the schedule update/backup tests can get this config and proceed with test only if its not disabled.

api-controllers/routes/admin.js
jagadish-kb: avoid verbs create/update/get in URIs. HTTP methods signify the same.

api-controllers/routes/admin.js
```diff
@@ -32,6 +32,12 @@ router.route('/backups')
...
+  .post(controller.handler('createUpdateConfig'))
```
jagadish-kb: change the HTTPMethod to PUT. POST is generally for create and PUT for update. In this case since updates will keep happening its better to make this PUT.

api-controllers/routes/admin.js
```diff
@@ -32,6 +32,12 @@ router.route('/backups')
...
+  .all(commonMiddleware.methodNotAllowed(['POST']));
```
jagadish-kb: once above line is changed, change here also to PUT.

common/constants.js
```diff
@@ -271,6 +274,12 @@ module.exports = Object.freeze({
     API_VERSION: 'v1alpha1',
     CRD_RESOURCE_GROUP: 'apiextensions.k8s.io',
     PATCH_CONTENT_TYPE: 'application/merge-patch+json',
+    CONFIG_MAP: {
+      API_VERSION: 'v1',
+      RESOURCE_TYPE: 'configmaps',
+      RESOURCE_NAME: 'sfconfig',
```
jagadish-kb: shouldnt this constant go into CONFIG? Because config_map is generic and one can create more than one config map with a different resource name. So i would move this to `CONFIG.SF_CONFIG`

common/constants.js
jagadish-kb: Better not to rename this, as the value under is not a key but a prefix to the key. Also anything else related to config can be put under this `CONFIG` key. Ex., RESOURCE_NAME can be moved from `CONFIG_MAP` to here & rename `RESOURCE_NAME` to `SF_CONFIG`

api-controllers/ServiceFabrikAdminController.js
kamath-prasad: Ok .. Fine ..

data-access-layer/eventmesh/ApiServerClient.js
```diff
@@ -488,6 +488,75 @@ class ApiServerClient {
...
+    const metadata = {
+      name: configName
+    };
+    let data = {};
+    data = _.set(data, configParam.key, configParam.value);
+    const resourceBody = {
+      apiVersion: CONST.APISERVER.CONFIG_MAP.API_VERSION,
+      kind: CONST.APISERVER.CONFIG_MAP.RESOURCE_KIND,
+      metadata: metadata,
+      data: data
```
jagadish-kb: remove `data` variable and directly put `_.pick(configParam, 'key', 'value')` here.

data-access-layer/eventmesh/ApiServerClient.js
```diff
@@ -488,6 +488,75 @@ class ApiServerClient {
...
+    };
+    return Promise.try(() => this.init())
```
jagadish-kb: Unncessary `Promise.try` - remove it. init already returns a promise, can just chain it.

data-access-layer/eventmesh/ApiServerClient.js
```diff
@@ -488,6 +488,75 @@ class ApiServerClient {
...
+    const metadata = {
+      name: configName
+    };
+    let data = {};
+    data = _.set(data, configParam.key, configParam.value);
+    const resourceBody = {
+      apiVersion: CONST.APISERVER.CONFIG_MAP.API_VERSION,
+      kind: CONST.APISERVER.CONFIG_MAP.RESOURCE_KIND,
+      metadata: metadata,
+      data: data
```
jagadish-kb: Same as above with regards to `data`

data-access-layer/eventmesh/ApiServerClient.js
```diff
@@ -488,6 +488,75 @@ class ApiServerClient {
...
+    };
+    return Promise.try(() => this.init())
```
jagadish-kb: Unnecessary Promise.try

data-access-layer/eventmesh/ApiServerClient.js
```diff
@@ -488,6 +488,75 @@ class ApiServerClient {
...
+      .catch(errors.NotFound, () => {
+        return this.createConfigMapResource(configName, configParam);
+      });
+  }
+
+  getConfigMap(configName, key) {
+    return this.getConfigMapResource(configName).then((body) => {
+      const configParam = body.data;
+      return _.get(configParam, key);
+    });
```
jagadish-kb: Can replace LN 554-557 with the below
```
return this.getConfigMapResource(configName).then((body)  => _.get(body.data, 'key'));
```

data-access-layer/eventmesh/ApiServerClient.js
```diff
@@ -488,6 +488,75 @@ class ApiServerClient {
...
+      apiVersion: CONST.APISERVER.CONFIG_MAP.API_VERSION,
+      kind: CONST.APISERVER.CONFIG_MAP.RESOURCE_KIND,
+      metadata: metadata,
+      data: data
+    };
+    return Promise.try(() => this.init())
+      .then(() => this.getConfigMapResource(configName))
+      .then((oldResourceBody) => {
+        resourceBody.data = oldResourceBody.data ? _.merge(oldResourceBody.data, data) : resourceBody.data;
+        return apiserver.api[CONST.APISERVER.CONFIG_MAP.API_VERSION]
```
amitmalav: In my view, this can lead to race conditions. Try patching with `resourceVersion` also in resource body.

Race condition:
=> Thread 1 => Get config map (c1)
=> Thread 2 => Update it with some random data ()
=> Thread 1 => Update config map c1
Now without resource Version in c1 `body.metadata` 3rd call will overwrite whatever changes done by 2nd call.
If you pass resource version of c1 in 3rd call, it will through `409` if object has been changed meanwhile (Here by 2nd call)

jobs/ServiceInstanceUpdateJob.js
kamath-prasad: As discussed with you there are 2 things : 

- If some service disables the scheduled update , then deploy-service-fabrik will always fail in the scheduled update test for that service until it gets re-enabled. Hence this can block the whole pipeline since we can never be sure how long that service takes to fix its updates.

- If some update of the test instances fails in post deploy of service-fabrik , then we must auto disable the scheduled update for that service so that it prevents other productive instances from getting updated and hence fail.

So I propose the following flow changes in the scheduled update test in SF post deploy with this PR : 

- Before scheduling , check if the config for that service is disabled . If disabled , then skip the test for that service.
- If some test fails , then update the config to disabled via script for that service. This will ensure that no other instances of that service type will be updated.
- if some test passes , then update the config to enabled via script for that service type.

data-access-layer/eventmesh/ApiServerClient.js
pritishmishra: Actually, @jagadish-kb, the idea is to have the resultant as `{keyName:valueName}`. I believe, `_.pick` will result in `{key: <keyName>, value: <valueName>}`. 
Hence, I modified this. What will be your suggestion?

common/constants.js
kamath-prasad: ok.. makes sense..

data-access-layer/eventmesh/ApiServerClient.js
pritishmishra: done!. just one change here, `key` is a variable not the literal string.

common/constants.js
pritishmishra: done!

api-controllers/routes/admin.js
pritishmishra: sure!

api-controllers/routes/admin.js
pritishmishra: okay..

data-access-layer/eventmesh/ApiServerClient.js
pritishmishra: This is a legacy code, which I got "inspired" from :)
It is present in numerous places in the same file.
As an example: https://github.com/cloudfoundry-incubator/service-fabrik-broker/blob/master/data-access-layer/eventmesh/ApiServerClient.js#L155

data-access-layer/eventmesh/ApiServerClient.js
jagadish-kb: both approaches lead to the same end result, though with 2 lesser lines of code per my comment.

data-access-layer/eventmesh/ApiServerClient.js
jagadish-kb: May be at the locations mentioned its missed earlier reviews. You can fix this instance and lead for a better inspiration  (may be to urself) when you add new methods to this class :-)

jobs/ServiceInstanceUpdateJob.js
jagadish-kb: @pritishmishra : hope these proposed changes are taken care in SFPC

common/constants.js
```diff
@@ -148,6 +148,10 @@ module.exports = Object.freeze({
     HUMAN_INTERVAL: 'human-readable-interval',
     CRON_EXPRESSION: 'cron_expression'
   },
+  CONFIG: {
+    DISABLE_SCHEDULED_UPDATE_CONFIG_PREFIX: 'disable_scheduled_update_',
+    RESOURCE_NAME: 'sfconfig'
```
jagadish-kb: can you name this const `RESOURCE_NAME` to `SF_CONFIG`? because down the line we might add another config & then we would have to name that constant something differently. Also, the following reads better `CONST.CONFIG.SF_CONFIG` as you are saying fetch a config param from SF_CONFIG.

kamath-prasad:
@jagadish-kb  : i just saw your approval and merged this PR .. hope the new comments are not must have :) .

PR #521, opened 2018-12-19
Adding extra field in deployment CRDs to keep track of current state in case of failure scenarios

jayeshjk:
* Currently we update desirable state in deployment resources during instance update, and Operators act to achieve this desirable state. However resource will continue to have this desirable state even if Operators experience failure during update operations.
* This is problematic since now there will be discrepancy between resource state and actual state of instance. (in case of update failures)
* To handle this, extra attribute called 'appliedOptions' is introduced in the CRDs, which will get updated only when create/update operations are successful. Therefore, 'appliedOptions' will continue to reflect current state of deployment even in case of update failures.

### DIFF
--- a/api-controllers/ServiceFabrikApiController.js
+++ b/api-controllers/ServiceFabrikApiController.js
@@ -147,8 +147,8 @@ class ServiceFabrikApiController extends FabrikBaseController {
         resourceId: req.params.instance_id
       })
       .then(resource => {
-        let resourceState = _.isEmpty(_.get(resource, 'status.actualState')) ? _.get(resource, 'spec.options') : 
-        _.get(resource, 'status.actualState');
+        let resourceState = _.isEmpty(_.get(resource, 'status.actualState')) ? _.get(resource, 'spec.options') :
+          _.get(resource, 'status.actualState');
         if (!_.get(req, 'body.context')) {
           _.set(req, 'body.context', _.get(resourceState, 'context'));
         }
--- a/api-controllers/ServiceFabrikApiController.js
+++ b/api-controllers/ServiceFabrikApiController.js
@@ -305,8 +305,8 @@ class ServiceFabrikApiController extends FabrikBaseController {
         resourceId: req.params.instance_id
       })
       .then(resource => {
-        const context = req.body.context || _.get(resource, 'spec.options.context');
-        const planId = req.body.plan_id || _.get(resource, 'spec.options.plan_id');
+        const context = req.body.context || _.get(resource, 'status.appliedOptions.context') || _.get(resource, 'spec.options.context');
+        const planId = req.body.plan_id || _.get(resource, 'status.appliedOptions.plan_id') || _.get(resource, 'spec.options.plan_id');
         const backupOptions = {
           guid: backupGuid,
           instance_guid: req.params.instance_id,
--- a/api-controllers/ServiceFabrikApiController.js
+++ b/api-controllers/ServiceFabrikApiController.js
@@ -299,6 +299,8 @@ class ServiceFabrikApiController extends FabrikBaseController {
   //TODO: Need to be revisited as these apis should be agnostic to resourceGroup and Type
 
   getBackupOptions(backupGuid, req) {
+    /* TODO: Conditional statements to fetch context and planId below is needed to be backwards compatible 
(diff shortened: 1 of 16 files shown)


broker/config/crds/deployment.servicefabrik.io_v1alpha1_dockers.yaml
```diff
@@ -35,6 +35,8 @@ spec:
               type: string
             state:
               type: string
+            actualState:
```
subhankarc: It is not `actualState` right? Rather `appliedOptions`.

jayeshjk: yes, that is correct. Will modify.

api-controllers/ServiceFabrikApiController.js
```diff
@@ -147,14 +147,16 @@ class ServiceFabrikApiController extends FabrikBaseController {
         resourceId: req.params.instance_id
       })
       .then(resource => {
+        let resourceState = _.isEmpty(_.get(resource, 'status.appliedOptions')) ? _.get(resource, 'spec.options') :
```
subhankarc: Why is the _.isEmpty check required?

api-controllers/ServiceFabrikApiController.js
```diff
@@ -147,14 +147,16 @@ class ServiceFabrikApiController extends FabrikBaseController {
...
+          _.get(resource, 'status.appliedOptions');
         if (!_.get(req, 'body.context')) {
-          _.set(req, 'body.context', _.get(resource, 'spec.options.context'));
+          _.set(req, 'body.context', _.get(resourceState, 'context'));
```
subhankarc: You are getting the options, so let's not call it state, let's call it resourceOptions.

api-controllers/ServiceFabrikApiController.js
subhankarc: A code comment here would help

operators/virtualhost-operator/VirtualHostOperator.js
```diff
@@ -69,7 +70,8 @@ class VirtualHostOperator extends BaseOperator {
         resourceId: changeObjectBody.metadata.name,
         status: {
           response: response,
-          state: CONST.APISERVER.RESOURCE_STATE.SUCCEEDED
+          state: CONST.APISERVER.RESOURCE_STATE.SUCCEEDED,
```
subhankarc: You have not changed the vhost CRD. This would fail then.

Better check it with @pranjaljain

broker/config/crds/deployment.servicefabrik.io_v1alpha1_virtualhosts.yaml
```diff
@@ -37,6 +37,8 @@ spec:
               type: string
             state:
               type: string
+            appliedOptions:
+              type: string
```
jayeshjk: @subhankarc is this modification in vhost CRD fine or is there any other CRD I need to modify? @pranjaljain

api-controllers/ServiceFabrikApiController.js
jayeshjk: Agreed. Will modify.

api-controllers/ServiceFabrikApiController.js
subhankarc: Ahh! I misread the braces.

broker/config/crds/deployment.servicefabrik.io_v1alpha1_virtualhosts.yaml
subhankarc: looks fine.

kamath-prasad:
@jayeshjk : i think you will need to handled it [here](https://github.com/cloudfoundry-incubator/service-fabrik-broker/blob/master/api-controllers/ServiceFabrikAdminController.js#L93) as well ?

jayeshjk:
@kamath-prasad Thanks for pointing this out. I think there are some other places in the codebase like [this](https://github.com/cloudfoundry-incubator/service-fabrik-broker/blob/master/api-controllers/DashboardController.js#L151), which will need modifications like this. I will handle them in this PR itself, as it seems feasible as of now.

PR #530, opened 2018-12-27
[WIP] Documentation for inter-operator architecture

amshuman-kr:
- [X] [Basic Architecture](https://github.com/amshuman-kr/service-fabrik-broker/blob/feature/inter-operator/interoperator/docs/architecture/basic.md)
- [x] [Cluster Landscape](https://github.com/amshuman-kr/service-fabrik-broker/blob/feature/inter-operator/interoperator/docs/architecture/cluster-landscape.md)
- [X] [Cluster Landscape Autoscaling](https://github.com/amshuman-kr/service-fabrik-broker/blob/feature/inter-operator/interoperator/docs/architecture/cluster-landscape-autoscaling.md)


interoperator/docs/architecture/basic.md
```diff
@@ -0,0 +1,493 @@
...
+This section presumes that the `SFService` and `sfplans` are already registered as describe [above](#catalog).
+
+![Service Fabrik Inter-operator Basic Control-flow Provision](images/basic-control-flow-provision.png)
+
+#### Service Fabrik Broker
+
+1. An OSB client makes a `provision` call to the [Service Manager](https://github.com/Peripli/service-manager).
+1. The Service Manager forwards the call (perhaps via some intermediaries) to Service Fabrik Broker if the `provision` call was for a service and plan that was published by the Service Fabrik Broker.
+The Service Manager adds some relevant additional context into the request.
+1. The Service Fabrik Broker creates an `SFServiceInstance` capturing all the details passed in the the `provision` request from the Service Manager.
```
jagadish-kb: I understand the need for SFServiceInstance CRD as this creates an intermediate OSB representation of the service. I have a suggestion which avoids this intermediate CRD. At individual service/plans we expect services to define the template (for both resource/status), so why not use that directly at the controller layers directly via the interoperator to create the service speicific CRD and also in lastOperation the status template to fetch the status back and return it? 
By doing so, we still achieve the end objective of not having services to adhere to OSB spec and additionally we completely eliminate the need for synching the resources/state between SFServiceInstance and service specific CRDs.
By doing so, Interoperator as a component will no longer be a controller but the functionality of it will be directly consumed in the OSB API controller layer (i.e. in SFBrokerAPIController layer directly).

kamath-prasad:
@amshuman-kr : can you pls rebase this PR ?

interoperator/docs/architecture/basic.md
```diff
@@ -0,0 +1,493 @@
...
+    providerDisplayName: 'SAP SE'
+    documentationUrl: 'https://sap.com/'
+    supportUrl: 'https://sap.com/'
+  dashboardClient:
+    id: postgresql-dashboard-client-id
+    secret: postgresql-dashboard-client-secret
+    redirectURI: 'https://sap.com/'
+  planUpdatable: true
+  context:
+    serviceFabrik:
```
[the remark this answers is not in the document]
amshuman-kr: I agree. @subhankarc could you please weigh in?

interoperator/docs/architecture/basic.md
```diff
@@ -0,0 +1,493 @@
...
+    longDescription: *description
+    providerDisplayName: 'SAP SE'
+    documentationUrl: 'https://sap.com/'
+    supportUrl: 'https://sap.com/'
+  dashboardClient:
+    id: postgresql-dashboard-client-id
+    secret: postgresql-dashboard-client-secret
+    redirectURI: 'https://sap.com/'
+  planUpdatable: true
+  context:
```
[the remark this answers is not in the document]
amshuman-kr: I guess, metadata is specific to OSB whereas context is something that may not be relevant for OSB?

@subhankarc could you please confirm?

interoperator/docs/architecture/basic.md
```diff
@@ -0,0 +1,493 @@
...
+The `url` field can be used to specify the location where the actual templates can be found. For example,
+
+```yaml
+apiVersion: osb.servicefabrik.io/v1alpha1
+kind: SFPlan
+spec:
+  templates:
+  - action: provision
+    type: gotemplate
+    url: "https://raw.githubusercontent.com/cloudfoundry-incubator/service-fabrik-broker/feature/inter-operator/interoperator/config/samples/templates/gotemplates/postgres/postgres.yaml"
```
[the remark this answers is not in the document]
amshuman-kr: Technically, URLs are fine if they are reachable from the private cloud. But not the URL I gave in the example, of course :-)

I will update document to mention your point. Thanks!

interoperator/docs/architecture/basic.md
amshuman-kr: The SFServiceInstance resource is serving two different purposes.
1. Decouple scheduling (which cluster to provision the instance in) from the actual provisioning of the instance.
   * Granted that the scheduling problem is relevant only in a multi-cluster landscape. But I think single-cluster is workable only for the very short term (especially, for something like Postgres which is liable to have large number of instances). I am adding some more content to the cluster-landscape to explain the details of the comparison of the different cluster landscapes and the justification for the recommendation for the optimal landscape (multi-cluster).
Such decoupling is required in multi-cluster scenario just to be able to schedule among a pool of available clusters. At the very least, there is a need to keep two separate CRDs.
      1. One CRD in the cluster where SF (broker and the inter-operator) is running. This is the CRD on which the scheduler would work.
      1. The other CRD in the actual cluster where the instance is provisioned. This is the CRD on which the service’s operator would act on.
   * It is a different matter if such a separation is best done by having the SFServiceInstance instance in the cluster where SF is running and the service’s CRD in the cluster where it is to be provisioned or if we can replicate the same service’s CRD in both locations. This is addressed below.
   * Such a decoupling is required to provide a stable identity for the service instance even while moving the actual provisioning from one cluster to another (if such a need arises).
1. Capture the details about the OSB request verbatim before it is processed by the inter-operator.
   *  Granted that this is not of immediately obvious value. But there is some value in capturing the request details as spec verbatim, especially, in debugging, tracing but also in reconciliation or even in update/upgrade logic.
   * Especially, when we anyway need two CRDs as discussed above, I think having SFServiceInstance that captures additional information is useful when compared to just replicating the service’s CRD.
   * It also makes SF fully stateful regarding all its instances.
   * There are additional advantages of providing a uniform interface to the scheduler (potentially custom schedulers) but that is a very minor advantage which we can even work without.
 
What do you think?

amshuman-kr: @jagadish-kb I have taken the liberty of cross posting from our email conversation here. I apologize if that is overstepping the bounds of decency.

> With regards to storing OSB request payload, one could always leverage annotations on service specific objects which are free flowing json.

Agreed. That was a weak argument. Only relevant in the presence of a different stronger argument as you agreed below.
 
> Decouple Scheduling is a strong reason, Yes I agree. I was of the impression that both the interoperator & services operator will be in same cluster. If they are in different cluster decoupling also helps in resiliency (Like the issue bosh-sf downtime today had issues with request processing in SF, which we had to address specifically). Also, apart from the other benefits you’ve mentioned additionally this decoupling could help us in streamlining some reporting requirements as well quite easily. My main concern was state synchronization, guess the above benefits now outweigh this concern.
Thanks for the clarification.

Thanks for both the counter-arguments as well as the eventual agreement and pointing our the additional benefits :-)

amshuman-kr:
> can you pls rebase this PR ?

@kamath-prasad I have rebased.

> Documentation generally goes under gh-pages branch or directly as WIKI pages. Its better they are not checked in along with code. Only Readme sections on getting the code up & running is checked in into the code branch. Thats a convention followed currently in SF. Would be, better if the same is as well followed here.

@jagadish-kb I am fine moving it to wiki pages. But review is easier as a PR on the project itself as shown by your very welcome review :-) I propose that once the review is fine I move it to wiki pages. What do you think?

interoperator/docs/architecture/cluster-landscape-autoscaling.md
```diff
@@ -0,0 +1,92 @@
...
+
+## Context
+
+The [context](basic.md#context) mentioned in the [basic architecture](basic.md) is applicable here.
+Also, the [context](cluster-landscape.md#context) mentioned in the [cluster landscape architecture](cluster-landscape.md) is very relevant here.
+Especially, the [recommendation](cluster-landscape.md#recommended-landscape-scenario) for the optimal [dedicated](cluster-landscape.md#an-optimal-dedicated-landscape-scenario) or [hybrid](cluster-landscape.md#hybrid-landscape-scenario) is relevant for scaling the number of Kubernetes clusters in a given landscape.
+
+## Motivation
+
+An [recommended](cluster-landscape.md#recommended-landscape-scenario) in the [cluster landscape architecture](cluster-landscape.md), a optimal [dedicated](cluster-landscape.md#an-optimal-dedicated-landscape-scenario) or [hybrid](cluster-landscape.md#hybrid-landscape-scenario) is the preferred landscape for the most common use-cases.
```
subhankarc: typo `As`

interoperator/docs/architecture/cluster-landscape.md
```diff
@@ -0,0 +1,332 @@
...
+  * [Managing the Cluster Landscape](#managing-the-cluster-landscape)
+    * [An analogy to kube\-scheduler](#an-analogy-to-kube-scheduler)
+    * [Service Instance Scheduling](#service-instance-scheduling)
+    * [A point to note about Service Instance Scheduling](#a-point-to-note-about-service-instance-scheduling)
+
+## Context
+
+The [context](basic.md#context) mentioned in the [basic architecture](basic.md) is applicable here.
+
+In addition to the above context, continuing the [operator design-pattern](https://github.wdf.sap.corp/CPonK8s/k8s-native-services-concept/blob/master/README.md#high-level-design-pattern) in Kubernetes, the individual service operators are expected to manage the service instances within a kubernetes cluster.
```
subhankarc: internal link!

amshuman-kr:
Closing this in favour of [PR#576](https://github.com/cloudfoundry-incubator/service-fabrik-broker/pull/576) which moves this documentation to gh-pages branch.

PR #547, opened 2019-01-17
Upgrade of deprecated libraries

pritishmishra:
Many of the libraries used by the service-fabrik-broker seem to be deprecated.

Some of these libraries/dependencies will be updated due to this change.
Most of the dependencies were child dependencies of the parent ones mentioned in the `package.json` of service-fabrik-broker. So, need to update the parent ones.

The entire list of update is long. This is one initial set of update of libraries, as we want to make these incrementally.


package.json
```diff
@@ -33,23 +33,23 @@
...
-    "azure-storage": "2.4.0",
+    "azure-arm-compute": "4.0.0",
+    "azure-storage": "2.8.3",
     "basic-auth": "1.0.4",
     "bluebird": "3.4.0",
     "body-parser": "1.15.1",
     "connect-timeout": "1.9.0",
     "dockerode": "2.2.10",
-    "etcd3": "0.2.9",
+    "etcd3": "0.2.10",
```
subhankarc: We don't use it anymore. Please check and remove this.

pritishmishra:
Dev pipeline is successful for the above change. Build no. 473.

package.json
pritishmishra: Done!

PR #550, opened 2019-01-18
Adding cross space restore capability

jayeshjk:
* This PR adds changes for enabling snapshot based restore across spaces of the same org. (PITR case is not addressed in this PR). (Related BL: HCPCFS-2162)
* Overall logic is as follows: ApiServer resource of the backup is used to find out the space details of the backup. If the source space (i.e., backup space) is different than the destination space (ie., instance space), then first the space developer permissions would be checked for the user in the source space also. 
* If the space developer verification is done successfully, then backup metadata file is fetched from the objectstorage using the source space guid and rest of the restore operation follows as it is.
* Apart from this, a small refactoring of adding/using method for space developer verification is also done as it is done in two different places now.


jayeshjk:
UTs are WIP.

api-controllers/ServiceFabrikApiController.js
```diff
@@ -195,28 +195,22 @@ class ServiceFabrikApiController extends FabrikBaseController {
...
-        }
-        logger.info('space develoopers done');
+        //TODO: Need to handle this separately for k8s consumption
+        return this.verifySpaceDevPermissions({
+            spaceGuid: space_guid,
+            user: user,
+            auth: opts
+          })
+          .then(isSpaceDeveloper => {
+            if (httpMethod !== 'GET' && !isSpaceDeveloper) {
```
sauravmndl: Use 'GET' from constant if possible though it is a legacy code.

api-controllers/ServiceFabrikApiController.js
```diff
@@ -540,6 +571,27 @@ class ServiceFabrikApiController extends FabrikBaseController {
           return this.validateUuid(backupGuid, 'Backup GUID');
         }
       })
+      .then(() => backupGuid ? this.getSpaceGuidOfBackup(backupGuid) : tenantId)
+      .then(spaceGuid => {
+        if (spaceGuid !== tenantId) {
+          backupSpaceGuid = spaceGuid;
+        }
+      })
+      .then(() => {
```
sauravmndl: L575-L579 is not required as discussed.

PR #561, opened 2019-01-24
Adding Metering Job

muralisc:
This pull request adds Metering Job
- Registers Metering CRD
- Metering job periodically sends metering documents to metering as a service
- Ability to exclude service and plans
- Webhook to create metering resources for lifecycle events Create Update , Delete


jobs/MeterInstanceJob.js
```diff
@@ -0,0 +1,186 @@
...
+  }
+
+  static meter(events) {
+    logger.info(`Number of events to be metered in this run - ${events.length}`);
+    return Promise.try(() => {
+      let successCount = 0,
+        failureCount = 0,
+        failedEvents = [];
+      return Promise.map(events, (event) => {
+          return this
```
sauravmndl: the explicit return is not required.

jobs/MeterInstanceJob.js
```diff
@@ -0,0 +1,186 @@
...
+const CONST = require('../common/constants');
+
+class MeterInstanceJob extends BaseJob {
+  static run(job, done) {
+    return Promise.try(() => {
+      logger.debug('Starting MeterInstanceJob Job');
+      this.getInstanceEvents()
+        .tap(events => logger.info('recieved events -> ', events))
+        .then(events => this.meter(events))
+        .then((meterResponse) => this.runSucceeded(meterResponse, job, done))
```
sauravmndl: Extra parenthesis not required  around `meterResponse `

data-access-layer/metering/MeteringClient.js
```diff
@@ -0,0 +1,69 @@
...
+        const serverResponse = JSON.parse(res.body);
+        return serverResponse.access_token;
+      })
+      .catch(err => {
+        logger.error('Error occurred while fetching metering auth token', err);
+        throw err;
+      });
+  }
+
+  putUsageRecord(usage_records) {
```
sauravmndl: Could you rename to 'updateUsageRecord' / 'sendUsageRecord' if possible, as apart from GET and DELETE function names are not generally named directly by http verb.

jobs/MeterInstanceJob.js
```diff
@@ -0,0 +1,185 @@
...
+    return false;
+  }
+
+  static enrichEvent(options) {
+    // Add region , service name and plan sku name of the event
+    options = JSON.parse(options);
+    const serviceId = _.get(options, 'service.id');
+    const planId = _.get(options, 'service.plan');
+    logger.info(`Enriching the metering event ${serviceId}, plan: ${planId}`);
+    options.service.id = this.getServiceNameFromServiceGUID(serviceId);
```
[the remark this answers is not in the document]
sauravmndl: Also Can't you use directly https://github.com/cloudfoundry-incubator/service-fabrik-broker/blob/master/common/models/catalog.js#L37 for the desired info.

jobs/MeterInstanceJob.js
```diff
@@ -0,0 +1,186 @@
...
+
+  static getServiceNameFromServiceGUID(serviceGuid) {
+    return _.chain(catalog.toJSON().services)
+      .map((s) => s.id === serviceGuid ? s.name : undefined)
+      .filter(s => s !== undefined)
+      .head()
+      .value();
+  }
+
+  static getPlanSKUFromPlanGUID(serviceGuid, planGuid) {
```
sauravmndl: Probably you can utilise some of the methods from catalog.js and fetch the info.

broker/config/settings.yml
```diff
@@ -342,6 +349,21 @@ defaults: &defaults
...
+  #####################
+  # METERING SETTINGS #
+  #####################
+  metering:
+    client_id: 'metering_client_id'
+    client_secret: 'metering_client_secret'
+    token_url: 'https://service-fabrik.authentication.domain.com'
+    metering_url: 'https://metering.cfapps.domain.com/dev'
+    region: "asia"
+    excluded_service_plans: [
```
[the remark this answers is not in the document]
subhankarc: +1 ...... the settings should be part of plan metadata.

common/utils/HttpClient.js
```diff
@@ -182,8 +182,8 @@ class HttpClient {
           try {
             const errResponse = JSON.parse(body);
             err.error = errResponse;
-          } catch (err) {
-            logger.info('Error occurred while parsing http response- ', err.error);
+          } catch (parseErr) {
```
subhankarc: `parseErr` ?

data-access-layer/metering/MeteringClient.js
```diff
@@ -0,0 +1,69 @@
...
+
+  constructor(options) {
+    super(_.defaultsDeep({
+      baseUrl: config.metering.metering_url,
+      headers: {
+        Accept: 'application/json',
+      },
+      followRedirect: false
+    }, options));
+    this.clientid = config.metering.client_id;
```
subhankarc: use camelcase. `clientId`, `clientSecret`, `tokenUrl`, `meteringUrl`

data-access-layer/metering/MeteringClient.js
subhankarc: `usageRecords` ?

subhankarc: Also, I am not sure what the method does. Does it do start and stop metering both?

jobs/MeterInstanceJob.js
```diff
@@ -0,0 +1,186 @@
...
+          };
+        })
+        .catch(err => {
+          logger.error(`Error occurred while metering all events : `, err);
+        });
+
+    });
+  }
+
+  static isServicePlanExcluded(options) {
```
subhankarc: Such a complex logic will not be needed if we have the metadata part in the individual plan...right?

jobs/MeterInstanceJob.js
subhankarc: also, u could enhance `catalog.js` to have these functions. Having these here does not seem correct.

jobs/MeterInstanceJob.js
```diff
@@ -0,0 +1,186 @@
...
+    const serviceId = _.get(options, 'service.id');
+    const planId = _.get(options, 'service.plan');
+    logger.info(`Enriching the metering event ${serviceId}, plan: ${planId}`);
+    options.service.id = this.getServiceNameFromServiceGUID(serviceId);
+    options.service.plan = this.getPlanSKUFromPlanGUID(serviceId, planId);
+    options.consumer.region = config.metering.region;
+    return options;
+  }
+
+  static getServiceNameFromServiceGUID(serviceGuid) {
```
subhankarc: You could enhance `catalog.js` to have these functions. Having these here does not seem correct.

jobs/MeterInstanceJob.js
```diff
@@ -0,0 +1,186 @@
...
+        logger.info(`The metering event guid: ${options.id} with ${serviceName}, ${planName} is excluded`);
+        return true;
+      }
+    }
+    return false;
+  }
+
+  static enrichEvent(options) {
+    // Add region , service name and plan sku name of the event
+    options = JSON.parse(options);
```
subhankarc: How do we store data in options? Can we make them objects so that we do not need to parse them?

jobs/MeterInstanceJob.js
```diff
@@ -0,0 +1,186 @@
...
+    const serviceId = _.get(options, 'service.id');
+    const planId = _.get(options, 'service.plan');
+    logger.info(`Enriching the metering event ${serviceId}, plan: ${planId}`);
+    options.service.id = this.getServiceNameFromServiceGUID(serviceId);
+    options.service.plan = this.getPlanSKUFromPlanGUID(serviceId, planId);
```
subhankarc: You are getting `planId` from `options.service.plan` and then storing sku in it. Does not seem right. Can you not have an attribute like `option.service.plan.sku` ?

jobs/MeterInstanceJob.js
```diff
@@ -0,0 +1,186 @@
...
+
+  static sendEvent(event) {
+    logger.info('Sending event:', event);
+    if (this.isServicePlanExcluded(event.spec.options)) {
+      return Promise.try(() => this.updateMeterState(CONST.METER_STATE.EXCLUDED, event))
+        .return(true);
+    }
+    return Promise
+      .try(() => this.enrichEvent(event.spec.options))
+      .then(enriched_usage_doc => {
```
subhankarc: use camelcase

jobs/MeterInstanceJob.js
```diff
@@ -0,0 +1,186 @@
...
+        logger.debug('Sending document:', enriched_usage_doc);
+        return maas.client.putUsageRecord({
+          usage: [enriched_usage_doc]
+        });
+      })
+      .then(validEvent => validEvent ? this.updateMeterState(CONST.METER_STATE.METERED, event) : false)
+      .return(true)
+      .catch(err => {
```
subhankarc: What if `maas.client.putUsageRecord` worked but `this.updateMeterState` failed. Then ? :)

operators/StartMeterOperator.js
```diff
@@ -0,0 +1,7 @@
+'use strict';
+
+const MeterOperator = require('./meter-operator/MeterOperator');
```
subhankarc: Where is this `MeterOperator` ? Could not find in this PR? Am I missing something?

broker/config/settings.yml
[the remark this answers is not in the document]
muralisc: How can we restrict it per landscape ? if we have this setting on plan metadata.
The requirement from @ashishjain14 was to have ability to meter specific plans per landscape.

common/utils/HttpClient.js
muralisc: There are two err here . `err` is defined in line 141. parseErr can occur in 183

data-access-layer/metering/MeteringClient.js
muralisc: Hi please see:
https://wiki.wdf.sap.corp/wiki/display/CPC15N/Developer+Guide+for+Metering-as-a-Service
section `Usage Recording`
This is the API endpoint to send usage records which can have both stop and start.
`usageRecords` since we can send an object with multiple usages

jobs/MeterInstanceJob.js
muralisc: How can we achieve landscape specific exclusion ?

jobs/MeterInstanceJob.js
muralisc: options are populated in webhook. 
Webhook has no knowledge of catalog, but only of the current CRD on which it is operating. 
So we store the plan guid and change it to sku before sending.

jobs/MeterInstanceJob.js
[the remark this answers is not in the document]
muralisc: Please see: 
https://wiki.wdf.sap.corp/wiki/display/CPC15N/Usage+Document+Detailed+Schema
The usage send to MaaS has object:
```
service (object)
 - id  // the name with which the service is registered eg postgresql, redis etc
 - plan // sku name 
```
We populate the guid in webhook and enrich it in the broker before sending.

common/utils/HttpClient.js
muralisc: This is required since metering responds with error message in `body`. Hence the parsing will fail in line 183

jobs/MeterInstanceJob.js
```diff
@@ -0,0 +1,186 @@
...
+              } else {
+                failureCount++;
+                failedEvents.push(event);
+              }
+            })
+            .catch(err => {
+              logger.error(`Error occurred while metering event : `, err);
+            });
+        }).then(() => {
+          return {
```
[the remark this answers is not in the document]
muralisc: unfortunately it doesn't e.g:
```
function logger(callback) {
  console.log(callback())
}

logger(() => { return { key: "value", key1: "value1" }});
// logger(() => { key: "value", key1: "value1" } ); // throws syntax error
```

jobs/MeterInstanceJob.js
muralisc: Currently the job would retry the same document again in the next run.
Since we have unique ids for each metering doc it will simply result in updating the same entry.

Do we need to add further error handling here ?

common/utils/HttpClient.js
subhankarc: ok.

data-access-layer/metering/MeteringClient.js
subhankarc: ah! Thanks!

jobs/MeterInstanceJob.js
subhankarc: if retry is not a harm, then it is ok. Can you put a comment here about this discussion?

jobs/MeterInstanceJob.js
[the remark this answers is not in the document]
kamath-prasad: ok , they have used slightly confusing names..

Just a question/suggestion : With this then we are tying SF to SAP based metering ? Can we introduce some flexibility here such that we can have some templates and we populate the data based on the same ? Does it make sense ? Or will Metering2.0 be opensource standard for metering service instances..

jobs/MeterInstanceJob.js
muralisc: Changed options to objects

jobs/MeterInstanceJob.js
```diff
@@ -0,0 +1,163 @@
+'use strict';
```
amitmalav: @abh1kg Have we standardised using `async` `await`. I remember discussing that for new files we shall start using async await. This will be a good start IMO

abh1kg: Agreed- in fact, I feel we should use async-await for all cases. The code becomes much cleaner and easier to skim through

jobs/MeterInstanceJob.js
muralisc: Since we decided to use plan level exclusion the code is modified as below

jobs/MeterInstanceJob.js
muralisc: Updated to use async/await 👍

broker/config/settings.yml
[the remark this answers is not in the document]
muralisc: Changed to use `metered: <true/false>` in plan metadata

jobs/MeterInstanceJob.js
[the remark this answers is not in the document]
muralisc: sure, now since we have Interoperator,  we could have metering data as part of the sfInstance CRD (part of interoperator) and then  interoperator can make metering specific CRDs . This way any other metering service can bring their templates for interoperator and use the metering data already present in SFInstance.

@subhankarc Does the above approach make sense ?

jobs/MeterInstanceJob.js
```diff
@@ -0,0 +1,185 @@
...
+    options = JSON.parse(options);
+    const serviceId = _.get(options, 'service.id');
+    const planId = _.get(options, 'service.plan');
+    logger.info(`Enriching the metering event ${serviceId}, plan: ${planId}`);
+    options.service.id = this.getServiceNameFromServiceGUID(serviceId);
+    options.service.plan = this.getPlanSKUFromPlanGUID(serviceId, planId);
+    options.consumer.region = config.metering.region;
+    return options;
+  }
+
```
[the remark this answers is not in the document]
muralisc: done

jobs/MeterInstanceJob.js
muralisc: Moved to catalog

jobs/MeterInstanceJob.js
muralisc: Done !!

jobs/MeterInstanceJob.js
```diff
@@ -0,0 +1,163 @@
...
+    return _
+      .chain(service.plans)
+      .map((p) => p.id === planGuid ? p.sku_name : undefined)
+      .filter(p => p !== undefined)
+      .head()
+      .value();
+  }
+
+  static sendEvent(event) {
+    logger.info('Sending event:', event);
```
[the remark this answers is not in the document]
muralisc: done

jobs/MeterInstanceJob.js
```diff
@@ -0,0 +1,163 @@
...
+    if (this.isServicePlanExcluded(event.spec.options)) {
+      return Promise.try(() => this.updateMeterState(CONST.METER_STATE.EXCLUDED, event))
+        .return(true);
+    }
+    return Promise
+      .try(() => this.enrichEvent(event.spec.options))
```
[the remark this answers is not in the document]
muralisc: done

data-access-layer/metering/MeteringClient.js
```diff
@@ -0,0 +1,70 @@
...
+    }, options));
+    this.clientId = config.metering.client_id;
+    this.clientSecret = config.metering.client_secret;
+    this.tokenUrl = config.metering.token_url;
+    this.meteringUrl = config.metering.metering_url;
+  }
+
+  async getAuthToken() {
+    try {
+      let res = await this
```
abh1kg: Minor comment: Use `const` instead of `let` since the reference is not being changed

data-access-layer/metering/MeteringClient.js
```diff
@@ -0,0 +1,70 @@
...
+        }, 200);
+      const serverResponse = JSON.parse(res.body);
+      return serverResponse.access_token;
+    } catch (err) {
+      logger.error('Error occurred while fetching metering auth token', err);
+      throw err;
+    }
+  }
+
+  async sendUsageRecord(usage_records) {
```
abh1kg: Minor comment: Keep the same variable naming format for arguments as well as the local variable declarations. instead of `usage_records`, `usageRecords` is more standard

data-access-layer/metering/MeteringClient.js
```diff
@@ -0,0 +1,70 @@
...
+    try {
+      let accessToken = await this.getAuthToken();
```
abh1kg: Question: wouldn't you want to cache the access token? Generally, multiple token fetch requests increase load on the token endpoint and can lead to failures.

jobs/MeterInstanceJob.js
```diff
@@ -0,0 +1,158 @@
...
+const CONST = require('../common/constants');
+
+class MeterInstanceJob extends BaseJob {
+
+  /* jshint ignore:start */
+  static async run(job, done) {
+    try {
+      logger.info(`-> Starting MeterInstanceJob -  name: ${job.attrs.data[CONST.JOB_NAME_ATTRIB]} - with options: ${JSON.stringify(job.attrs.data)} `);
+      let events = await this.getInstanceEvents()
+      logger.info('recieved events -> ', events)
```
abh1kg: `received` :-)

jobs/MeterInstanceJob.js
```diff
@@ -0,0 +1,158 @@
...
+const apiServerClient = require('../data-access-layer/eventmesh').apiServerClient;
+const CONST = require('../common/constants');
+
+class MeterInstanceJob extends BaseJob {
+
+  /* jshint ignore:start */
+  static async run(job, done) {
+    try {
+      logger.info(`-> Starting MeterInstanceJob -  name: ${job.attrs.data[CONST.JOB_NAME_ATTRIB]} - with options: ${JSON.stringify(job.attrs.data)} `);
+      let events = await this.getInstanceEvents()
```
abh1kg: Minor comment: Use `const` instead of `let` since the reference is not being changed

jobs/MeterInstanceJob.js
```diff
@@ -0,0 +1,158 @@
...
+      logger.info('recieved events -> ', events)
+      let meterResponse = await this.meter(events)
```
abh1kg: Minor comment: Use `const` instead of `let` since the reference is not being changed

jobs/MeterInstanceJob.js
```diff
@@ -0,0 +1,158 @@
...
+    return apiServerClient.getResources(options);
+  }
+
+  /* jshint ignore:start */
+  static async meter(events) {
+    try {
+      logger.info(`Number of events to be metered in this run - ${events.length}`);
+      // Adding this comment as we are transitioning to async/await
+      // Note: The below Promise is not bluebird promise
+      let promiseArray = await Promise.all(_.map(events, async (event) =>
```
abh1kg: Minor comment: Use `const` instead of `let` since the reference is not being changed

jobs/MeterInstanceJob.js
```diff
@@ -0,0 +1,158 @@
...
+        await this.sendEvent(event)));
+      let successCount = promiseArray.filter(r => r === true).length;
```
abh1kg: Minor comment: Use `const` instead of `let` since the reference is not being changed

jobs/MeterInstanceJob.js
```diff
@@ -0,0 +1,158 @@
...
+      if (this.isServicePlanExcluded(event.spec.options) === true) {
+        await this.updateMeterState(CONST.METER_STATE.EXCLUDED, event);
+        return true;
+      }
+      let enrichedUsageDoc = await this.enrichEvent(_.get(event.spec,'options'));
+      logger.info('Sending enriched document:', enrichedUsageDoc);
+      let validEvent = await maas.client.sendUsageRecord({
+        usage: [enrichedUsageDoc]
+      });
+      validEvent ? this.updateMeterState(CONST.METER_STATE.METERED, event) : false;
```
abh1kg: I feel it would be better to use standard `if-else` instead of ternary here for better readability

jobs/MeterInstanceJob.js
```diff
@@ -0,0 +1,158 @@
...
+
+  /* jshint ignore:start */
+  static async sendEvent(event) {
+    try {
+      logger.debug('Metering Event details before enriching:', event);
+      if (this.isServicePlanExcluded(event.spec.options) === true) {
+        await this.updateMeterState(CONST.METER_STATE.EXCLUDED, event);
+        return true;
+      }
+      let enrichedUsageDoc = await this.enrichEvent(_.get(event.spec,'options'));
```
abh1kg: Minor comment: Use `const` instead of `let` since the reference is not being changed

jobs/MeterInstanceJob.js
```diff
@@ -0,0 +1,158 @@
...
+      logger.info('Sending enriched document:', enrichedUsageDoc);
+      let validEvent = await maas.client.sendUsageRecord({
```
abh1kg: Minor comment: Use `const` instead of `let` since the reference is not being changed

jobs/MeterInstanceJob.js
```diff
@@ -0,0 +1,158 @@
...
+        usage: [enrichedUsageDoc]
+      });
+      validEvent ? this.updateMeterState(CONST.METER_STATE.METERED, event) : false;
+      return true;
+    } catch (err) {
+      logger.error('Error occured while metering:', err);
+      await this._logMeteringEvent(err, event);
```
abh1kg: Since the method called is `static`, its better to invoke it on the class name rather than the instance using `this`

jobs/MeterInstanceJob.js
```diff
@@ -0,0 +1,158 @@
...
+      await this.updateMeterState(CONST.METER_STATE.FAILED, event, err);
+      return false;
+    }
+  }
+  /* jshint ignore:end */
+
+  static _logMeteringEvent(err, event) {
+    let now = new Date();
+    let secondsSinceEpoch = Math.round(now.getTime() / 1000);
```
abh1kg: Minor comment: Use `const` instead of `let` since the reference is not being changed

data-access-layer/metering/MeteringClient.js
muralisc: Done

data-access-layer/metering/MeteringClient.js
muralisc: Done

jobs/MeterInstanceJob.js
muralisc: Done

jobs/MeterInstanceJob.js
muralisc: Done

jobs/MeterInstanceJob.js
ketakig12: Hope this won't be another logs filling info. Else please put it in debug.

jobs/MeterInstanceJob.js
```diff
@@ -0,0 +1,160 @@
...
+      .filter(s => s !== undefined)
+      .head()
+      .value();
+    return _
+      .chain(service.plans)
+      .map((p) => p.id === planGuid ? p.sku_name : undefined)
+      .filter(p => p !== undefined)
+      .head()
+      .value();
+  }
```
ketakig12: I think you have moved it to [catalog](https://github.com/cloudfoundry-incubator/service-fabrik-broker/pull/561/files#diff-55c660028f2fa662d1ac1a23277c1213R47). Then can you remove from here?

data-access-layer/metering/MeteringClient.js
```diff
@@ -0,0 +1,72 @@
...
+    } catch (err) {
+      logger.error('Error occurred while fetching metering auth token', err);
+      throw err;
+    }
+  }
+
+  async sendUsageRecord(usageRecords) {
+    try {
+      if (this.accessToken === undefined) {
+        this.accessToken = await this.getAuthToken();
```
ketakig12: I think you will need to check if token is expired, if expired, refresh and cache again.

jobs/MeterInstanceJob.js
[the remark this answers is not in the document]
abh1kg: just to add, if you want to avoid a `return`, you can use the new ES6 way for functional composition:

e.g.

```
.then(() => ({key1: 'value1', key2: 'value2'})
```

webhooks/pkg/webhooks/manager/event.go
```diff
@@ -0,0 +1,284 @@
...
+	if err != nil {
+		return false, err
+	}
+	isPlanChanged, err := e.isPlanChanged()
+	if err != nil {
+		return false, err
+	}
+	if e.isDirector() && e.isStateChanged() {
+		if e.isSucceeded() {
+			return (isUpdate && isPlanChanged) || isCreate, nil
```
ketakig12: Since we have seen many cases, where last_operation type / appliedOptions is not populated in error cases, maybe it is good idea to move init of ```isUpdate```, ```isCreate``` and ```isPlanChanged``` variables here after ```isSucceeded``` check.

data-access-layer/metering/MeteringClient.js
muralisc: Thanks :)

data-access-layer/metering/MeteringClient.js
muralisc: Yes, thanks

jobs/MeterInstanceJob.js
```diff
@@ -0,0 +1,145 @@
...
+    } catch (err) {
+      logger.error('Error occured while metering:', err);
+      await MeterInstanceJob._logMeteringEvent(err, event);
+      await this.updateMeterState(CONST.METER_STATE.FAILED, event, err);
+      return false;
+    }
+  }
+  /* jshint ignore:end */
+
+  static _logMeteringEvent(err, event) {
```
jagadish-kb: shouldnt this method have asynch in it since its returning a promise? or asynch is to be added only to those methods which have await stmts in them and not returning a promise?

jobs/MeterInstanceJob.js
```diff
@@ -0,0 +1,145 @@
...
+
+  static updateMeterState(status, event, err) {
+    logger.debug(`Updating meter state to ${status} for event`, event);
+    let status_obj = {
+      state: status
+    };
+    if (err !== undefined) {
+      status_obj.error = utils.buildErrorJson(err);
+    }
+    return apiServerClient.updateResource({
```
jagadish-kb: better to do this in a retry block here.

jobs/MeterInstanceJob.js
```diff
@@ -0,0 +1,145 @@
...
+      const validEvent = await maas.client.sendUsageRecord({
+        usage: [enrichedUsageDoc]
+      });
+      if (validEvent !== undefined) {
+        this.updateMeterState(CONST.METER_STATE.METERED, event);
+      }
+      return true;
+    } catch (err) {
+      logger.error('Error occured while metering:', err);
+      await MeterInstanceJob._logMeteringEvent(err, event);
```
jagadish-kb: Are we logging every meter instance failure attempt to Riemann? What will be the strategy to raise AVS alerts? Do we flag each failure to update as downtime?

data-access-layer/metering/MeteringClient.js
```diff
@@ -0,0 +1,74 @@
...
+      return this.request({
+        url: CONST.URL.METERING_USAGE,
+        method: CONST.HTTP_METHOD.PUT,
+        auth: {
+          bearer: this.tokenInfo.accessToken
+        },
+        body: usageRecords,
+        json: true
+      }, CONST.HTTP_STATUS_CODE.OK);
+    } catch (err) {
```
jagadish-kb: catch block is not adding any value rather than just throwing the error back. Can remove it. The invoker anyways has a try/catch so this log will also be redundant.

jobs/MeterInstanceJob.js
muralisc: We need `async` in definition only if we are awaiting for a promise inside the function.
Here we don't want to wait for the promise to complete before returning it

data-access-layer/metering/MeteringClient.js
```diff
@@ -0,0 +1,74 @@
...
+            pass: this.clientSecret
+          },
+          qs: {
+            grant_type: 'client_credentials'
+          }
+        }, 200);
+      const serverResponse = JSON.parse(res.body);
+      return serverResponse.access_token;
+    } catch (err) {
+      logger.error('Error occurred while fetching metering auth token', err);
```
jagadish-kb: catch block not adding any value, just rethrowing it back and it leads to redundant logging of error. Can be removed.

jobs/MeterInstanceJob.js
muralisc: `_logMeteringEvent` has internal logic to raise events only on failures that occur after `thresholdHours`. Hence we can raise all Riemann logs as alerts.
As long as the instance is usable we need not flag it as downtime.

data-access-layer/metering/MeteringClient.js
muralisc: Sure 👍

data-access-layer/metering/MeteringClient.js
muralisc: Done

data-access-layer/metering/MeteringClient.js
```diff
@@ -0,0 +1,74 @@
...
+      throw err;
+    }
+  }
+
+  async sendUsageRecord(usageRecords) {
+    try {
+      if (this.tokenInfo.expiresSoon(this.tokenInfo.accessToken) == true) {
```
abh1kg: you can remove the `== true` or if you want to keep it, have strict equals `===`

jobs/MeterInstanceJob.js
muralisc: Added 👍

data-access-layer/metering/MeteringClient.js
muralisc: Done

PR #562, opened 2019-01-24
Minor Fixes for SF2.1

amitmalav:
* Don't create namespace during createBinding
* Minor fixes in bind and status template
* Update secret if already exist in createBind
* Compare subfields of spec and data in reconcile
* Test cases


kamath-prasad:
merging as already approved.

PR #564, opened 2019-01-25
Define multi az downgrade service flow definition

jagadish-kb:
Earlier as part of service flows implementation only the flow for upgrade to multi-az was defined. As part of the below PR, we are defining the flow definition for bringing the deployment back to single az. In order to activate this flow user needs to input the -c param 'multi_az' with false as the value,  as part of the instance update operation.

ex: 
```
cf update-service bleuprint01  -c '{"multi_az" : false}'
```


kamath-prasad:
Dev pipeline 536 build # green... and the coverage reduction is not related to this PR... merging..

PR #565, opened 2019-01-28
Fixing json-fying of response

pritishmishra:
Earlier one of the admin api-endpoints, responsible for creating or updating config, used to return a message in the response. However, this was not in json format. 
This was breaking some of other script utilities.
All other responses sent from broker are in json format, so, to maintain the convention and fix the failures of scripts, returning the response in json format, in this change request.

### DIFF
--- a/api-controllers/ServiceFabrikAdminController.js
+++ b/api-controllers/ServiceFabrikAdminController.js
@@ -758,9 +758,12 @@ class ServiceFabrikAdminController extends FabrikBaseController {
       key: req.query.key,
       value: req.query.value
     };
+    const body = {
+      message: `Created/Updated ${req.query.key} with value ${req.query.value}`
+    };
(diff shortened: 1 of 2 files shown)


pritishmishra:
Coverage reduction is not related to the changes introduced in this PR.

api-controllers/ServiceFabrikAdminController.js
```diff
@@ -758,9 +758,12 @@ class ServiceFabrikAdminController extends FabrikBaseController {
       key: req.query.key,
       value: req.query.value
     };
+    const body = {
+      message: `Created/Updated ${req.query.key} with value ${req.query.value}`
+    };
     return eventmesh.apiServerClient.createUpdateConfigMapResource(CONST.CONFIG.RESOURCE_NAME, config)
       .then(() => {
-        return res.status(201).send(`Created/Updated ${req.query.key} with value ${req.query.value}`);
+        return res.status(201).send(body);
```
abh1kg: Rather than doing this, why didn't you do `res.status(201).json(body)`?

PR #570, opened 2019-01-30
Enable sharing instances of Docker dev plans as well

abh1kg:
Instance sharing with group creation was only allowed for productive plans earlier.
With this change, sharing of dev instances is also supported i.e. target space apps
can now connect to dev instances if shared.


abh1kg:
@kamath-prasad : Dev pipeline run was successful. Will run another pipeline with the integration tests.
Merging this

PR #573, opened 2019-02-01
Restructuring travis job

subhankarc:
* Trufflehog was removed long back. Bringing it back now.
* Introducing multiple jobs in parallel for shorter execution time
* Keeping longer jobs in the front so that total time is reduced.


kamath-prasad:
@subhankarc  : can we merge this ?

PR #582, opened 2019-02-07
Add configurable indentity_provider

amitmalav:
* If `cf.indentity_provider` is `undefined` then `login_hint` property is not sent in body of `/oauth/token` request

### DIFF
--- a/broker/config/settings.yml
+++ b/broker/config/settings.yml
@@ -380,6 +380,7 @@ defaults: &defaults
     url: https://api.bosh-lite.com
     username: admin
     password: admin
+    identity_provider: uaa
   
   ###################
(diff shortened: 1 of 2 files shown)


abh1kg:
I saw this late- with this, the password OAuth token flow is now customized according to whether the technical user is present in the UAA or in LDAP.
On the other hand, wouldn't you foresee that we forego the password grant flow altogether and use user-token workflow from UAA? This would anyway be needed if we attempt Kubernetes later

kamath-prasad:
@abh1kg : Yes , we can pursue that .. For now this change was required as this breaks the flow for brokers like ABAP whose CF user will use LDAP only..

PR #584, opened 2019-02-07
fix Rsyslog-Error 2077

Espe0n:
We are receiving Error-Messages regarding the Service-Fabrik.

```
Could not create tcp listener, ignoring port 514 bind-address (null). [v8.1901.0 try https://www.rsyslog.com/e/2077 ]
````

The syslog configuration currently will attempt to create a TCP-Listener for Port 514. 

https://github.com/cloudfoundry-incubator/service-fabrik-boshrelease/blob/master/jobs/service-fabrik-broker/templates/config/syslog_forwarding.conf.erb#L6

However, this port 514 is located in the privileged area for which higher privileges are needed.
For safety reasons, the new version of Rsyslog will drop the privileges to a standard User. 
See changelog Rsyslog 4.1.1

> Dropping Privileges 
> Starting with 4.1.1, rsyslogd provides the ability to drop privileges by impersonating as another user and/or group after startup.
> 
> Please note that due to POSIX standards, rsyslogd always needs to start up as root if there is a listener who must bind to a network port below 1024. For example, the UDP listener usually needs to listen to 514 and as such rsyslogd needs to start up as root.
> 
> If you do not need this functionality, you can start rsyslog directly as an ordinary user. That is probably the safest way of operations. However, if a startup as root is required, you can use the $PrivDropToGroup and $PrivDropToUser config directives to specify a group and/or user that rsyslogd should drop to after initialization. Once this happend, the daemon runs without high privileges (depending, of course, on the permissions of the user account you specified).
> 

It works at the moment, because the port is already created in the master configuration (/etc/Rsyslog.conf) before the permissions are dropped.

See:
https://github.com/rsyslog/rsyslog/blob/69f8e1d1f7fe62fd2c5f38a81d4102a9a62d1722/grammar/debian.conf#L21

This PR will reallocate the Listening-Port from 514 to 1514.

NOTE: Two-part PR
Please only merge this PR if the second PR ( https://github.com/cloudfoundry-incubator/service-fabrik-boshrelease/pull/159 ) can also be merged.

- [x] This pull request is ready to be merged
- [x] The depending pull request is ready to be merged

### DIFF
--- a/common/logger.js
+++ b/common/logger.js
@@ -28,6 +28,8 @@ const transports = [
   new winston.transports.Syslog({
     level: config.sys_log_level || 'info',
     protocol: 'tcp4',
+    port: '1514',
+    app_name: config.broker_name,
     eol: '\n',
     formatter: (options) => `[${config.broker_name}] ${options.level.toUpperCase()}  ${options.message || ''}`
   })

### CODE REVIEW

File: common/logger.js [hunk 1]
```diff
@@ -28,6 +28,7 @@ const transports = [
   new winston.transports.Syslog({
     level: config.sys_log_level || 'info',
     protocol: 'tcp4',
+    port: '1514',
```
subhankarc: Please add the app_name property as well.

Espe0n: Hi , I used the `config.broker_name` for the `app_name` as suggested in the email.
Best Regards

subhankarc: Thanks @Espe0n 

Can you please take care of the boshrelease comments as well?

PR #585, opened 2019-02-08
Make hostname in the docker credentials mandatory

subhankarc:
During docker bind, sometimes if the docker host is down, bind does not fail but the credential object does not contain host information which makes it unusable. In this fix, we would fail the binding if hostname is missing

### DIFF
--- a/operators/docker-operator/DockerService.js
+++ b/operators/docker-operator/DockerService.js
@@ -502,7 +502,9 @@ class DockerService extends BaseService {
 
   createCredentials() {
     const networkInfo = this.getNetworkInfo(this.containerInfo);
-    return this.credentials.create(this.getEnvironment(), networkInfo.ip, networkInfo.ports);
+    const creds = this.credentials.create(this.getEnvironment(), networkInfo.ip, networkInfo.ports);
+    assert.ok(creds.hostname, `Attribute 'hostname' is missing in credentials, could be if your container or host is down`);
+    return creds;
   }
 
   ensureContainerIsRunning(removeVolumes) {


operators/docker-operator/DockerService.js
```diff
@@ -502,7 +502,9 @@ class DockerService extends BaseService {
 
   createCredentials() {
     const networkInfo = this.getNetworkInfo(this.containerInfo);
-    return this.credentials.create(this.getEnvironment(), networkInfo.ip, networkInfo.ports);
+    const creds = this.credentials.create(this.getEnvironment(), networkInfo.ip, networkInfo.ports);
+    assert.ok(creds.hostname, `Attribute 'hostname' is missing in credentials, could be if your container or host is down`);
```
jagadish-kb: are there any other reasons why host name could be undefined other than host being down? This is a container info call to swarm manager which should ideally return even if container is down na? (Just a thought as i thought swarm manager also maintains an internal state)

subhankarc: Currently if the docker vm is not reachable or is undergoing stemcell update , then the bind request sends back incomplete information to the user and hence there are lot of issues reported.

Below is what I could figure our from the code and logs :

In DockerService.bind, we call createCredentials

```javascript
  bind(params) {
    /* jshint unused:false */
    let credentials;
    return this.platformManager.preBindOperations({
        context: params.context,
        bind_resource: params.bind_resource,
        bindingId: params.binding_id
      })
      .then(() => this.inspectContainer())
      .catchThrow(DockerError.NotFound, new ServiceInstanceNotFound(this.guid))
      .then(() => this.createCredentials())
      .tap(creds => {
        credentials = creds;
      })
      .then(() => this.platformManager.postBindOperations({
        context: params.context,
        bind_resource: params.bind_resource,
        bindingId: params.binding_id,
        ipRuleOptions: this.buildIpRules()
      }))
      .then(() => credentials);
  }

```
 
In createCredentials,
 
```javascript
  createCredentials() {
    const networkInfo = this.getNetworkInfo(this.containerInfo);
    return this.credentials.create(this.getEnvironment(), networkInfo.ip, networkInfo.ports);
  }
```
 
 
Now, credentials.create does the following
 
```javascript
  create(environment, hostname, ports) {
    const credentials = {
      hostname: hostname
    };
    const portValues = _.values(ports);
    if (portValues.length) {
      credentials.ports = ports;
    }
    if (this.uriPort) {
      credentials.port = _.get(ports, this.uriPort);
    } else if (portValues.length === 1) {
      credentials.port = _.first(portValues);
    }
    if (this.usernameKey) {
      credentials.username = _.get(environment, this.usernameKey);
    }
    if (this.passwordKey) {
      credentials.password = _.get(environment, this.passwordKey);
    }
    if (this.dbnameKey) {
      credentials.dbname = _.get(environment, this.dbnameKey);
    }
    if (this.uriPrefix) {
      const uri = {
        slashes: true,
        protocol: this.uriPrefix,
        hostname: credentials.hostname,
        auth: credentials.username || ''
      };
      if (credentials.password) {
        uri.auth += ':' + credentials.password;
      }
      if (credentials.port) {
        uri.port = credentials.port;
      }
      if (credentials.dbname) {
        uri.pathname = credentials.dbname;
      }
      credentials.uri = formatUrl(uri);
    }
    if (credentials.hostname && credentials.port) {
      credentials.end_points = [{
        network_id: CONST.NETWORK_MANAGER.NETWORK_ID,
        host: credentials.hostname,
        port: credentials.port
      }];
    }
    return credentials;
  }
}
```
 
So, if the hostname and ports are not there, it does not add them in the uri. Also, in the credentials object, endpoint, host and ports  are missing.
 
 
Now, why ip and ports are null?
 
It could be if the inspect container returns empty host.

I am not sure why is it returning empty, but this is happening to hosts which are down.
 
 ```javascript
  inspectContainer() {
    logger.info(`Inspecting docker container '${this.container.id}'...`);
    return this.container
      .inspectAsync()
      .tap(info => {
        logger.info(`+-> docker container has been inspected '${this.container.id}'`);
        if (this.container.id !== info.Id) {
          this.container = this.docker.getContainer(info.Id);
        }
        this.containerInfo = info;
      })
      .catch(err => {
        logger.error(`+-> Failed to inspect docker container '${this.container.id}'`);
        logger.error(err);
        throw err;
      });
  }
  ```

I think swarm manager does not return the host info if it is down, hence the problem. I am not sure if any other problem would also cause this, we can give a little generic message as well.

Let me know what you feel.

jagadish-kb: Fix was Ok. I had the same doubts as you have put in the above, hence the question :-)Thanks for the detailed clarification.

subhankarc: I also validated my understanding with the clarification. ;)

PR #589, opened 2019-02-08
Define error codes

amitmalav:


common/utils/index.js
```diff
@@ -758,4 +759,8 @@ function loadCatalogFromAPIServer() {
       .tap(() => logger.silly('Loaded Services in catalog Are ', catalog.services))
       .tap(() => logger.silly('Loaded Plans in catalog Are ', catalog.plans));
   }
+}
+
+function getDefaultErrorMsg(statusCode, errCode) {
```
jagadish-kb: method signature can take in err object instead of statusCode & error code ? 
Reason: You can then build an error message like this:
`Service Broker Error, status code: ${err.statusCode, error code:${err.errCode} , description: ${err.message}`

Now that you are flagging errorCode at individual error classes, they can also set a message which can be utilized here to add some more additional info directly in the message. The above message is similar to what you had earlier, just that the message part is not hardcoded and its paramaterized. Makes sense?

amitmalav: Yeah. Agree with passing err object to function.
But for message: Main idea is to not show explicit error message to user. So in this case we are showing user error codes, then on user can map error codes to probable error scenarios (2xxxx means bosh; 3xxxx means docker and so on). So IMO there is no need to have a message if we are anyway having error codes

PR #593, opened 2019-02-13
Rel 2019.t01a hotfix

vivekzhere:


subhankarc:
Tests are passing, but jshint is failing. Which was the case before as well. This is fixed in the master branch. Hence, merging it.

PR #594, opened 2019-02-13
Adding placeholders for instance sharing related functions in K8sPlatformManager

jayeshjk:
* This PR addresses NGPBUG-78199.
* With https://github.com/cloudfoundry-incubator/service-fabrik-broker/pull/528/ and https://github.com/cloudfoundry-incubator/service-fabrik-broker/pull/570, instance sharing feature was introduced for director and docker instances. However, corresponding implementation was not added for requests originating from kubernetes platform, leading to failure in such bind requests.
* This PR provides a temporary fix for this by adding corresponding placeholders in K8sPlatformManager.

### DIFF
--- a/platform-managers/K8sPlatformManager.js
+++ b/platform-managers/K8sPlatformManager.js
@@ -10,14 +10,17 @@ class K8sPlatformManager extends BasePlatformManager {
   }
 
   preUnbindOperations(options) {
+    /* jshint unused:false */
     return Promise.resolve();
   }
 
   preBindOperations(options) {
+    /* jshint unused:false */
     return Promise.resolve();
   }
 
   postBindOperations(options) {
+    /* jshint unused:false */
     return Promise.resolve();
   }
 
--- a/platform-managers/K8sPlatformManager.js
+++ b/platform-managers/K8sPlatformManager.js
@@ -1,11 +1,24 @@
 'use strict';
 
 const BasePlatformManager = require('./BasePlatformManager');
+const Promise = require('bluebird');
 
 class K8sPlatformManager extends BasePlatformManager {
   constructor(platform) {
     super(platform);
     this.platform = platform;
+  }
+
+  preUnbindOperations(options) {
+    return Promise.resolve();
+  }
+
+  preBindOperations(options) {
+    return Promise.resolve();
+  }
+
+  postBindOperations(options) {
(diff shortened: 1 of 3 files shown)


platform-managers/K8sPlatformManager.js
```diff
@@ -1,13 +1,29 @@
...
 const BasePlatformManager = require('./BasePlatformManager');
+const Promise = require('bluebird');
 
 class K8sPlatformManager extends BasePlatformManager {
   constructor(platform) {
     super(platform);
     this.platform = platform;
   }
 
+  preUnbindOperations(options) {
```
abh1kg: Wouldn't it be better to have the default implementations in `BasePlatformManager`? Because I don't see sharing instances as a feature coming from K8S anytime soon

platform-managers/K8sPlatformManager.js
```diff
@@ -1,13 +1,29 @@
...
+    /* jshint unused:false */
+    return Promise.resolve();
+  }
+
+  preBindOperations(options) {
```
subhankarc: Ideally we follow empty base implementation in the `BasePlatformManager` and implementers doing specific things. It would make sense in this case as well. You can move the empty methods and put cf specific implementation in the `CfPlatformManager`

abh1kg: The problem here is that the empty base implementation is not returning a promise and that's why the binding operations fail. I guess it's just semantics- we can return an empty resolved promise in the base (as I suggested in the comment above) and not implement anything in the K8SPlatformManager or stay with what @jayeshjk has done here. I am fine either way

jayeshjk: @abh1kg @subhankarc with this implementation, if the incoming request doesn't contain `context.platform`, it will be processed by `BasePlatformManager` and will fail. We can maybe take a call based on this behaviour, i.e., if we want to enforce that request should contain `context.platform`, then we could go with current way.

ketakig12:
@jayeshjk have you tested this for k8s both below cases:
1. bosh service - i. bind ii. unbind
1. docker service - i. bind ii. unbind

jayeshjk:
@ketakig12 , yep all the four cases have been tested on aws-staging with request having `context.platform: 'kubernetes'`.

PR #595, opened 2019-02-14
Allows for clients to customize dashboard URL

jagadish-kb:
Currently the dashboard URL sent back by broker as part of create/update requests are all back to its inbuilt dashboard URLs. However there are some usecases where individual services might like to redirect the dashboard requests to their own hosted solution. To address this, we provide a mechanism via Service plan for end users to specify a dashboard_url_template at plan level which will be evaluated with the following context object:
1. Request body as recieved in the PUT/PATCH OSBAPI request
2. [Plan object](https://github.com/cloudfoundry-incubator/service-fabrik-broker/blob/master/common/models/Plan.js)

Services are free to use any of the values of the above context when constructing their template URL (refer sample dashboard url template in settings.yml). The template URL must result in an valid absolute URL, else before the request is even processed an error is thrown back to the end user.

Also since the config will be evaluated via lodash at the start of app when setting up config, the url template provided in the config must be base64 encoded in order to avoid false evaluation of the same upfront.

### DIFF
--- a/api-controllers/ServiceBrokerApiController.js
+++ b/api-controllers/ServiceBrokerApiController.js
@@ -64,11 +64,17 @@ class ServiceBrokerApiController extends FabrikBaseController {
     const params = req.body;
     const planId = params.plan_id;
     const plan = catalog.getPlan(planId);
+    const context = _
+      .chain({})
+      .merge(req.params, req.body)
+      .set('plan', plan)
+      .value();
+    const dashboardUrl = this.getDashboardUrl(context);
 
     function done() {
       let statusCode = CONST.HTTP_STATUS_CODE.CREATED;
       const body = {
-        dashboard_url: ServiceBrokerApiController.getDashboardUrl(plan, req.params.instance_id)
+        dashboard_url: dashboardUrl
       };
       if (plan.manager.async) {
         statusCode = CONST.HTTP_STATUS_CODE.ACCEPTED;
@@ -122,11 +128,17 @@ class ServiceBrokerApiController extends FabrikBaseController {
     const planId = params.plan_id;
     const plan = catalog.getPlan(planId);
     let serviceFlow;
+    const context = _
+      .chain({})
+      .merge(req.params, req.body)
+      .set('plan', plan)
+      .value();
+    const dashboardUrl = this.getDashboardUrl(context);
 
     function done() {
       let statusCode = CONST.HTTP_STATUS_CODE.OK;
       const body = {
-        dashboard_url: ServiceBrokerApiController.getDashboardUrl(plan, req.params.instance_id)
+        dashboard_url: dashboardUrl
       };
       if (plan.manager.async) {
         statusCode = CONST.HTTP_STATUS_CODE.ACCEPTED;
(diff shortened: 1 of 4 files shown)


api-controllers/ServiceBrokerApiController.js
```diff
@@ -64,11 +64,17 @@ class ServiceBrokerApiController extends FabrikBaseController {
...
+      .merge(req.params, req.body)
+      .set('plan', plan)
+      .value();
+    const dashboardUrl = this.getDashboardUrl(context);
 
     function done() {
       let statusCode = CONST.HTTP_STATUS_CODE.CREATED;
       const body = {
-        dashboard_url: ServiceBrokerApiController.getDashboardUrl(plan, req.params.instance_id)
+        dashboard_url: dashboardUrl
```
subhankarc: `dashboard_url` is not mandatory response field. Can we please enhance it so that if a service does not want it, it does not send any dashboard url? We would need for k8s deployments to keep it optional.

jagadish-kb: how do you propose we do this? Thats a new requirement, should there be another JIRA for it?

api-controllers/ServiceBrokerApiController.js
```diff
@@ -424,12 +436,20 @@ class ServiceBrokerApiController extends FabrikBaseController {
...
 
-  static getDashboardUrl(plan, instanceId) {
+  getDashboardUrl(context) {
+    if (_.get(context, 'plan.manager.settings.dashboard_url_template') !== undefined) {
+      const urlTemplate = new Buffer(context.plan.manager.settings.dashboard_url_template, 'base64');
+      const dashboardUrl = encodeURI(_.template(urlTemplate)(context));
+      if (CONST.REGEX_PATTERN.URL.test(dashboardUrl)) {
+        return dashboardUrl;
+      }
+      throw new errors.UnprocessableEntity(`Unable to generate valid dashboard URL with the template '${urlTemplate}'`);
```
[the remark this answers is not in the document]
jagadish-kb: Didnt understand this? `UnprocessableEntity` is a error class in the above stmt, which one do you want to put it in constant? Can you put that code snippet in here?

common/constants.js
```diff
@@ -475,5 +475,8 @@ module.exports = Object.freeze({
   SYSTEM_ERRORS: ['ECONNREFUSED', 'ECONNRESET', 'EPIPE', 'ETIMEDOUT', 'ESOCKETTIMEDOUT'],
   INTERNAL: 'internal',
   ALL: 'all',
-  DISABLED: 'disabled'
+  DISABLED: 'disabled',
+  REGEX_PATTERN: {
```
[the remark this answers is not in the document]
jagadish-kb: A simple comment would mean ending up explaining the entire regex :-). There are many sites which do these easily. One such sample can be found [here](https://regex101.com/r/gHjwWN/1) Just copy paste your required regex and it will explain.

api-controllers/ServiceBrokerApiController.js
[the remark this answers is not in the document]
amitmalav: I mean conventionally we define errors on top of class.
```js
const UnprocessableEntity = errors.UnprocessableEntity;
new UnprocessableEntity(....)
```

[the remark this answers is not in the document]
jagadish-kb: Generally the usage is if there is more than one error you are using in your class then you have ref of errors and refer the other error classes via errors so that you dont have to import each error individually. Using each error also works but its verbose. Dont see how one is better than the other. 
Only scenario is if in the entire class i have only a specific type of error, then i generally use the way you have mentioned above. 
Can refer existing code for similar code.

api-controllers/ServiceBrokerApiController.js
subhankarc: I leave it to you. :)

I thought now that dashboard_url are not provided by SF and one can bring in their own, one can also ask how do they not send it. K8s based services would be those examples. We could conditionally check and send if needed may be.

jagadish-kb: i was thinking of adding 'dashboard' as one of the supported features. If dashboard is supported, we check the template url and use it if its present else use SF default URL. But this would require changing product component of all services yml file and adding dashboard in them. Better to fix this additional skipping logic in another PR. 

Another way would be add an additinonal flag called 'dashboard_supported' at plan level whose default value is treated as true unless its defined and set to false. This way no changes in any existing services, but i prefer the first approach to this one as it goes well with the supported features logic.

subhankarc: That's fine.

PR #600, opened 2019-02-22
[interoperator] Improve logging

vivekzhere:
- Add object id to logs from controller
- Special delete even when the status is in progress


subhankarc:
@vivekzhere @amitmalav 

The PR looks fine. Please run the pipeline and test once before merging.

PR #601, opened 2019-02-25
Improve eventlogs generated for metering 

muralisc:
- Log successful metering events also.
- Log instance id also


kamath-prasad:
@muralisc : can you pls check and add missing UT's ?

muralisc:
Pipeline # 665

PR #603, opened 2019-03-05
[interoperator]  Runtime configs

vivekzhere:
- Worker count as runtime time config
- Read from configmap
- Use client-go to fetch configmap
- Informers not setup before controllers

### DIFF
--- a/interoperator/pkg/internal/config/config.go
+++ b/interoperator/pkg/internal/config/config.go
@@ -1,0 +1,113 @@
+package config
+
+import (
+	"fmt"
+	"os"
+	"reflect"
+	"strconv"
+	"strings"
+
+	"k8s.io/client-go/rest"
+
+	"github.com/cloudfoundry-incubator/service-fabrik-broker/interoperator/pkg/constants"
+
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/client-go/kubernetes"
+	logf "sigs.k8s.io/controller-runtime/pkg/runtime/log"
+)
+
+var log = logf.Log.WithName("config.manager")
+
+// InteroperatorConfig contains tuneable configs used by interoperator
+type InteroperatorConfig struct {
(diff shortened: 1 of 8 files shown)


interoperator/pkg/internal/config/config.go
```diff
@@ -0,0 +1,108 @@
...
+		kubeConfig: kubeConfig,
+		clientset:  clientset,
+	}, nil
+}
+
+func (cfg *config) updateConfig() error {
+	var err error
+	cfg.configMap, err = cfg.clientset.CoreV1().
+		ConfigMaps(constants.ServiceFabrikNamespace).
+		Get(constants.ConfigMapName, metav1.GetOptions{})
```
amitmalav: Function name is confusing, why name it `updateConfig` when we are fetching config?

vivekzhere: Renamed

vivekzhere:
> Can we also mention this information somewhere? Maybe create a new wiki page [here](https://github.com/cloudfoundry-incubator/service-fabrik-broker/wiki)

#605 updates the SF2.0 read me with the interoperator config details also

### Configs supported by Interoperator

The number of workers for the service instance controller and service binding controller can be configured using a config map. The name of the config map must be `interoperator-config` and the namespace must be `default`. Sample config map is:


```
---
apiVersion: v1
kind: ConfigMap
metadata:
  name: interoperator-config
  namespace: default
data:
  instanceWorkerCount: "5"
  bindingWorkerCount: "10"
```

The current supported configs are
* instanceWorkerCount - number of workers for the service instance controller
* bindingWorkerCount - number of workers for the service binding controller

If the config map is updated, interoperator must be restarted for the configs like number of workers to take effect.

amitmalav:
@vivekzhere Can you make travis pass and also please run dev pipeline once

vivekzhere:
> 
> 
> @vivekzhere Can you take care of creating the ConfigMap in the same NS as of the pods?

Reads the configmap from the same namespace as pods. If the namespace details is not there, reads from default namespace

PR #606, opened 2019-03-13
Use options for docker

muralisc:
- For old docker deployments no AppliedOptions field is present.
So for Docker Delete events use option field
- Also validate the options/appliedOptions field so that invalid metering events are not created


subhankarc:
@muralisc please add a test for the validateOptions method which is newly added.

kamath-prasad:
@muralisc : is dev pipeline run ? Can this be merged now ?

muralisc:
@kamath-prasad # 717 succeeded.

PR #608, opened 2019-03-15
Send instance type EventLog

muralisc:
- The event logs for both docker and director are now sent under 'director' nesting.
this changes conditionally sends docker/director based on the nature of plan SKU


muralisc:
Pipeline # 732

PR #610, opened 2019-03-16
Using dummy context while computing outdated deployments

jayeshjk:
* This PR removes unnecessary per deployment call to ApiServer for fetching `context` while computing outdated deployments.

### DIFF
--- a/api-controllers/ServiceFabrikAdminController.js
+++ b/api-controllers/ServiceFabrikAdminController.js
@@ -267,7 +267,7 @@ class ServiceFabrikAdminController extends FabrikBaseController {
       return _
         .chain(deployment)
         .pick('name', 'releases', 'stemcell')
-        .set('diff', utils.unifyDiffResult(deployment))
+        .set('diff', utils.unifyDiffResult(deployment, true))
         .set('instance', _
           .chain(deployment.entity)
           .pick('name', 'last_operation', 'space_guid', 'service_plan_id', 'service_plan_guid', 'dashboard_url')
@@ -280,8 +280,11 @@ class ServiceFabrikAdminController extends FabrikBaseController {
     return this
       .findOutdatedDeployments()
       .then(deployments => {
+        const mappedDeployments = _.map(deployments, mapDeployment);
         const locals = {
-          deployments: _.map(deployments, mapDeployment)
+          deployments: _.filter(mappedDeployments, deployment => {
+            return !_.isEmpty(deployment.diff);
+          })
         };
         res.format({
           html: () => res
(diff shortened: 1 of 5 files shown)


jayeshjk:
Adding 'DO-NOT_MERGE' as few sanity checks are pending.

jayeshjk:
@vinaybheri FYI.

api-controllers/ServiceFabrikAdminController.js
```diff
@@ -346,25 +346,20 @@ class ServiceFabrikAdminController extends FabrikBaseController {
...
-                .chain(deployment)
-                .assign(_.pick(result, 'diff', 'manifest'))
-                .get('diff')
-                .size()
-                .gt(0)
-                .value()
-              );
-          });
+        const opts = {};
+        opts.context = {
```
amitmalav: @jayeshjk I don't think it will work as other fields like org_guid and space_guid also are part of original context. It will always show diff as `tags.org_guid` and `tags.space_guid` in [manifest](https://github.com/cloudfoundry-incubator/service-fabrik-broker/blob/master/operators/bosh-operator/DirectorService.js#L870)

amitmalav:
I think what we can do for this particular api is that we just ignore diff in `tags` field. But again whatever we return won't be the true diff. Not sure if org_guid/space_guid/context of instance can change during lifecycle operations. If they don't change then we can safely ignore this diff in tags field.
cc: @subhankarc

jayeshjk:
@amitmalav  @subhankarc 

IIUC fix is necessary in order to avoid swamping of ApiServer with many requests for `platformContext`. A possible way is making `platformContext` request for only deployments with diffs and regenerate the diff. Code might get ugly and little bit inefficient, but result will be correct. Maybe let's discuss such approaches once.

jayeshjk:
@amitmalav @subhankarc Have implemented a logic to ignore tags altogether from the difference as we had discussed. I have compared the output of the original API and this modified API and apart from few extra empty lines (this could be also handled, but seems unnecessary and will make it clumsier) everything looks good. Please have a look.

common/utils/index.js
```diff
@@ -550,19 +550,31 @@ function hasChangesInForbiddenSections(diff) {
   return false;
 }
 
-function unifyDiffResult(result) {
+function unifyDiffResult(result, ignoreTags) {
```
amitmalav: @jayeshjk Can you please write one sample diff result in comment above this method. It is really confusing and difficult to understand w/o it.

jayeshjk:
@amitmalav @subhankarc Added UTs. Deliberately kept the realistic data instead of dummy info so that we have a genuine sample.

subhankarc:
@jayeshjk please run pipeline. Let's send it tomorrow.

PR #612, opened 2019-03-19
Don't merge spec.parameters in sfserviceinstance on update call

amitmalav:
Earlier on receiving update call, we were doing deep merge of spec. It was leading to merging of parameters field also.
For example old existing spec was like below:
```javascript
spec: {
  parameters: {
    param1: value1
  }
}
```
Now when user fires update call with -c params as `-c {param2 : value2}`
Ideally old parameters should get overwritten in spec object but with current approach it was resulting in following spec because of deep merging
```javascript
spec: {
  parameters: {
    param1: value1,
    param2: value2,
  }
}
```

With this fix, we will start getting desired result

### DIFF
--- a/api-controllers/ServiceBrokerApiController.js
+++ b/api-controllers/ServiceBrokerApiController.js
@@ -196,7 +196,7 @@ class ServiceBrokerApiController extends FabrikBaseController {
             }
           });
         } else {
-          return eventmesh.apiServerClient.patchOSBResource({
+          return eventmesh.apiServerClient.updateOSBResource({
             resourceGroup: CONST.APISERVER.RESOURCE_GROUPS.INTEROPERATOR,
             resourceType: CONST.APISERVER.RESOURCE_TYPES.INTEROPERATOR_SERVICEINSTANCES,
             resourceId: req.params.instance_id,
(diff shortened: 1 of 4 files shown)


api-controllers/ServiceBrokerApiController.js
```diff
@@ -196,7 +196,7 @@ class ServiceBrokerApiController extends FabrikBaseController {
             }
           });
         } else {
-          return eventmesh.apiServerClient.patchOSBResource({
+          return eventmesh.apiServerClient.updateOSBResource({
```
subhankarc: will it take care of context being patched?

amitmalav: Why do we want to patch context? Why not overwrite the context object?

subhankarc: We want to patch context as delete call does not send context and we do not want to loose the context overwriting it.

amitmalav: **What it will do**
Old CRD is
```javascript
spec:
  context:
    org_guid: xx
    platform: cf
```
If Patch call has context: {org_guid: yy}
New CRD will become
```javascript
spec:
  context:
    org_guid: yy
```

If Patch call doesn't have context
New CRD doesn't change context field
```javascript
spec:
  context:
    org_guid: xx
    platform: cf
```

subhankarc: hold on, if the spec is having no context object, will it keep the context?

amitmalav: apiserver patch call doesn't change the fields which are not provided in patch call AFAIK. Tested . It works 😄

subhankarc: cool.

subhankarc:
@amitmalav not merging in the week. Will merge early next week.

PR #614, opened 2019-03-25
Choose options if appliedOptions is empty

muralisc:
When create of director/docker fails the 'appliedOptions' field is empty.
In such scenarios use 'options'


subhankarc:
@muralisc Did you test the dev pipeline? Can this be merged?

PR #615, opened 2019-03-25
Deployments with '0000' network segment index don't get updated

amitmalav:


sauravmndl:
Add UT.

amitmalav:
@sauravmndl UT is already present for this scenario. IMO this if condition shouldn't even be present here. Kept it just so that we don't have any unintended consequences.

sauravmndl:
> @sauravmndl UT is already present for this scenario. IMO this if condition shouldn't even be present here. Kept it just so that we don't have any unintended consequences.

Code coverage went down. That is why asked. Otherwise looks fine.

amitmalav:
@sauravmndl Those 2 files are not related to the changes. Coverage flickers in those 2 files :)

sauravmndl:
@amitmalav Ok. Perfect ! Thanks for clarifying. We can merge the PR then.

PR #622, opened 2019-04-08
 [interoperator] Dynamic Watches 

vivekzhere:
- Setup watches reading a config map
- Updated binding and instance controllers
- Watches set only on startup
- Add sample config map
- Populate configmap on startup
- Remove use go client-go directly
- Use sources template to compute watches
- Restart interoperator on change in watch list


vivekzhere:
Tested On K8s. Few more unit tests need to be added. Working on it. Ready for Review

vivekzhere:
@amitmalav @subhankarc @jayeshjk Ready for review

interoperator/pkg/controller/sfplan/sfplan_controller.go
```diff
@@ -65,18 +74,39 @@ func add(mgr manager.Manager, r reconcile.Reconciler) error {
 	return nil
 }
 
+func restartOnWatchUpdate(mgr manager.Manager, initWatches chan bool) {
+	for {
+		select {
+		case <-initWatches:
```
amshuman-kr: Shouldn't we select from `stopCh` as well to handle the graceful termination?

interoperator/config/samples/interoperator-config.yaml
```diff
@@ -0,0 +1,35 @@
+---
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: interoperator-config
+data:
+  instanceWorkerCount: "5"
```
amshuman-kr: Anyway we have to parse YAML for the deeper structs for `instanceContollerWatchList` and `bindingContollerWatchList`. Why not just bite the bullet and dump the whole YAML as a single key in the configmap?

interoperator/pkg/internal/config/config.go
```diff
@@ -98,16 +118,102 @@ func (cfg *config) GetConfig() *InteroperatorConfig {
...
+	toCreate := false
+	if apiErrors.IsNotFound(err) {
+		toCreate = true
+		cfg.configMap = &corev1.ConfigMap{}
+		cfg.configMap.SetName(constants.ConfigMapName)
+		cfg.configMap.SetNamespace(cfg.namespace)
+		cfg.configMap.Data = make(map[string]string)
+	}
+
+	val := reflect.ValueOf(interoperatorConfig)
```
amshuman-kr: If we dump the whole yaml as a single key/value in the `ConfigMap` as mentioned [above](https://github.com/cloudfoundry-incubator/service-fabrik-broker/pull/622/files#diff-5c71366ff7ef7d1a65501881ecb6544cR7), we can then simply use YAML/JSON (en|de)coder on the whole thing instead the current mixture of `reflect` and (en|de)coding.

interoperator/pkg/controller/sfplan/sfplan_controller.go
vivekzhere: Added

interoperator/config/samples/interoperator-config.yaml
vivekzhere: Updated.

interoperator/pkg/internal/config/config.go
vivekzhere: Use of reflect removed

PR #625, opened 2019-04-14
Adding schedule backup call after successful BoshBasedRestore for pitr enabled services

jayeshjk:
* Adding a schedule backup call similar to [this](https://github.com/cloudfoundry-incubator/service-fabrik-broker/blob/rel-2019.T03a/operators/restore-operator/RestoreService.js#L120) for BoshBasedRestore.

### DIFF
--- a/operators/bosh-restore-operator/BoshRestoreService.js
+++ b/operators/bosh-restore-operator/BoshRestoreService.js
@@ -510,12 +510,7 @@ class BoshRestoreService extends BaseDirectorService {
 
   async processPostStart(resourceOptions) { 
     await this.runErrand(resourceOptions, 'postStartErrand'); 
-    if (this.service.pitr === true) {
-      await this.reScheduleBackup({
-        instance_id: resourceOptions.instance_guid,
-        afterXminute: config.backup.reschedule_backup_delay_after_restore || CONST.BACKUP.RESCHEDULE_BACKUP_DELAY_AFTER_RESTORE
-      });
-    }
+    
     const patchObj = await this.createPatchObject(resourceOptions, 'succeeded');
     let patchResourceObj = {
       resourceGroup: CONST.APISERVER.RESOURCE_GROUPS.RESTORE,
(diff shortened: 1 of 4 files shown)


jayeshjk:
Pipeline green with build # 808

operators/bosh-restore-operator/BoshRestoreService.js
```diff
@@ -509,6 +510,12 @@ class BoshRestoreService extends BaseDirectorService {
 
   async processPostStart(resourceOptions) { 
     await this.runErrand(resourceOptions, 'postStartErrand'); 
+    if (this.service.pitr === true) {
+      await this.reScheduleBackup({
```
amitmalav: Do we want to fail the restore if reScheduling fails?
In RestoreService this part is put under tap, https://github.com/cloudfoundry-incubator/service-fabrik-broker/blob/master/operators/restore-operator/RestoreService.js#L111

jayeshjk: Good point! I will make the changes. @abh1kg I think this makes sense, right?

abh1kg: makes sense to go with @amitmalav on this

subhankarc: @amitmalav can you check this now? I guess Jayesh has taken care of it in the latest fix. Please approve if it looks fine to you.

operators/bosh-restore-operator/BoshRestoreService.js
```diff
@@ -522,7 +524,13 @@ class BoshRestoreService extends BaseDirectorService {
       _.set(patchResourceObj, 'status.response', patchObj);
     }
     await eventmesh.apiServerClient.patchResource(patchResourceObj);
-    return this.patchRestoreFileWithFinalResult(resourceOptions, patchObj);
+    await this.patchRestoreFileWithFinalResult(resourceOptions, patchObj);
+    if (this.service.pitr === true) {
+      this.reScheduleBackup({
```
jayeshjk: @amitmalav modified flow to schedule backup after the state is updated in CRD and restoreFile. Also as the `reScheduleBackup` call is not being awaited, it's failure won't affect restore flow, similar to old restore flow. Please have a look once.

PR #629, opened 2019-04-23
Returning dashboardUrl only if external endpoint attributes are set

subhankarc:
* Currently, if the external app is not used, dashboard url is returned with broker link
* Returning the custom dashboard if the template is set, else return the SF dashboard if external attributes exist, or else do not return any dashboard url


amitmalav:
Coverage reduced but still all checks passed? :o

subhankarc:
@amitmalav yeah, only return null is not covered. :)

PR #630, opened 2019-04-23
Add Aliclient support for BoshRestore and Delete Snapshot

amitmalav:
* Added required methods for bosh restore in AliClient
* Added required method for snapshot deletion

### DIFF
--- a/data-access-layer/iaas/AliClient.js
+++ b/data-access-layer/iaas/AliClient.js
@@ -58,23 +58,25 @@ class AliClient extends BaseCloudClient {
       options = container;
       container = this.containerName;
     }
-    return this.storage
-      .useBucket(container)
-      .list(options)
-      .then(listOfFiles => {
-        let list = listOfFiles.objects;
-        let isTruncated = listOfFiles.isTruncated;
-        let marker = listOfFiles.nextMarker;
-        const files = [];
-        _.each(list, file => files.push(_
-          .chain(file)
-          .pick('name', 'lastModified')
-          .set('isTruncated', isTruncated)
-          .set('marker', marker)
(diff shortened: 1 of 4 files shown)


data-access-layer/iaas/AliClient.js
```diff
@@ -139,15 +147,152 @@ class AliClient extends BaseCloudClient {
...
+      timeout: CONST.ALI_CLIENT.ECS.REQ_TIMEOUT,
+      method: 'POST'
+    };
+    return Promise
+      .try(() => {
+        return this.computeClient
+          .request('CreateDisk', reqParams, requestOption)
+          .tap(result => logger.info(`Created disk ${result.DiskId} from snapshot ${snapshotId}, now wait for it to be available...`))
+          .then(result => this._waitForDiskAvailability(result.DiskId))
+          .tap(diskDetails => logger.info(`Created disk ${diskDetails.DiskId} from snapshot ${snapshotId} is now Available with status: ${diskDetails.Status}`))
```
jayeshjk: Can you return `diskDetails` in a format similar to output of `getDiskMetadata `. For example, consider `AzureClient`: https://github.com/cloudfoundry-incubator/service-fabrik-broker/blob/rel-2019.T03b/data-access-layer/iaas/AzureClient.js#L68.

Currently BoshBasedRestore Operator doesn't have a tight dependency on this diskInfo format. However we store information of old and new disks created as part of restore operation and it would be nice to have formats for both of them to be same.

abh1kg: 👍

amitmalav: Done

PR #632, opened 2019-05-02
Set Bind Operation timeout to < 60sec

amitmalav:
Cloud controller req. times out in `60sec`
https://github.com/cloudfoundry/cloud_controller_ng/blob/master/config/cloud_controller.yml#L192
It means broker should return bind credentials within these 60 seconds else it should throw timeout error.
To support this 2 changes were made
1. retry count of createCredential req to agent set to 2 and timeout set to 25sec/req
2. Entire operation should finish within 55 sec hence timeout is changed for `sync operation poller` in broker controller

### DIFF
--- a/operators/bosh-operator/DirectorService.js
+++ b/operators/bosh-operator/DirectorService.js
@@ -742,6 +742,7 @@ class DirectorService extends BaseDirectorService {
       this.executeActions(CONST.SERVICE_LIFE_CYCLE.PRE_BIND, actionContext),
       this.getDeploymentIps(deploymentName),
       (preBindResponse, ips) => utils.retry(() => this.agent.createCredentials(ips, binding.parameters, preBindResponse), {
+        operation: 'CREATE_CREDENTIAL_AGENT',
         maxAttempts: 2,
         timeout: config.agent_operation_timeout || CONST.AGENT.OPERATION_TIMEOUT_IN_MILLIS
       })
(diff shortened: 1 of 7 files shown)


amitmalav:
@sauravmndl Didn't understand your comment. That is what we changed as part of this PR (https://github.com/cloudfoundry-incubator/service-fabrik-broker/pull/632/files#diff-9701b968147d47a5682ea3bee6fdf41aR745). Are you referring to this same line?

sauravmndl:
I missed that DirectorService.js file is also changed. The  OPERATION_TIMEOUT_IN_MILLIS is used in the file during bind/unbind operation. I was referring to that. Perfect! No need to change anything according to me. Ignore the comment.

subhankarc:
@amitmalav  dev pipeline verified ? Please label it accordingly

operators/bosh-operator/DirectorService.js
```diff
@@ -742,7 +742,7 @@ class DirectorService extends BaseDirectorService {
       this.executeActions(CONST.SERVICE_LIFE_CYCLE.PRE_BIND, actionContext),
       this.getDeploymentIps(deploymentName),
       (preBindResponse, ips) => utils.retry(() => this.agent.createCredentials(ips, binding.parameters, preBindResponse), {
-        maxAttempts: 3,
+        maxAttempts: 2,
```
subhankarc: Our intention was also to not show operation failed after 3(2) attempts only as message and also show specific message. How would that be achieved ?

amitmalav: How does it look now?

subhankarc: cool!

PR #635, opened 2019-05-09
Bosh operator should be able to create/update/delete deployment with 0000 network segment index

amitmalav:


subhankarc:
hope this is all where we have these checks ! :)

PR #644, opened 2019-06-06
Fix metering job update

amitmalav:


jayeshjk:
Can you also please modify [this logger](https://github.com/cloudfoundry-incubator/service-fabrik-broker/blob/rel-2019.T05a/jobs/ScheduleManager.js#L210)? I think this was the source of all confusion.

PR #651, opened 2019-07-01
Making scheduler and instance controller serial

subhankarc:


subhankarc:
@vivekzhere  can you please change the logger or point me to it ?

Issue #389, opened 2018-09-03
Service specific Deployment Status


welfwalter:
After Cloud Foundry's Cloud Controller has sent the service provisioning request to Service Fabrik, it polls for the status.
When the Service Fabrik triggered bosh and bosh finished the deployment, the Service Fabrik sends back the status "Created".

But for some services, some further initialization are then still necessary. These may take a while before the service is really consumable.
In Cloud Cockpit we would then still want to see status "Creating". (Or maybe an additional status like "Finalizing".)
So we would like to have the possibility to control which status is returned from Service Fabrik to Cloud Controller – specific for a service.
 
A possible implementation could be a new method “getDeploymentStatus” in the Service Fabrik Agent API.

subhankarc:
@welfwalter For asynchronous service provisioning, initial create request is responded with 202 and then platform(cloud controller) starts polling for the last operation. According to the OSBAPI [spec](https://github.com/openservicebrokerapi/servicebroker/blob/v2.14/spec.md#polling-last-operation-for-service-instances) the broker should return state as part of the body and [here](https://github.com/openservicebrokerapi/servicebroker/blob/v2.14/spec.md#body-1) it mentions that `the valid values are in progress, succeeded, and failed`.

So, I don't think we can send anything else other than in progress where the service instance is not yet ready to be used.

To support the use-case you have, I could think of some post processing hook which polls agent for its readiness and then sets the status as succeeded.

@jagadish-kb please let us know your view.

welfwalter:
Hi subhankarc,
if the list of status is fixed to these three values then the value should stay as `in progress`.
The post processing hook polling the agent sounds fine.
With regards
 Welf

PR #655, opened 2019-07-05
Add post processing polling to bosh manager

steinri:
This add the feature to query the agent for post processing operations after creating and updating a bosh deployment as requested in #389.

If the bosh operation was successful the state of the CR is switch from `in_progress` to `post_processing` instead of to `succeeded`. Then a status poller is polling the corresponding agent for the last operation of the corresponding event type. When this polling returns that the post processing succeeded the state in the CR is changed to `succeeded`.

- I implemented this behavior is an opt-in with a feature switch.
- It is explicitly written in a way that the pre_deploy and pre_delete operations may also be asynchronous in the future
- This does not require any changes in the interoperator or the templates (since the used template currently maps unknown states to in process)
- I tried to stick with the current api and conventions, but you can give me feedback if you want something different
  + ~feature switch: `feature.ServiceInstancePostProcessing`~
  + agent feature names: `lifecycle.async.postcreate` and `lifecycle.async.postupdate`
  + agent api: `GET /lifecycle/postcreate` and `GET /lifecycle/postupdate`

### DIFF
--- a/data-access-layer/service-agent/Agent.js
+++ b/data-access-layer/service-agent/Agent.js
@@ -209,8 +209,8 @@ class Agent extends HttpClient {
    * @returns {*} The response of the agent
    */
   getProcessingState(ips, eventType, lifecycleState) {
-    const featureName = `processing.${lifecycleState}${eventType}`; // TODO check feature name
-    const pathname = `processing/${lifecycleState}${eventType}`; // TODO check route
+    const featureName = `lifecycle.async.${lifecycleState}${eventType}`;
+    const pathname = `lifecycle/${lifecycleState}${eventType}`;
 
     return this
       .getHost(ips, featureName)
--- a/data-access-layer/service-agent/Agent.js
+++ b/data-access-layer/service-agent/Agent.js
@@ -200,6 +200,29 @@ class Agent extends HttpClient {
       .then(ip => this.post(ip, 'lifecycle/preupdate', context, CONST.HTTP_STATUS_CODE.OK, agentCredsBeforeUpdate));
   }
 
+  /**
+   * Poll the agent for the process of the lifecycle last operation within the deployment
+   *
+   * @param ips The ip addresses of all available agents
+   * @param {string} eventType The current lifecycle event, supported values: "create", "update"
+   * @param {string} lifecycleState The current lifecycle state of the operation, supported values: "post"
+   * @returns {*} The response of the agent
+   */
+  getProcessingState(ips, eventType, lifecycleState) {
+    const featureName = `processing.${lifecycleState}${eventType}`; // TODO check feature name
+    const pathname = `processing/${lifecycleState}${eventType}`; // TODO check route
+
+    return this
+      .getHost(ips, featureName)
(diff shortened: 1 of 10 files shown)


steinri:
Note that this is not ready to merge, but ready for review comments. I am still looking forward to write tests and add a PR for the agents api documentation.

test/test_broker/mocks/agent.js
```diff
@@ -70,6 +73,19 @@ function preUpdate(expectedReturnStatusCode) {
     .reply(expectedReturnStatusCode || 200, {});
 }
 
+function getPostCreateProcessingState(body, status) {
+  return nock(agentUrl)
+    .replyContentLength()
+    .get('/v1/processing/postcreate')
```
subhankarc: Can we name the feature as lifecycle.post[create|update] ? We already have `lifecycle.preupdate`. In that case, it will be consistent.

data-access-layer/service-agent/Agent.js
```diff
@@ -200,6 +200,29 @@ class Agent extends HttpClient {
...
 
+  /**
+   * Poll the agent for the process of the lifecycle last operation within the deployment
+   *
+   * @param ips The ip addresses of all available agents
+   * @param {string} eventType The current lifecycle event, supported values: "create", "update"
+   * @param {string} lifecycleState The current lifecycle state of the operation, supported values: "post"
+   * @returns {*} The response of the agent
+   */
+  getProcessingState(ips, eventType, lifecycleState) {
```
subhankarc: Can we name the feature as lifecycle.post[create|update] ? We already have `lifecycle.preupdate`. In that case, it will be consistent. You can then handle this in a generic way.
`lifecycleState` can then support pre and post both.

test/test_broker/mocks/agent.js
steinri: The difference in that cases is that this is an async polling, for the service instance to finish with some internal logic after the create/update. The preupdate is a syncronous action, which is executed without polling afterwards.
It may be desirable to add an asynchronous preupdate (first call POST and then poll with get GET) in the future, and then the featureflag `lifecycle.preupdate` would not support that.

steinri: We could call the feature `lifecycle.async.postupdate` and `lifecycle.async.preupdate`, which calls the router `GET /lifecycle/postupdate`. For the preupdate case it could be `POST /lifecycle/postupdate?async=true`

subhankarc: makes sense! 👍

subhankarc:
@steinri Is this PR and the other dependent PR https://github.com/cloudfoundry-incubator/service-fabrik-broker/pull/656 good to merge ?

@amitmalav  Can you please run a dev pipeline and merge it once @steinri gives a go ahead.

steinri:
From my point of view they are ready to merge. I have deployed the corresponding bosh release in our landscape and everything behaves as expected.

amitmalav:
@steinri
Can you please rebase your branch with master

steinri:
I have rebased it

PR #660, opened 2019-07-28
[interoperator] Bind timeout during watch refresh

vivekzhere:
- Fetch clusterID only if needed.
- Ignore clusterID err on delete

Feature: NGPBUG-91418


amitmalav:
Can you please raise PR on rel-2019.T07a branch also

PR #667, opened 2019-07-30
Let the deletion fail, if there are still resources for the CR

steinri:
If the last operation was marked as successful, but there are still resources attached to the CR. Then the deletion did not succeed and a successful respond leads to inconsistencies.

This is very critical if the last operation of the cloud controller is delete, but there are still resources in the landscape.

### DIFF
--- a/api-controllers/ServiceBrokerApiController.js
+++ b/api-controllers/ServiceBrokerApiController.js
@@ -294,6 +294,13 @@ class ServiceBrokerApiController extends FabrikBaseController {
       logger.debug('returning ..', body);
       return Promise.try(() => {
         if (_.get(operation, 'type') === 'delete' && body.state === CONST.OPERATION.SUCCEEDED && resourceGroup === CONST.APISERVER.RESOURCE_GROUPS.INTEROPERATOR) {
+          // consistency check that a successfully deleted resource has no resources attached
+          if (!_.isEmpty(result.resources)) {
+            logger.error(`The Service instance ${resourceId} still has resources, although the deletion operation is marked as successful after the deletion.`);
+            body.state = CONST.OPERATION.FAILED;
+            body.description = `Failed to delete the Service Instance ${resourceId}. There are still resources attached after the deletion.`;
+            return Promise.resolve({});
+          }
           return this.removeFinalizersFromOSBResource(
             resourceType,
             resourceId,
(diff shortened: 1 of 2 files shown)


amitmalav:
@steinri,
What is the status of this issue. Were you able to figure out the root cause of why this happens?

steinri:
Hello @amitmalav ,

this problem occurred two times in a few hundred cases, but is not reproducible:
We had a system, which was deleted in cloud foundry, but still exists in the landscape and the last operation of the CR was "create" (not delete).
Using dynatrace deep inspection on one of the occurrences I was able to see that the service fabrik accepted the delete request (but the information wasn't in the CR).
Afterwards only one polling request was send, therefore I assume that it directly send `succeeded`, because from the service fabrik point of view the last operation was `creation` and this operation succeeded.
Looking at the code this should never happen and happens in two of a few hundred deployments, however especially in the deletion case it leads to a problematic inconsistency. So I was not able to find a the root cause and can only describe what happened.

I created the Pull Request in the assumption, that "if there are still resources left in the landscape, then by definition the deletion was not successful". Another approach would be to compare the last operation with the last operation of the polling with the last operation in the resource. However this last operation is currently only set by the inter-operation and it seems to be wrong to use it in the broker.

Best Regards
Richard

api-controllers/ServiceBrokerApiController.js
```diff
@@ -294,6 +294,13 @@ class ServiceBrokerApiController extends FabrikBaseController {
       logger.debug('returning ..', body);
       return Promise.try(() => {
         if (_.get(operation, 'type') === 'delete' && body.state === CONST.OPERATION.SUCCEEDED && resourceGroup === CONST.APISERVER.RESOURCE_GROUPS.INTEROPERATOR) {
+          // consistency check that a successfully deleted resource has no resources attached
+          if (!_.isEmpty(result.resources)) {
```
amitmalav: @steinri 
This change will put the instance in inconsistent state. Ideally when the state is set to succeeded, finalizer of interoperator is also removed from sfserviceinstance CR [(ref.)](https://github.com/cloudfoundry-incubator/service-fabrik-broker/blob/master/interoperator/pkg/controller/sfserviceinstance/sfserviceinstance_controller.go#L371-L376).

When next time delete is tried on the instance, broker will set the state as delete but now interoperator does not act on this sfserviceinstance as interoperator finalizer was already removed. It means that state of sfserviceinstance will always be delete in progress.

Ideally what you are reporting should not have happened. I'd suggest that we figure out the root cause in case it pops up again and fix it there.

steinri: I agree that the root cause should be fixed and that this might not be the correct implementation for the desired behavior.
However I think that the service fabrik should be a little more robust to such cases. In general it might be a better solution to store the last operation in the sfserviceinstance CR and compare it to the operation from the polling request instead of trusting the Cloud Controller to send the correct last operation.

PR #670, opened 2019-08-04
[Potential test cycle fix]introducing separate error for ApiServer page not found

jayeshjk:
* Sometimes, during initialisation, 404 is thrown by ApiServer with the message as `page not found`, even when the resource actually exists.
* `DbManager` treats it as binding not found scenario and stops retrying for connection, which is incorrect behaviour. Modifying this flow by distinguishing between `resource not found` and `page not found` scenarios.

### DIFF
--- a/data-access-layer/db/DBManager/DBManager.js
+++ b/data-access-layer/db/DBManager/DBManager.js
@@ -71,7 +71,7 @@ class DBManager {
       }
     })
       .catch(NotFound , err => {
-          logger.warn('MongoDB binding to ServiceFabrik not found. This generally should not occur. More Info:', err);
+        logger.warn('MongoDB binding to ServiceFabrik not found. This generally should not occur. More Info:', err);
       })
       .catch(err => {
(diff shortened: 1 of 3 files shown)


data-access-layer/db/DBManager/DBManager.js
```diff
@@ -115,8 +122,8 @@ class DBManager {
           resourceId: _.toLower(CONST.FABRIK_INTERNAL_MONGO_DB.BINDING_ID)
         });
       }, {
-        maxAttempts: 5,
-        minDelay: 1000,
+        maxAttempts: 3,
+        minDelay: CONST.APISERVER.RETRY_DELAY,
         predicate: err => !(err instanceof NotFound)
```
amitmalav: `predicate: err => !(err instanceof NotFound) ||  (err instanceof PageNotFound)`

jayeshjk: Updated, but used slightly different predicate which felt little more intuitive.

PR #673, opened 2019-08-13
Feature/interop multiwatch

vivekzhere:
[interoperator] Multi Cluster Watch

- Move owner Reference to annotations
- Watch from channels
- For each cluster watch all resources
- Dynamically pickup plan changes
- Remove restart on plan change
- TODO : Unit tests

Feature: HCPCFS-2381


vivekzhere:
@amshuman-kr @subhankarc @amitmalav
Please Review. Testcases are WIP

interoperator/pkg/internal/resources/resources.go
```diff
@@ -178,12 +179,20 @@ func (r resourceManager) ComputeExpectedResources(client kubernetes.Client, inst
 }
 
 // SetOwnerReference updates the owner reference for all the resources
-func (r resourceManager) SetOwnerReference(owner metav1.Object, resources []*unstructured.Unstructured, scheme *runtime.Scheme) error {
+// For multi cluster support we cannot use k8s OwnerReference as
+// owner object might be in a different cluster. So use annotations.
+// Not using labels as it was restrictions of field values.
+func (r resourceManager) SetOwnerReference(owner metav1.Object, ownerGVK schema.GroupVersionKind, resources []*unstructured.Unstructured) error {
```
amshuman-kr: Why not replicate the `SfServiceInstance`/`SfServiceBinding` instance  to the target cluster?

#### Pros
1. Ownership of resources from the `SfPlan` template can be determined locally within the target cluster without reference to the central cluster where interoperator is deployed.
1. Regular ownership management for resources created based on the `SfPlan`.
1. Provides a mechanism to horizontal/distributed scaling (if required in the future). Think about the restart times on a landscape with multiple target clusters.
1. Provides a way to reduce the number of watches across clusters (if required in the future)

#### Cons
1. Duplication of the resources between central and target clusters. But duplication will be only in two clusters for any given interoperator resource. But this might be worth it if it enables the horizontal/distributed scaling.
1. Special handling would still be required to manage the ownership for the replicated `SfServiceInstance`/`SfServiceBinding` instances. But this would be limited to the interoperator's resources.

interoperator/pkg/watches/channel.go
```diff
@@ -0,0 +1,252 @@
...
+		}
+		err = clusterWatcherList[i].start()
+		if err != nil {
+			log.Error(err, "failed to start watch on cluster",
+				"clusterID", cluster.GetName(), "controller", controllerName)
+			continue
+		}
+	}
+	go func() {
+		<-stopCh
```
amshuman-kr: Why not simply pass the `stopCh` [above](https://github.com/cloudfoundry-incubator/service-fabrik-broker/pull/673/files#diff-4e65bc02c98373a2de32876b47339db5R87) avoid this goroutine?

interoperator/pkg/watches/channel.go
```diff
@@ -0,0 +1,252 @@
...
+
+type watcher struct {
+	controllerName string
+	clusterID      string
+	dynamicClient  dynamic.Interface
+	gvr            schema.GroupVersionResource
+	events         chan<- event.GenericEvent
+	stop           <-chan struct{}
+	// stopper is the write side of the stop channel. They should have the same value.
+	stopper   chan<- struct{}
```
amshuman-kr: What is this for?

interoperator/pkg/controller/sfplan/sfplan_controller.go
```diff
@@ -76,18 +77,29 @@ func add(mgr manager.Manager, r reconcile.Reconciler) error {
...
 }
 
-func restartOnWatchUpdate(mgr manager.Manager, initWatches, stop <-chan struct{}) {
+func reconfigureWatches(mgr manager.Manager, reconfigureWatch, stop <-chan struct{}) {
 	for {
 		select {
-		case <-initWatches:
-			toUpdate, err := watches.InitWatchConfig(mgr.GetConfig(), mgr.GetScheme(), mgr.GetRESTMapper())
+		case <-reconfigureWatch:
+			drainTimeout := time.After(constants.ChannelDrainTimeout)
```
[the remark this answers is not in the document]
vivekzhere: This is done so that during watch refreshes, lets say there are total 50 sfplans, reconcile gets called 50 times and will result in recomputing watch lists from sources templates 50 times. So a logic is added that will drain all the reconcile requests happening within .5 sec and recompute watch list only once. This was done just for performance improvement. The entire Drain loop and this timeout is not functionally required.

interoperator/pkg/watches/channel.go
vivekzhere: The stop is signalled by a write to the channel (Not closing the channel). So a event sent to stopCh must be forwarded to
many go routines (one routine per cluster per gvk). If we pass the channel only one routine will read the stop event. For multiplexing the stop event this is required.

interoperator/pkg/watches/channel.go
vivekzhere: `stop `and `stopper `are the same channel. `Watchers `listen on the read only channel `stop `for the stop event. `clusterWatcher `issues stop signal to watcher using the write only `stopper `channel. 

It could be just one field `stop chan struct{}`

interoperator/pkg/controller/sfplan/sfplan_controller.go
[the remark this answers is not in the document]
amshuman-kr: No. I understood the requirement. My question was about the implementation/solution. I was asking if instead of draining the watch channel for a time period, if we can use a timestamp and and ignore the reconciliation if we are still within the time period. Could reduce additional goroutine/parallel flows and keep the logic simple.

interoperator/pkg/watches/channel.go
amshuman-kr: > If we pass the channel to all go routines, only one routine will read the stop event. For multiplexing the stop event this is required.

Why not close the channel?

vivekzhere: Will update

PR #676, opened 2019-08-19
Add schema validation for service instance operation via ajv

steinri:
Add a  middleware, which validates the parameters based on the schemas defined in the service plan.

TODO:

- [x] Add tests
- One may add a special handling for the update case: If the deployment is updates via the `/admin` api there are special service fabrik parameters. This should still be possible. However this may also be handled in the schemas itself, since the service also needs to be able to handle it.

As discussed before this is my current implementation state as a basis for the schema validation.

### DIFF
--- a/broker/lib/middleware.js
+++ b/broker/lib/middleware.js
@@ -1,5 +1,6 @@
 'use strict';
 
+const ajv = require('ajv');
 const _ = require('lodash');
 const errors = require('../../common/errors');
 const logger = require('../../common/logger');
@@ -11,6 +12,7 @@ const catalog = require('../../common/models/catalog');
 const eventmesh = require('../../data-access-layer/eventmesh');
 const CONST = require('./../../common/constants');
 const DeploymentAlreadyLocked = errors.DeploymentAlreadyLocked;
+const InvalidServiceParameters = errors.InvalidServiceParameters;
 const UnprocessableEntity = errors.UnprocessableEntity;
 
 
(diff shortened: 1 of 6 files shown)


broker/lib/middleware.js
```diff
@@ -44,6 +46,38 @@ exports.validateCreateRequest = function () {
   };
 };
 
+exports.validateSchemaForRequest = function (target, operation) {
+  return function (req, res, next) {
+    const plan = getPlanFromRequest(req);
+
+    const schema = _.get(plan, `schemas.${target}.${operation}.parameters`);
```
amitmalav: Can you please update/check-in local config [`settings.yml`](https://github.com/cloudfoundry-incubator/service-fabrik-broker/blob/master/broker/config/settings.yml) with schema definition.

broker/lib/middleware.js
```diff
@@ -44,6 +46,38 @@ exports.validateCreateRequest = function () {
...
+
+    if (schema) {
+      const parameters = _.get(req, 'body.parameters', {});
+
+      const schemaVersion = schema['$schema'] || '';
+      const validator = new ajv({ schemaId: 'auto' });
+      if (schemaVersion.includes('draft-06')) {
+        validator.addMetaSchema(require('ajv/lib/refs/json-schema-draft-06.json'));
+      } else if (schemaVersion.includes('draft-04')) {
+        validator.addMetaSchema(require('ajv/lib/refs/json-schema-draft-04.json'));
```
jayeshjk: From [ajv documentation](https://github.com/epoberezkin/ajv#using-version-6),

```
To use Ajv with draft-04 schemas in addition to explicitly adding meta-schema you also need to use option schemaId:
var ajv = new Ajv({schemaId: 'id'});
```
Has this been considered here?

broker/lib/middleware.js
```diff
@@ -44,6 +46,38 @@ exports.validateCreateRequest = function () {
   };
 };
 
+exports.validateSchemaForRequest = function (target, operation) {
+  return function (req, res, next) {
+    const plan = getPlanFromRequest(req);
+
+    const schema = _.get(plan, `schemas.${target}.${operation}.parameters`);
+
```
jayeshjk: Along with this, could you please also add few unit tests covering this middleware method?

broker/lib/middleware.js
steinri: Yes, this is handled by `new ajv({ schemaId: 'auto' });` (I just see that this should be PascalCase). This will use `id` for v4 and `#id` for v6 and v7.

broker/lib/middleware.js
steinri: Yes, I have added this as TODO :)

broker/lib/middleware.js
jayeshjk: Ohh, I had missed that statement! (y)

amitmalav:
@steinri Can you please handle TODOs and Review comments so that this PR can be merged.

broker/lib/middleware.js
steinri: I have added it to the settings

broker/lib/middleware.js
steinri: I have added tests

amitmalav:
@steinri 
Can you please fix eslint: https://travis-ci.org/cloudfoundry-incubator/service-fabrik-broker/jobs/583179633#L218

Rest of the PR looks great. Thanks.

amitmalav:
Hi @steinri 
I have fixed the eslint error, I hope that's alright.
Thanks again for the PR. merging...

steinri:
@amitmalav thanks =)

PR #734, opened 2019-10-03
[interoperator] Add sprig libary to gotemplates

vivekzhere:
- Provide a hierarchical structure for SFPlan's Context attribute
- Removed our fns which were exact copy
- `b64dec` is not exact copy. So kept both `b64dec` and `b64enc`

Feature: HCPCFS-2483

### DIFF
--- a/interoperator/pkg/internal/renderer/gotemplate/functions_test.go
+++ b/interoperator/pkg/internal/renderer/gotemplate/functions_test.go
@@ -9,6 +9,8 @@ import (
 func TestGoTemplateFunctions(t *testing.T) {
 	g := gomega.NewGomegaWithT(t)
 
+	funcMap := getFuncMap()
+
 	g.Expect(encodeToString("Hello")).To(gomega.Equal("SGVsbG8="))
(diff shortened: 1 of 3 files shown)


interoperator/pkg/internal/renderer/gotemplate/functions_test.go
```diff
@@ -36,21 +36,4 @@ func TestGoTemplateFunctions(t *testing.T) {
 	invalidObj2, err2 := unmarshalJSON(invalidObjStr2)
 	g.Expect(invalidObj2).To(gomega.BeNil())
 	g.Expect(err2).To(gomega.HaveOccurred())
-
```
subhankarc: Can we keep the tests? Even though it will test the library, but I guess that is fine. WDYT?

vivekzhere: Kept and Updated the tests the `quote`, `squote` and `strval` fn from the libarary

vinaybheri:
Validation on k8s cluster: n failed.

vinaybheri:

vivekzhere:
> 
> 
> Validation on k8s cluster: n failed.

Pipeline 1160 green except this

PR #740, opened 2019-10-16
[Doc] Mass Update of CRs in case of update of SFplan and SFService

subhankarc:


subhankarc:
@amshuman-kr Can you please have a look if the proposal look fine ?

inter-operator/architecture/basic.md
```diff
@@ -625,3 +626,15 @@ After the interoperator is ready and setup across multiple clusters as described
 3. Interoperator does not take care of the Kubernetes and OS updates to the onboarded clusters.
 4. Service owners will have to monitor the clusters and their resource situations and add additional sister clusters if required.
 
+# Mass Update of Custom Resources for Interoperator Custom Resource changes
+
+## Context
+Interoperator has custom resources like SFPlans and SFServices which one has to provide and deploy before working with interoperator. This was already described in [here](#service-fabrik-inter-operator-custom-resources). Based on the templates defined in `SFPlan`, service specific custom resources are rendered and with that, service instance creation takes place. So, in a way, `SFPlan` and also `SFService` CRs are used as references when the service specific CRs created. However, when these reference CRs like `SFPlans` and `SFServices` change, the changes are not automatically reflected on the service specific CRs. Because of that, when a service owner changes `SFPlans` and some of its attributes and templates, older service CRs are not automatically changed.
```
amshuman-kr:
```suggestion
Interoperator has custom resources like SFPlans and SFServices which one has to provide and deploy before working with interoperator. This was already described in [here](#service-fabrik-inter-operator-custom-resources). Based on the templates defined in `SFPlan`, service specific custom resources are rendered during service instance creation. So, `SFPlan` and also `SFService` CRs are used as references when the service specific CRs created. However, when these reference CRs like `SFPlans` and `SFServices` change, the changes are not automatically reflected on the service specific CRs. Because of that, when a service owner changes `SFPlans` and some of its attributes and templates, already existing service instance CRs are not automatically changed.
```

inter-operator/architecture/basic.md
```diff
@@ -625,3 +626,15 @@ After the interoperator is ready and setup across multiple clusters as described
...
+              The situation is similar if a service broker updates its catalog and any of it's metadata, which is used by service to configure a specific service instance. Should the broker trigger an update of all the service instances immediately or should it wait for an user initiated update operation ?
```
amshuman-kr:
```suggestion
              The situation is similar if a service broker updates its catalog and any of it's metadata, which is used by service to configure a specific service instance. Should the broker trigger an update of all the service instances immediately or should it wait for a user initiated update operation ?
```

inter-operator/architecture/basic.md
```diff
@@ -625,3 +626,15 @@ After the interoperator is ready and setup across multiple clusters as described
...
+              
+## Solution
+There are no generic guideline from the OSB spec as well and the solution to this would entirely depend on the service broker implementation. The problem with having immediate trigger of an update for all affected service instances would be that updates can cause downtime depending on how services are handling it.
```
amshuman-kr:
```suggestion
There are no generic guidelines from the OSB spec as well and the solution to this would entirely depend on the service broker implementation. The problem with having immediate trigger of an update for all affected service instances would be that updates can cause downtime depending on how services are handling it.
```

inter-operator/architecture/basic.md
```diff
@@ -625,3 +626,15 @@ After the interoperator is ready and setup across multiple clusters as described
...
+
+## Proposal
+Interoperator being a generic broker, it should not trigger update blindly as well. We can provide a flag which can be turned on if services can afford to have blind/immediate update. In that case, a controller will reconcile all `SFServiceInstances` with update status, which would render the templates again and CRs updated again. However, this would not take care of the deleted/removed CRs if any, and the service operator will have to take care of obsolete CRs.
```
amshuman-kr: How about keeping the flag at the `SFService` and/or SFPlan` level instead of a global flag?

subhankarc: Yes, it was not explicit in the document. Thanks, I made it explicit.

PR #741, opened 2019-10-31
Returning 429 (too-many-requests) on timeout

jayeshjk:
* Return HTTP Code 429 (indicating "Too Many Requests") on Bind timeouts instead of 500, as Bind timeout will usually mean high load scenario.

### DIFF
--- a/api-controllers/ServiceBrokerApiController.js
+++ b/api-controllers/ServiceBrokerApiController.js
@@ -14,6 +14,7 @@ const PreconditionFailed = errors.PreconditionFailed;
 const NotFound = errors.NotFound;
 const ContinueWithNext = errors.ContinueWithNext;
 const Conflict = errors.Conflict;
+const Timeout = errors.Timeout;
 const CONST = require('../common/constants');
 const eventmesh = require('../data-access-layer/eventmesh');
 const formatUrl = require('url').format;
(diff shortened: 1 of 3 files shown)


api-controllers/ServiceBrokerApiController.js
```diff
@@ -434,7 +438,10 @@ class ServiceBrokerApiController extends FabrikBaseController {
         timeout_in_sec: CONST.OSB_OPERATION.OSB_SYNC_OPERATION_TIMEOUT_IN_SEC
       }))
       .then(done.bind(this))
-      .catch(NotFound, gone);
+      .catch(NotFound, gone)
```
subhankarc: Can we test this scenario and see if/what is the error message ? If message is not set, we should set a message here.

jayeshjk: Following is test output:
```
$ cf csk bp1 mykey1
Creating service key mykey1 for service instance bp1 as admin_service-fabrik...
The service broker rejected the request to https://service-fabrik-internal.cf.gcp-dev18.gcp.sapcloud.io/cf/v2/service_instances/ac310148-4f8a-41fa-918c-e94f7f80d5b5/service_bindings/4dbbcf43-5dcb-434f-85ed-70746a631674. Status Code: 429 Too Many Requests, Body: {}
FAILED
```
Body is empty here, should we send an error description in body? @subhankarc  @amitmalav

amitmalav: IMO it's okay to keep the body empty. `Status Code: 429 Too Many Requests` is anyway shown in the message.

subhankarc: yup. Agree.

PR #742, opened 2019-11-06
[Proposal] Label Selector based Scheduling

subhankarc:


inter-operator/architecture/basic.md
```diff
@@ -603,8 +605,21 @@ Schedulers are basically custom controller running on master cluster watching on
...
+This scheduler schedules instance in the least utilized cluster.
+#### Label Selector based Scheduler
+Label selector based scheduler chooses clusters for a service instance based on the label selector defined. Label selector is a go template defined within the `SFPlan`, The template can be evaluated to a label selector string which the scheduler uses to choose cluster for the service instance provisioning. An example of such a template can be the following.
+```yaml
+  - action: clusterlabel
+    type: gotemplate
+    content: |
+      {{- $organizationGuid := "" }}
+      {{- with .instance.spec.organizationGuid }} {{ $organizationGuid = . }} {{ end }}
+      organization={{ $organizationGuid }}
```
amitmalav: Having an example with multiple conditions will help.
Ex: `organization={{ $organizationGuid }},platform:kubernetes` ... is this valid?

inter-operator/architecture/basic.md
```diff
@@ -603,8 +605,21 @@ Schedulers are basically custom controller running on master cluster watching on
 This is just a sample scheduler suitable only for the single cluster setup. In that case, it schedules all the instances in the one cluster which is part of the setup. It is not suitable for the multi-cluster setup.
 #### Round Robin Scheduler
 As the name suggests, round robin scheduler schedules instances in round robin fashion. At this point, it does not take care of capacity and if interoperator restarts, it loses the state about the next cluster to be scheduled and starts scheduling from the beginning. 
-#### Least filled Scheduler
-This scheduler schedules instance in the least filled cluster.
+#### Least Utilized Scheduler
+This scheduler schedules instance in the least utilized cluster.
+#### Label Selector based Scheduler
```
amitmalav: Label Selector based Scheduler should work as  `Least Utilized Scheduler` if labelSelector string is empty or `clusterlabel` action is not present in the plan. In that case we won't require a separate scheduler for `Least Utilized Scheduler`.

subhankarc: `Label Selector based Scheduler` takes mroe time as it fetches plan and service of the instance during scheduling. If it is made default, it will be executed for all even if someone do not need label selector based scheduling and just need least-utilized one. 

However, this is a good point you have made and with more schedulers coming in, it feels like one scheduler is enough. We had same discussion about default scheduler as well. Let us also discuss this in a bigger round, but in general, I would still prefer keeping them for the sake of flexibility.

inter-operator/architecture/basic.md
subhankarc: Yes, the label selector string works same as the `-l` argument of `kubectl`. A complex template would be 

```
    content: |
      {{- $name := "" }}
      {{- with .instance.metadata.name }} {{ $name = . }} {{ end }}
      {{- $organizationGuid := "" }}
      {{- with .instance.spec.organizationGuid }} {{ $organizationGuid = . }} {{ end }}
      service-name={{ $name }},organization={{ $organizationGuid }}
```

subhankarc:
@PadmaB and @amshuman-kr : Could you please have a look ?

inter-operator/architecture/basic.md
```diff
@@ -603,8 +605,21 @@ Schedulers are basically custom controller running on master cluster watching on
...
+Label selector based scheduler chooses clusters for a service instance based on the label selector defined. Label selector is a go template defined within the `SFPlan`, The template can be evaluated to a label selector string which the scheduler uses to choose cluster for the service instance provisioning. An example of such a template can be the following.
+```yaml
+  - action: clusterlabel
```
amshuman-kr:
```suggestion
  - action: clusterSelector
```

Also, is `action` the right way to capture this? It is not really an *action*. Besides, is there any possibility to support anything other than `gotemplate` type here?

inter-operator/architecture/basic.md
amshuman-kr: If the rendered `clusterSelector` matches more than one cluster, even the `Label Selector based Scheduler` should apply the `Least Utilised Scheduler` logic on the matching clusters.

We can think about optimising the case where the `clusterSelector` is known to be *empty*.

subhankarc: > If the rendered clusterSelector matches more than one cluster, even the Label Selector based Scheduler should apply the Least Utilised Scheduler logic on the matching clusters.

The current poc is implemented that way. Do you recommend we then have one Scheduler only? Ideally, as @amitmalav also vouched for it, I would also want it that way. I was thinking of more step by step ways of removing it.

inter-operator/architecture/basic.md
subhankarc: These are certainly not `osb` actions. But I think the current set of actions are also convoluted, with `status` and `sources` template. Type would mostly be `gotemplate` only. One option is use something else other than `templates`.

inter-operator/architecture/basic.md
amshuman-kr: > Do you recommend we then have one Scheduler only?

I agree. I think we can optimize the special case where the `clusterSelector` is empty.

PR #743, opened 2019-11-07
[interoperator] Retry on Update for delete Failure

vivekzhere:
- Retry update on failure

Bug: NGPBUG-98685

### DIFF
--- a/interoperator/pkg/internal/resources/resources.go
+++ b/interoperator/pkg/internal/resources/resources.go
@@ -18,6 +18,7 @@ import (
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/types"
+	"k8s.io/client-go/util/retry"
 	kubernetes "sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
 	logf "sigs.k8s.io/controller-runtime/pkg/runtime/log"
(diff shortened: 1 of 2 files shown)


interoperator/pkg/internal/resources/resources.go
```diff
@@ -491,26 +492,33 @@ func (r resourceManager) deleteSubResource(client kubernetes.Client, resource *u
...
-			content := resource.UnstructuredContent()
-			statusInt, ok := content["status"]
-			var status map[string]interface{}
-			if ok {
-				status, ok = statusInt.(map[string]interface{})
-				if !ok {
-					return fmt.Errorf("status field not map for resource %v", resource)
+
+			err := retry.RetryOnConflict(retry.DefaultRetry, func() error {
+				err := client.Get(context.TODO(), namespacedName, resource)
```
amitmalav: Will it be a good idea to ignore `IsNotFound ` errors and not retry in that case? (It will retry [10 times](https://github.com/kubernetes/client-go/blob/master/util/retry/util.go#L28))

interoperator/pkg/internal/resources/resources.go
```diff
@@ -491,26 +492,33 @@ func (r resourceManager) deleteSubResource(client kubernetes.Client, resource *u
...
-			}
-			content := resource.UnstructuredContent()
-			statusInt, ok := content["status"]
-			var status map[string]interface{}
-			if ok {
-				status, ok = statusInt.(map[string]interface{})
-				if !ok {
-					return fmt.Errorf("status field not map for resource %v", resource)
+
+			err := retry.RetryOnConflict(retry.DefaultRetry, func() error {
```
subhankarc: Wouldn't it retry every-time an error occurs ? Also the `DefaultRetry` seems to be 5 times.

Is it possible to do it only for the specific error type ?

interoperator/pkg/internal/resources/resources.go
amitmalav: IMO it will end up block sync operations like docker, bindings

vivekzhere: Retry happens only on Conflict error

interoperator/pkg/internal/resources/resources.go
vivekzhere: Retry happens only on Conflict error

PR #746, opened 2019-11-13
Override canaries and max_in_flight bosh properties

RandolfWerner:
To speedup an initial bosh deployment it can be useful to override the "canaries" and "max_in_flight" settings defined in the manifest, which are typically optimized to allow downtime free updates. bosh cli and REST API offer this feature. The feature was temporarily broken in the bosh director but now it is fixed:

https://github.com/cloudfoundry/bosh/issues/2193#issuecomment-514406655
https://github.com/cloudfoundry/bosh/commit/756d6f0fcab557708149bdc1392bb106298204c9#diff-3eaf271cb508f0088110a241848205cd

This change allows to specify plan.manager.settings.canaries and plan.manager.settings.max_in_flight in a service plan. If done so this settings are passed to the bosh director when creating a new deployment, updates are not effected an continue using the settings from the deployment manifest.

This small change allow us to significantly speedup the deployment of ABAP systems in our Steampunk project.

### DIFF
--- a/data-access-layer/bosh/BoshDirectorClient.js
+++ b/data-access-layer/bosh/BoshDirectorClient.js
@@ -460,7 +460,7 @@ class BoshDirectorClient extends HttpClient {
   }
 
   createOrUpdateDeployment(action, manifest, opts, scheduled) {
-    const query = opts ? _.pick(opts, 'recreate', 'skip_drain', 'context') : undefined;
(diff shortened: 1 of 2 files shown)


amitmalav:
@RandolfWerner Can you please fix build. Looks like there are some `eslint` errors. PR looks good otherwise.

PR #749, opened 2019-11-19
Avoid defer statement when notifying the WaitGroup

shivansh:


shivansh:
My bad - I had some misconception on how [this](https://github.com/cloudfoundry-incubator/service-fabrik-broker/pull/749/files#diff-8a3394cd2755cd1923a58367ba3af50cR85) line behaves on failure. Closing!

PR #757, opened 2019-11-21
Update basic.md

subhankarc:


subhankarc:
@vivekzhere : Good point, somehow I missed it. Added it now.

jayeshjk:
@subhankarc @vivekzhere Should we also specify that we support only gotemplate for now for label-selector scheduler? Refer [this](https://github.com/cloudfoundry-incubator/service-fabrik-broker/blob/master/interoperator/controllers/schedulers/sflabelselectorscheduler/sflabelselectorscheduler_controller.go#L144-L148)

subhankarc:
@jayeshjk I do not think it is needed. We would try to make it implicit later on, maybe.

PR #761, opened 2019-11-26
[interoperator] Add sprig libary to gotemplates

vivekzhere:
- Provide a hierarchical structure for SFPlan's Context attribute
- Removed our fns which were exact copy
- b64dec is not exact copy. So kept both b64dec and b64enc
- Modify tests

Feature: HCPCFS-2483


vivekzhere:
Somehow #734 was not included in kubebuilder 2.0 changes

PR #767, opened 2019-12-06
Use sfserviceinstance in dashboard template

amitmalav:


api-controllers/ServiceBrokerApiController.js
```diff
@@ -215,7 +216,8 @@ class ServiceBrokerApiController extends FabrikBaseController {
           return eventmesh.apiServerClient.getOSBResourceOperationStatus(lastOperationState);
         }
       })
-      .then(done);
+      .then(() => _.get(context, 'plan.manager.settings.dashboard_url_template') !== undefined ? eventmesh.apiServerClient.waitTillInstanceIsScheduled(req.params.instance_id) : {})
```
subhankarc: update does not need it. You'll have `clusterId` set already.

broker/config/settings.yml
```diff
@@ -782,7 +782,7 @@ defaults: &defaults
           update_predecessors:
           - 'bc158c9a-7934-401e-94ab-057082a5073f'
           template: <%= base64_template('templates/blueprint-manifest.yml.ejs') %>
-          dashboard_url_template: <%= new Buffer('http://service-fabrik-broker.bosh-lite.com/manage/dashboards/director/instances/${instance_id}?planName=${plan.name}&serviceId=${plan.service.id}').toString('base64') %>
+          dashboard_url_template: <%= new Buffer('${spec.clusterId == 1 ? \'http://service-fabrik-broker.bosh-lite.com/manage/dashboards/director/instances/\'+metadata.name+\'?planId=\'+spec.planId+\'&serviceId=\'+spec.serviceId : \'\'}').toString('base64') %>
```
subhankarc: I am little hesitant about two things here.

1. We are not passing context anymore. Are we missing out any attribute ? I guess no.
2. ABAP colleagues use this feature. They have to adapt when changing to this version as the keys are changed. 

Should we keep context and add `sfserviceinstance`'s `clusterId` as it's sub attribute ?

common/constants.js
```diff
@@ -52,6 +52,7 @@ module.exports = Object.freeze({
     UPDATE_EVENT: 'update_instance'
   },
   EVENTMESH_POLLER_DELAY: 200,
+  CLUSTER_SCHEDULER_DELAY: 50,
```
subhankarc: Isn't it too low? let us keep 200 ms

data-access-layer/eventmesh/ApiServerClient.js
```diff
@@ -206,6 +206,26 @@ class ApiServerClient {
         };
       });
   }
+  /**
+     * @description Waits till clusterId is set for sfserviceinstance
+     * @param {string} resourceId - id of resource
+     */
+  waitTillInstanceIsScheduled(resourceId) {
```
vivekzhere: Should we add a timeout for the wait also

data-access-layer/eventmesh/ApiServerClient.js
```diff
@@ -206,6 +206,26 @@ class ApiServerClient {
...
+    assert.ok(resourceId, 'Argument \'resourceId\' is required to get scheduled cluster for instance');
+    logger.debug(`Waiting for scheduler to set ClusterId on ${resourceId}`);
+    return Promise.delay(CONST.CLUSTER_SCHEDULER_DELAY)
+      .then(() => this.getResource({
+        resourceGroup: CONST.APISERVER.RESOURCE_GROUPS.INTEROPERATOR,
+        resourceType: CONST.APISERVER.RESOURCE_TYPES.INTEROPERATOR_SERVICEINSTANCES,
+        resourceId: resourceId,
+        namespaceId: this.getNamespaceId(resourceId)
+      }))
+      .then(sfserviceinstance => _.get(sfserviceinstance, 'spec.clusterId') ? sfserviceinstance : this.waitTillInstanceIsScheduled(resourceId))
```
subhankarc: We should break after 50/55 seconds

broker/config/settings.yml
amitmalav: Changed it such that status quo is maintained. Now you can access `sfserviceinstance` values inside template by using `instance` objejct
Example: `instance.spec.clusterId`

api-controllers/ServiceBrokerApiController.js
amitmalav: Done it considering in future if operators want to change dashboard on plan update etc.

common/constants.js
amitmalav: Delay will be too much if we keep it 200 I feel. `100`?

subhankarc: okay

vinaybheri:
Validation on k8s cluster: n failed.

vinaybheri:

vinaybheri:

PR #775, opened 2019-12-10
[interoperator] Remote Helm support

vivekzhere:
- Use helm3
- Add support for remote helm chart
- URL to tgz supported
- Use gotemplate to renders values.yaml
- content/contentEncoded must contain template to genarate values.yaml
- Remove helm from status template
- Add unit tests

Feature: HCPCFS-2577


vivekzhere:
@amshuman-kr Remote helm support is added with this PR. Please review

interoperator/internal/renderer/helm/renderer.go
```diff
@@ -159,3 +187,25 @@ func (r *helmRenderer) renderResources(ch *chartapi.Chart, values chartutil.Valu
 		Files: files,
 	}, nil
 }
+
+func (r *helmRenderer) processCRDS(ch *chartapi.Chart) (map[string]string, error) {
+	crds := make(map[string]string)
+	for _, cr := range ch.CRDs() {
+		//Add chart name so that crds are applied first
```
amshuman-kr: Why should CRDs be applied first?

interoperator/internal/renderer/helm/samples/postgresql/crds/sample.yaml
```diff
@@ -0,0 +1,5 @@
+apiVersion: apiextensions.k8s.io/v1beta1
+kind: CustomResourceDefinition
+metadata:
+  creationTimestamp: null
+  name: abc
```
amshuman-kr: What's this?

interoperator/internal/renderer/helm/renderer.go
vivekzhere: The templates can have instances of these CRDs.

interoperator/internal/renderer/helm/samples/postgresql/crds/sample.yaml
vivekzhere: Some sample for testing used in the unit tests

interoperator/internal/renderer/helm/renderer.go
amshuman-kr: Ok. But why do they need to be applied first?

vivekzhere: Helm installs the CRDs first. Then installs the other templates. So I thought we will also do the same. https://helm.sh/docs/topics/charts/#custom-resource-definitions-crds

Since we apply resources serially and retry starting from beginning again on first failure, we will need the order to be correct. If a cr instance is tried to be created before the crd, it will fail and keep on failing on retries because the resources will tried to be created in the same order(cr instance before crd) in the retry also.

amshuman-kr: 👍 We should follow the order that helm follows

amshuman-kr:
Just before I forget, please check that the helm packages we use for rendering also support Lua scripts if they are used in the charts.

interoperator/internal/renderer/factory/factory.go
```diff
@@ -96,9 +100,6 @@ func GetStatusRendererInput(template *osbv1alpha1.TemplateSpec, name types.Names
 	}
 
 	switch rendererType {
-	case "helm", "Helm", "HELM":
```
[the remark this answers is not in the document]
vivekzhere: I dont think helm charts for status template makes much sense. Thats why I removed it. I can add it back.

interoperator/internal/renderer/factory/factory.go
```diff
@@ -64,21 +64,25 @@ func GetRendererInput(template *osbv1alpha1.TemplateSpec, service *osbv1alpha1.S
...
+	if template.Content != "" {
+		content = template.Content
+	} else if template.ContentEncoded != "" {
+		decodedContent, err := base64.StdEncoding.DecodeString(template.ContentEncoded)
+		if err != nil {
+			return nil, fmt.Errorf("unable to decode base64 content %v", err)
+		}
+		content = string(decodedContent)
+	}
+	if content == "" {
```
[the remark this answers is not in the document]
vivekzhere: I kept values as mandatory (which is not actually) for helm chart. Will make it optional.

interoperator/internal/renderer/factory/factory.go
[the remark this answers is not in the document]
subhankarc: Agree. Let us not have it then.

vivekzhere:
> 
> 
> Just before I forget, please check that the helm packages we use for rendering also support Lua scripts if they are used in the charts.

From helm/helm#5084
>The current plan is to ensure we lay the groundwork in the Go SDK, allowing the Lua library APIs to be added post-3.0. 

So the lua support is not yet added to helm. What I understand is we will able to continue using `e := engine.New()` and `e.Render()`

PR #783, opened 2020-01-16
Adding support for Mass Update of service instances

subhankarc:
* Added a flag in SFPlan for setting autoUpdatable as true if one wants to mass update serviceinstances when sfplan changes.

* Added a property called specHash in SFPlan to check if the spec change is addressed by the sfserviceinstanceupdater controller.

* Added a field indexer in the main.go to index sfserviceinstances based on spec.planID field.

* Added SFServiceInstanceUpdater controller as part of scheduler. This controller watches on the SFPlans and for any change in spec, it sets the status of all the sfserviceinstances to "update", if the autoUpdateable feature of that plan is turned on.

* Added display-name as additional printer column for SFPlan and SFService.

### DIFF
--- a/interoperator/api/osb/v1alpha1/sfplan_types.go
+++ b/interoperator/api/osb/v1alpha1/sfplan_types.go
@@ -71,30 +71,33 @@ type ServiceSchemas struct {
 
 // SFPlanSpec defines the desired state of SFPlan
 type SFPlanSpec struct {
-	Name          string                `json:"name"`
-	ID            string                `json:"id"`
-	Description   string                `json:"description"`
-	Metadata      *runtime.RawExtension `json:"metadata,omitempty"`
-	Free          bool                  `json:"free"`
-	Bindable      bool                  `json:"bindable"`
-	PlanUpdatable bool                  `json:"planUpdatable,omitempty"`
-	Schemas       *ServiceSchemas       `json:"schemas,omitempty"`
-	Templates     []TemplateSpec        `json:"templates"`
-	ServiceID     string                `json:"serviceId"`
-	RawContext    *runtime.RawExtension `json:"context,omitempty"`
-	Manager       *runtime.RawExtension `json:"manager,omitempty"`
+	Name                string                `json:"name"`
+	ID                  string                `json:"id"`
+	Description         string                `json:"description"`
+	Metadata            *runtime.RawExtension `json:"metadata,omitempty"`
+	Free                bool                  `json:"free"`
+	Bindable            bool                  `json:"bindable"`
+	PlanUpdatable       bool                  `json:"planUpdatable,omitempty"`
+	AutoUpdateInstances bool                  `json:"autoUpdateInstances,omitempty"`
+	Schemas             *ServiceSchemas       `json:"schemas,omitempty"`
+	Templates           []TemplateSpec        `json:"templates"`
+	ServiceID           string                `json:"serviceId"`
+	RawContext          *runtime.RawExtension `json:"context,omitempty"`
+	Manager             *runtime.RawExtension `json:"manager,omitempty"`
 	// Add supported_platform field
 }
 
 // SFPlanStatus defines the observed state of SFPlan
 type SFPlanStatus struct {
-	// INSERT ADDITIONAL STATUS FIELD - define observed state of cluster
-	// Important: Run "make" to regenerate code after modifying this file
+	SpecHash string `json:"specHash,omitempty"`
 }
 
 // +kubebuilder:object:root=true
 // +genclient
 // +genclient:noStatus
+// +kubebuilder:subresource:status
+// +kubebuilder:printcolumn:name="display-name",type=string,JSONPath=`.spec.name`
+// +kubebuilder:printcolumn:name="age",type=date,JSONPath=`.metadata.creationTimestamp`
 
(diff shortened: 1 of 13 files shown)


interoperator/controllers/schedulers/sfserviceinstanceupdater/sfserviceinstanceupdater_controller.go
```diff
@@ -0,0 +1,148 @@
...
+	}
+
+	if plan.Spec.AutoUpdatable == false {
+		arrBytes := []byte{}
+		jsonBytes, _ := json.Marshal(plan.Spec)
+		arrBytes = append(arrBytes, jsonBytes...)
+		hash := md5.Sum(arrBytes)
+		currentSpecHash := base64.StdEncoding.EncodeToString(hash[:])
+		if plan.Status.SpecHash == "" {
+			err = retry.RetryOnConflict(retry.DefaultRetry, func() error {
```
vivekzhere: Should we do the hash calculation also within the retry?

interoperator/controllers/schedulers/sfserviceinstanceupdater/sfserviceinstanceupdater_controller.go
```diff
@@ -0,0 +1,148 @@
...
+	ctrl "sigs.k8s.io/controller-runtime"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/controller"
+)
+
+// SFServiceInstanceUpdater reconciles a SFServiceInstance object
+type SFServiceInstanceUpdater struct {
+	client.Client
+	Log             logr.Logger
+	scheme          *runtime.Scheme
```
vivekzhere: Scheme and clusterRegistry  not used. Can be removed.

interoperator/controllers/schedulers/sfserviceinstanceupdater/sfserviceinstanceupdater_controller.go
```diff
@@ -0,0 +1,148 @@
...
+					}
+					instance.Status.State = "update"
+					return r.Update(ctx, &instance)
+				})
+				if err != nil {
+					log.Error(err, "Error occured while updating the instance", "instance-name", instance.GetName())
+					return ctrl.Result{}, err
+				}
+			}
+			err = retry.RetryOnConflict(retry.DefaultRetry, func() error {
```
vivekzhere: Should we do the hash calculation also within the retry?

interoperator/api/osb/v1alpha1/sfplan_types.go
```diff
@@ -78,6 +78,7 @@ type SFPlanSpec struct {
 	Free          bool                  `json:"free"`
 	Bindable      bool                  `json:"bindable"`
 	PlanUpdatable bool                  `json:"planUpdatable,omitempty"`
+	AutoUpdatable bool                  `json:"autoUpdatable,omitempty"`
```
vivekzhere: Its probably better to have the default behavior(when it is not set at all) as not doing autoUpdate. So the field name does not feel right. We should do auto update only if field is explicitly set to true.

interoperator/controllers/schedulers/sfserviceinstanceupdater/sfserviceinstanceupdater_controller.go
subhankarc: makes sense. I'll change it.

interoperator/controllers/schedulers/sfserviceinstanceupdater/sfserviceinstanceupdater_controller.go
subhankarc: 👍

interoperator/controllers/schedulers/sfserviceinstanceupdater/sfserviceinstanceupdater_controller.go
subhankarc: 👍

interoperator/api/osb/v1alpha1/sfplan_types.go
subhankarc: I am not sure about the naming, but you are right, it should kick in only when it is true. So if it is not set(which means it is false), it should not kick in. I have changed that logic. But, for the naming, let's discuss on that.

interoperator/controllers/schedulers/sfserviceinstanceupdater/sfserviceinstanceupdater_controller.go
subhankarc: done. Can you please check if it looks good now ?

interoperator/controllers/schedulers/sfserviceinstanceupdater/sfserviceinstanceupdater_controller.go
subhankarc: done. Can you please check if it looks good now ?

interoperator/controllers/schedulers/sfserviceinstanceupdater/sfserviceinstanceupdater_controller.go
subhankarc: done. Can you please check if it looks good now ?

PR #803, opened 2020-04-01
Refactor Quota Manager

amitmalav:


api-controllers/ServiceBrokerApiController.js
```diff
@@ -101,6 +101,10 @@ class ServiceBrokerApiController extends FabrikBaseController {
         metadata: {
           finalizers: [`${CONST.APISERVER.FINALIZERS.BROKER}`]
         },
+        labels: {
+          organizationGuid: _.get(params, 'organization_guid'),
```
subhankarc, line 105: Let us make it generic. We put everything that is there in the context. Just loop through context ad put everything. That should also include 'organization_guid'

amitmalav: Done.

broker/lib/middleware.js
```diff
@@ -105,7 +105,7 @@ exports.checkQuota = function () {
   function shouldCheckQuotaForPlatform(platform, origin) {
     return (platform === CONST.PLATFORM.CF ||
       (platform === CONST.PLATFORM.SM &&
-        origin === CONST.PLATFORM.CF));
+        (origin === CONST.PLATFORM.CF || origin === CONST.PLATFORM.K8S)));
```
subhankarc, line 108: This condition basically means everything, right ? Should we remove this check altogether ?

amitmalav: `platform === CONST.PLATFORM.K8S` is not supported. Or shall we support it?

subhankarc: We do not have a usecase for it as of now...but if we have, why not! One can anyways turn off quota check with an additional flag. right ?

Current code would technically work even for `platform === CONST.PLATFORM.K8S` provided the quota APIs support it.

amitmalav: Makes sense, I will remove the condition.

amitmalav: Done

amitmalav:
Tested `labelSelector` string, it works.

subhankarc:
@jayeshjk @vivekzhere Can you guys please review ?

quota/BaseQuotaManager.js
```diff
@@ -0,0 +1,88 @@
...
+      }
+    });
+    logger.debug('sameskuplanids are ', planIdsWithSameSKU);
+    return planIdsWithSameSKU;
+  }
+
+  async isOrgWhitelisted(orgId) {
+    return false;
+  }
+
```
jayeshjk, line 76: Will it make sense to move this function to `K8SPlatformQuotaManager`, because IIUC, this is meant to handle the k8s consumption case and if we decide to handle some whitelisting for k8s case, this function will come handy. Also I think we can loose `async` here.

amitmalav: Removing from Base class means child classes will have to necessarily need to implement this function, which is not intended, async I have put just to maintain parity.

quota/cf-platform-quota-manager/CFPlatformQuotaManager.js
```diff
@@ -0,0 +1,42 @@
...
+    const instances = await cf.cloudController.getServiceInstancesInOrgWithPlansGuids(orgId, planGuids);
+    return _.size(instances);
+  }
+
+  async isOrgWhitelisted(orgId) {
+    const org = await cf.cloudController.getOrganization(orgId);
+    logger.debug('current org details are ', org);
+    logger.debug('current org name is ', org.entity.name);
+    logger.debug('Whitelisted orgs are ', config.quota.whitelist);
+    return await _.includes(config.quota.whitelist, org.entity.name);
```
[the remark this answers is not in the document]
amitmalav: Done.

PR #809, opened 2020-04-12
Adding flag to conditionally disable iptables_manager job in services

jayeshjk:


data-access-layer/bosh/manifest/Addons.js
```diff
@@ -24,6 +25,15 @@ class Addons {
     }
   }
 
+  shouldEnableConnectivity() {
+    const connectivityAllowedServicesList = _.get(this.context, 'connectivityAllowedServicesList');
```
subhankarc, line 29: Is this necessary ? If this is for test, then can we not emulate with the params rather ?

jayeshjk: My understanding was that we need to do this for only postgresql, hence this extra check. WDYT from that perspective?

subhankarc: yeah...makes total sense not to open up for all. 🥇

operators/bosh-operator/DirectorService.js
```diff
@@ -953,6 +953,8 @@ class DirectorService extends BaseDirectorService {
   }
 
   configureAddOnJobs(serviceManifest, context) {
+    _.set(context, 'serviceId', this.service.id);
+    _.set(context, 'connectivityAllowedServicesList', ['6db542eb-8187-4afc-8a85-e08b4a3cc24e']);
```
subhankarc, line 957: Can we remove this hardcoded part then ?

jayeshjk: sure, I will modify to accept this as part of service catalog attribute, and will enable for postgresql for now.

PR #815, opened 2020-04-21
Making namespace based separation configurable

subhankarc:
SF/Interoperator currently creates CRs either all in default namespaces, or creates namespaces for each service instance and puts the CRs there. With this change, we also enable a configurable namespace instead of default namespace.

* If the broker has `apiserver. enable_namespaced_separation` flag as true, it creates separate namespace, else if it has `apiserver.services_namespace` set, it takes that as the namespace name. otherwise, it goes to the default namespace(BOSH based usecase).

* Helm chart now supports both these values, `enable_namespaced_separation` which is by default true and `services_namespace` which is not used by default.

### DIFF
--- a/data-access-layer/eventmesh/src/ApiServerClient.js
+++ b/data-access-layer/eventmesh/src/ApiServerClient.js
@@ -329,7 +329,7 @@ class ApiServerClient {
    */
   createNamespace(name) {
     assert.ok(name, 'Property \'name\' is required to create namespace');
-    if (name === CONST.APISERVER.DEFAULT_NAMESPACE) {
+    if (!_.get(config, 'apiserver.enable_namespaced_separation')) {
       return Promise.resolve();
     }
     const resourceBody = {
@@ -357,19 +357,20 @@ class ApiServerClient {
   }
 
   getNamespaceId(resourceId) {
-    return _.get(config, 'apiserver.enable_namespace') ? `sf-${resourceId}` : CONST.APISERVER.DEFAULT_NAMESPACE;
+    return _.get(config, 'apiserver.enable_namespaced_separation') ? `sf-${resourceId}` : (_.get(config, 'apiserver.services_namespace') ? _.get(config, 'apiserver.services_namespace') : CONST.APISERVER.DEFAULT_NAMESPACE);
   }
 
   /**
    * @description Gets secret
    * @param {string} secretId - Secret Id
-   * @param {string} namespaceId - Optional; Namespace id if given
+   * @param {string} namespaceId - mandatory namespaceId
    */
   getSecret(secretId, namespaceId) {
     assert.ok(secretId, 'Property \'secretId\' is required to get Secret');
+    assert.ok(namespaceId, 'Property \'namespaceId\' is required to get Secret');
     return Promise.try(() => apiserver
       .api[CONST.APISERVER.SECRET_API_VERSION]
-      .namespaces(namespaceId ? namespaceId : CONST.APISERVER.DEFAULT_NAMESPACE)
+      .namespaces(namespaceId)
       .secrets(secretId).get())
       .then(secret => secret.body)
       .catch(err => {
@@ -540,7 +541,7 @@ class ApiServerClient {
     return Promise.try(() => apiserver.apis[opts.resourceGroup][CONST.APISERVER.API_VERSION]
       .namespaces(namespaceId)[opts.resourceType](opts.resourceId).delete())
       .then(res => {
-        if (namespaceId !== CONST.APISERVER.DEFAULT_NAMESPACE && opts.resourceType === CONST.APISERVER.RESOURCE_TYPES.INTEROPERATOR_SERVICEINSTANCES) {
+        if (_.get(config, 'apiserver.enable_namespaced_separation') && opts.resourceType === CONST.APISERVER.RESOURCE_TYPES.INTEROPERATOR_SERVICEINSTANCES) {
           return this.deleteNamespace(namespaceId);
         }
         return res;
@@ -589,7 +590,7 @@ class ApiServerClient {
     assert.ok(opts.resourceGroup, 'Property \'resourceGroup\' is required to get resource');
     assert.ok(opts.resourceType, 'Property \'resourceType\' is required to get resource');
     assert.ok(opts.resourceId, 'Property \'resourceId\' is required to get resource');
-    const namespaceId = opts.namespaceId ? opts.namespaceId : CONST.APISERVER.DEFAULT_NAMESPACE;
+    const namespaceId = opts.namespaceId ? opts.namespaceId : this.getNamespaceId(opts.resourceId);
     return Promise.try(() => apiserver.apis[opts.resourceGroup][CONST.APISERVER.API_VERSION]
       .namespaces(namespaceId)[opts.resourceType](opts.resourceId).get())
       .then(resource => {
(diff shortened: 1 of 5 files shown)


data-access-layer/eventmesh/src/ApiServerClient.js
```diff
@@ -630,7 +631,7 @@ class ApiServerClient {
     if (opts.query) {
       query.qs = opts.query;
     }
-    const namespaceId = opts.namespaceId ? opts.namespaceId : CONST.APISERVER.DEFAULT_NAMESPACE;
+    const namespaceId = opts.namespaceId ? opts.namespaceId : CONST.APISERVER.DEFAULT_NAMESPACE; // Currently most callers are calling this function with allNamespaces: true, only metering jobs are calling without NS and defaults to. Should not be used in any other context.
```
vivekzhere, line 634: Should we check if `apiserver.services_namespace` is set in config?
Like

 `const namespaceId = opts.namespaceId ? opts.namespaceId : (_.get(config, 'apiserver.services_namespace') ? _.get(config, 'apiserver.services_namespace') : CONST.APISERVER.DEFAULT_NAMESPACE);`

data-access-layer/eventmesh/src/ApiServerClient.js
```diff
@@ -691,7 +692,7 @@ class ApiServerClient {
       data: data
     };
     return Promise.try(() => apiserver.api[CONST.APISERVER.CONFIG_MAP.API_VERSION]
-      .namespaces(CONST.APISERVER.DEFAULT_NAMESPACE)[CONST.APISERVER.CONFIG_MAP.RESOURCE_TYPE].post({
+      .namespaces(CONST.APISERVER.DEFAULT_NAMESPACE)[CONST.APISERVER.CONFIG_MAP.RESOURCE_TYPE].post({ // Currently only admin controller calls this to create configs in default NS. Should not be used in any other context.
```
vivekzhere, line 695: Same query as above

data-access-layer/eventmesh/src/ApiServerClient.js
```diff
@@ -702,7 +703,7 @@ class ApiServerClient {
   getConfigMapResource(configName) {
     logger.debug('Get resource with opts: ', configName);
     return Promise.try(() => apiserver.api[CONST.APISERVER.CONFIG_MAP.API_VERSION]
-      .namespaces(CONST.APISERVER.DEFAULT_NAMESPACE)[CONST.APISERVER.CONFIG_MAP.RESOURCE_TYPE](configName).get())
+      .namespaces(CONST.APISERVER.DEFAULT_NAMESPACE)[CONST.APISERVER.CONFIG_MAP.RESOURCE_TYPE](configName).get()) // Currently only admin controller calls this to create configs in default NS. Should not be used in any other context.
```
vivekzhere, line 706: Same query as above

data-access-layer/eventmesh/src/ApiServerClient.js
```diff
@@ -728,7 +729,7 @@ class ApiServerClient {
         resourceBody.data = oldResourceBody.data ? _.merge(oldResourceBody.data, data) : resourceBody.data;
         resourceBody.metadata.resourceVersion = oldResourceBody.metadata.resourceVersion;
         return apiserver.api[CONST.APISERVER.CONFIG_MAP.API_VERSION]
-          .namespaces(CONST.APISERVER.DEFAULT_NAMESPACE)[CONST.APISERVER.CONFIG_MAP.RESOURCE_TYPE](configName).patch({
+          .namespaces(CONST.APISERVER.DEFAULT_NAMESPACE)[CONST.APISERVER.CONFIG_MAP.RESOURCE_TYPE](configName).patch({ // Currently only admin controller calls this to create configs in default NS. Should not be used in any other context.
```
vivekzhere, line 732: Same query as above

data-access-layer/eventmesh/src/ApiServerClient.js
```diff
@@ -1033,15 +1034,15 @@ class ApiServerClient {
     const resourceType = crd.kind === 'SFService' ? CONST.APISERVER.RESOURCE_TYPES.INTEROPERATOR_SERVICES : CONST.APISERVER.RESOURCE_TYPES.INTEROPERATOR_PLANS;
     return Promise.try(() => apiserver
       .apis[CONST.APISERVER.RESOURCE_GROUPS.INTEROPERATOR][CONST.APISERVER.API_VERSION]
-      .namespaces(CONST.APISERVER.DEFAULT_NAMESPACE)[resourceType].post({
+      .namespaces(CONST.APISERVER.DEFAULT_NAMESPACE)[resourceType].post({ // Default NS is used in this context since it is only done for the BOSH usecase, shouldn;t be used in any other context.
```
vivekzhere, line 1037: Same query as above

data-access-layer/eventmesh/src/ApiServerClient.js
```diff
@@ -1033,15 +1034,15 @@ class ApiServerClient {
...
         body: crd
       }))
       .catch(err => {
         return convertToHttpErrorAndThrow(err);
       })
       .catch(Conflict, () => apiserver
         .apis[CONST.APISERVER.RESOURCE_GROUPS.INTEROPERATOR][CONST.APISERVER.API_VERSION]
-        .namespaces(CONST.APISERVER.DEFAULT_NAMESPACE)[resourceType](crd.metadata.name).patch({
+        .namespaces(CONST.APISERVER.DEFAULT_NAMESPACE)[resourceType](crd.metadata.name).patch({ // Default NS is used in this context since it is only done for the BOSH usecase, shouldn;t be used in any other context.
```
vivekzhere, line 1045: Same query as above

data-access-layer/eventmesh/src/ApiServerClient.js
subhankarc: As I have commented in the code, currently only metering jobs are calling this to fetch the metering CRs, and they are created in default namespaces. Since this is not utilized in the interoperator, I thought of not disturbing it at this moment, I have just added the comment as warning sign. ;)

data-access-layer/eventmesh/src/ApiServerClient.js
subhankarc: same here, these utility functions are used in BOSH based deployment to create some configmaps which are used as flags for shceduled updates. Hence I didn't disturb these and kept them as is, added comment so that it is not used elsewhere.

data-access-layer/eventmesh/src/ApiServerClient.js
subhankarc: same here, these utility functions are used in BOSH based deployment to create some configmaps which are used as flags for shceduled updates. Hence I didn't disturb these and kept them as is, added comment so that it is not used elsewhere.

data-access-layer/eventmesh/src/ApiServerClient.js
subhankarc: same here, these utility functions are used in BOSH based deployment to create some configmaps which are used as flags for shceduled updates. Hence I didn't disturb these and kept them as is, added comment so that it is not used elsewhere.

data-access-layer/eventmesh/src/ApiServerClient.js
subhankarc: same here, these utility functions are used in BOSH based deployment to create SFplans and SFServices from the catalog, which happens separately in the K8s based deployments. Hence I didn't disturb these and kept them as is, added comment so that it is not used elsewhere.

data-access-layer/eventmesh/src/ApiServerClient.js
subhankarc: same here, these utility functions are used in BOSH based deployment to create SFplans and SFServices from the catalog, which happens separately in the K8s based deployments. Hence I didn't disturb these and kept them as is, added comment so that it is not used elsewhere.

PR #828, opened 2020-05-05
Removing pages which are already moved.

subhankarc:


inter-operator/architecture/update.md
```diff
@@ -1,54 +0,0 @@
-# Upgrade Interoperator from the earlier releases(special handling, downtime if any)
```
vivekzhere, line 1: This is not put anywhere else. Should we keep this

subhankarc: I am adding it with https://github.com/cloudfoundry-incubator/service-fabrik-broker/pull/829

PR #834, opened 2020-05-11
Separating quota related computation in separate application

jayeshjk:


broker/applications/osb-broker/src/api-controllers/middleware/QuotaClient.js
```diff
@@ -0,0 +1,43 @@
...
+        Accept: 'application/json'
+      },
+      followRedirect: false
+    }, options));
+    this.username = config.quota_app.username;
+    this.password = config.quota_app.password;
+  }
+
+  async getQuotaValidStatus(options) {
+    const orgOrSubaccountId = _.get(options, 'orgOrSubaccountId');
```
subhankarc, line 26: Ad we discussed earlier, do you plan to use a PUT API in case the quota calculation is dynamic ?

jayeshjk: Yeah, I will add code to determine which endpoint to use based on a setting in plan/service.

subhankarc: ah ok...Sorry for jumping in early before you opened up for review. :)

jayeshjk: No issues.. I was anyways going to ask for a customary look at structure etc before starting on UTs.

broker/applications/osb-broker/src/api-controllers/middleware/QuotaClient.js
```diff
@@ -0,0 +1,64 @@
+/* jshint ignore:start */
```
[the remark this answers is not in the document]
jayeshjk: it's not needed.. removed.

helm-charts/interoperator/values.yaml
```diff
@@ -28,6 +28,23 @@ broker:
...
+    oauthDomain: https://myauth-domain.com
+    serviceDomain: https://my-tenant-onboarding.com
+    username: quota_user
+    password: quota_password
+  image:
+    repository: servicefabrikjenkins/service-fabrik-broker
+    tag: 0.6.0
+    pullPolicy: Always
+  service:
+    type: LoadBalancer
```
vivekzhere, line 46: Default  service type as `ClusterIP` might be better. Since this is only consumed from within the cluster.

helm-charts/interoperator/templates/quota_app.yaml
```diff
@@ -0,0 +1,87 @@
...
+            fieldRef:
+              fieldPath: metadata.namespace
+        volumeMounts:
+        - name: settings
+          mountPath: /opt/sf-config
+        livenessProbe:
+          failureThreshold: 3
+          httpGet:
+            path: /
+            port: 9296
```
vivekzhere, line 63: `{{ .Values.quota_app.port }}` instead of 9296

helm-charts/interoperator/conf/settings.yaml
```diff
@@ -38,6 +38,19 @@ defaults: &defaults
     protocol: http
     host: 10.0.2.3:9293
 
+  ##############################
+  # QUOTA APP SETTINGS #
+  ##############################
+  quota_app:
```
vivekzhere, line 44: Most of these are there in values also. Read from there like `{{ .Values.quota_app.port }}`

helm-charts/interoperator/conf/settings.yaml
```diff
@@ -38,6 +38,19 @@ defaults: &defaults
...
+    port: 9296
+    protocol: http
+    host: {{ .Release.Name }}-quota-service.interoperator.svc.cluster.local
+    username: quota
+    password: secret
+    quota_app_url: http://{{ .Release.Name }}-quota-service.interoperator.svc.cluster.local
+    quota_endpoint: /v1/account
+
   ##############################
   # APISERVER SETTINGS #
```
vivekzhere, line 55: Namespace also must be templated like `{{ .Release.Namespace }}`

helm-charts/interoperator/values.yaml
```diff
@@ -28,6 +28,23 @@ broker:
...
+  password: secret
+  quota:
+    enabled: false
+    oauthDomain: https://myauth-domain.com
+    serviceDomain: https://my-tenant-onboarding.com
+    username: quota_user
+    password: quota_password
+  image:
+    repository: servicefabrikjenkins/service-fabrik-broker
+    tag: 0.6.0
```
subhankarc, line 43: This tag needs to be changed. Also, since the Jenkins release creattion modifies the tag, please check with @vinaybheri to adjust the automation

vivekzhere: Another way, will be to remove the image field totally from quota values and use the same value from broker in the quota app deployment.

subhankarc: good point. We should just use the same for the broker app, since the image is same.

jayeshjk: But I think we still have to update tag under broker, right?

subhankarc: can you rebase ? It should be updated already

subhankarc: https://github.com/cloudfoundry-incubator/service-fabrik-broker/blob/master/helm-charts/interoperator/values.yaml#L26

helm-charts/interoperator/values.yaml
```diff
@@ -28,6 +28,19 @@ broker:
...
+quota_app:
+  port: 9296
+  username: quota
+  password: secret
+  quota:
+    enabled: false
+    oauthDomain: https://myauth-domain.com
+    serviceDomain: https://my-tenant-onboarding.com
+    username: quota_user
+    password: quota_password
```
subhankarc, line 40: Can be removed in favor of LN18-23

PR #847, opened 2020-06-08
Adding default login_hint as sap.ids

jayeshjk:
* It is necessary as currently `login_hint` param is not set when sending authorization request to UAA if it's not provided in the dashboard URL. And with this change by default `sap.ids` will be used as `login_hint`.
* Also this enables adding check for `login_hint` in `validateSession` while dashboard rendering.


jayeshjk:
@abh1kg  @subhankarc I have not tested this yet, hence adding `DO_NOT_MERGE`.

subhankarc:
would you please fix the tests as well @jayeshjk

PR #877, opened 2020-07-03
Change Docker build context

vivekzhere:
- Set context to project root for all images
- Xmake does docker build from project root (With new config)
- Move dockerignore to project root

Feature: HCPCFS-2722


subhankarc:
@vivekzhere Do we need to adapt the Jenkins builds which build docker images ?

vivekzhere:
Re based with Master

PR #924, opened 2020-07-22
conditional startup of admin application

jayeshjk:


jayeshjk:
Not needed anymore as separate app for Interoperator Admin APIs will be created.

PR #982, opened 2020-08-07
Removing org based quota check.

subhankarc:
Keeping only Sub-account based quota check

### DIFF
--- a/broker/applications/osb-broker/src/api-controllers/middleware/QuotaClient.js
+++ b/broker/applications/osb-broker/src/api-controllers/middleware/QuotaClient.js
@@ -28,9 +28,9 @@ class QuotaClient extends HttpClient {
     }
   }
   async getQuotaValidStatus(options) {
-    const orgOrSubaccountId = _.get(options, 'orgOrSubaccountId');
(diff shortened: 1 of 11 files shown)


jayeshjk:
@subhankarc 
Should we also remove `useSubaccountForQuotaCheck` altogether?
This will mean signature change here:  https://github.com/cloudfoundry-incubator/service-fabrik-broker/blob/063d0477af2d494e08193c2a68ff40459d80130a/broker/data-access-layer/quota/src/BaseQuotaManager.js#L17
as well as changes in https://github.com/cloudfoundry-incubator/service-fabrik-broker/blob/063d0477af2d494e08193c2a68ff40459d80130a/broker/applications/quota-app/src/QuotaApiController.js
and 
https://github.com/cloudfoundry-incubator/service-fabrik-broker/blob/063d0477af2d494e08193c2a68ff40459d80130a/broker/applications/osb-broker/src/api-controllers/middleware/index.js#L69

subhankarc:
@jayeshjk I updated the PR based on your comment.

jayeshjk (approved):
LGTM

PR #1000, opened 2020-08-17
Removing scheduler type config and default scheduler for pure Interoperator based deployment

subhankarc:
…eduler


interoperator/controllers/schedulers/setup_schedulers.go
```diff
@@ -1,4 +1,4 @@
-// +build schedulers
+// +build schedulers,default
```
vivekzhere, line 1: In this file line 57, should be removed. (Registering the default scheduler).

subhankarc:
@vivekzhere I made changes as per our discussion. Please have a look.

PR #1001, opened 2020-08-19
Fix delete failure before create over

vivekzhere:
- Sometime delete fails if triggered while create in progress
- Add check to use status temp only in "in progress" state
- Improve logging

Bug: NGPBUG-126050


vivekzhere:
Similar issue can occur in setInProgress function of instance controller. But not confident about a fix. So not adding it to the PR.

subhankarc (approved):

jayeshjk (approved):
LGTM

PR #1011, opened 2020-08-27
[Bug] Fix NS replication when NS sepration is disabled

vivekzhere:
- Do not delete ns if ns seperation is disabled

Bug:


jayeshjk (approved):
LGTM

subhankarc (approved):

PR #1029, opened 2020-09-08
Allow custom error messages for invalid quota & send instance_id

steinri:
This adds the possibility to allow custom error messages if the quota is invalid. The reason for this is that we may want to provide additional information why the entitlements are not sufficient.

In addition it forwards the instance_id to the non-instance based quota check. This is quite important to have for the update case.


jayeshjk (approved):
LGTM

PR #1042, opened 2020-09-15
Multi-AZ and High availability for Interoperator

vivekzhere:
- Adjust Helm Chart Values
- Set default replica to 2
- Lower default cpu requests
- Set resources for broker, quota and operatorapi
- Add HA and multi az docs

Feature: HCPCFS-2472


vivekzhere:
On our dev1 landscape, the max cpu usage in the last 30 days for any interoperator pods was `10m`. So `100m` cpu request should be more than sufficient.

subhankarc (approved):

PR #1048, opened 2020-09-15
Set log level to debug for pre-actions

steinri:
Since the hooks may respond with service instance specific credentials, the response should not be logged on the info level.


jayeshjk (approved):
LGTM

PR #1081, opened 2020-09-28
[Bug] Fix state getting lost on parallel requests

vivekzhere:
- Set in_progress only if no operation happend in between
- Never over write resources array of sister
- Fix UTs
- Add tests

Bug: NGPBUG-127887


vivekzhere:
Duplicate of #1083 1083

PR #1088, opened 2020-10-05
restructured osb-broker UTs

visargsoneji:
Restructured the UTs of `osb-broker` and changed the coverage report generator module to `nyc` for `broker`


visargbsoneji:
ok to test

PR #1089, opened 2020-10-06
osb-broker UTs restructured

visargsoneji:
Restructuring osb-broker UTs and using isparta + istanbul module to generate the coverage report


visargsoneji:
After restructuring the `osb-broker` UTs, the order of execution of `mocha` test changes. Previously, `dashboard.director.spec.js` was executed first and then `osb-broker` UTs. Now the reverse has happened and one of the mocks created in `dashboard.director.spec.js` was not getting hit, resulting into failure of the UT(pending nock). This is a dynamic change and it does not mean any error in the way the UT was written. To solve this, I modified the `verify` function of `mocks` by passing an optional array to ignore the pending mocks passed by the UT itself.

broker/test/test_broker/mocks/index.js
```diff
@@ -44,8 +44,20 @@ function init() {
...
 }
 
-function verify() {
+function verify(ignoreList = []) {
   /* jshint expr:true */
+  if(nock.pendingMocks().length > 0) {
+    if(nock.pendingMocks().length == ignoreList.length) {
+      let count = 0;
+      for(let i = 0; i < ignoreList.length; i++) {
+        if(nock.pendingMocks()[i] == ignoreList[i])
```
jayeshjk, line 53: Implementaion looks good. Only thing is, do we need a sort step before so that we are sure about the ordering of both lists?

broker/test/test_broker/mocks/index.js
```diff
@@ -44,8 +44,20 @@ function init() {
   return exports;
 }
 
-function verify() {
+function verify(ignoreList = []) {
   /* jshint expr:true */
+  if(nock.pendingMocks().length > 0) {
+    if(nock.pendingMocks().length == ignoreList.length) {
```
jayeshjk, line 50: For better readability store `nock.pendingMocks()` in a variable.

broker/test/test_broker/mocks/index.js
```diff
@@ -44,8 +44,20 @@ function init() {
...
+      let count = 0;
+      for(let i = 0; i < ignoreList.length; i++) {
+        if(nock.pendingMocks()[i] == ignoreList[i])
+          count++;
+      }
+      if(count == ignoreList.length)
```
jayeshjk, line 56: Add logger statements for future debugging.

broker/test/test_broker/mocks/index.js
```diff
@@ -44,8 +44,20 @@ function init() {
   return exports;
 }
 
-function verify() {
+function verify(ignoreList = []) {
```
jayeshjk, line 47: Add a single line comment to explain  purpose, format of the `ignoreList` variable

jayeshjk (changes requested):

broker/test/test_broker/mocks/index.js
visargsoneji: We are sending the list from the UT itself. So, this sorting thing does not make much sense as we will be making the list in the order of pending mocks only.

broker/test/test_broker/mocks/index.js
visargsoneji: Sure. Making the changes

jayeshjk (approved):
LGTM

PR #1098, opened 2020-10-12
Support uppercase instance id

vivekzhere:
- If instanceId/bindingId is not in k8s format, use hash
- Use sha224 hash of the id
- Trim label values and validate them
- Add instanceId to SfServiceInstance spec
- Set by broker for new instances
- Set by broker on update of old instances
- Update docs

Feature: HCPCFS-2851

### DIFF
--- a/broker/applications/osb-broker/src/api-controllers/ServiceBrokerApiController.js
+++ b/broker/applications/osb-broker/src/api-controllers/ServiceBrokerApiController.js
@@ -87,7 +87,9 @@ class ServiceBrokerApiController extends FabrikBaseController {
       'space_guid',
       'origin',
       'subaccount_id'
-    ]); 
+    ]);
+    _.set(params, 'instance_id', req.params.instance_id);
+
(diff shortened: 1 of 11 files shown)


loewenstein (commented):
1. I can't judge on the code, i.e. if you hit all the spots where instance ids are coming in, but the idea of hashing with sha224 to force all possible instance id values into a Kubernetes compatible unique id lgtm.
Also not touching it, if the instance id is already Kubernetes compatible, sounds like a reasonable approach for backwards compatibility. 
Although, bumping the CRD version, and providing a conversion function might be considered. That would lead to unified code and CRs, basically making backwards compatibility a one-time effort instead of sticking with the codebase forever.

1. Should changing the CRD to support an `instanceId` field in the spec and a conversion that transfers all existing instance ids from `metadata.name` to that new field be part of this change or done seperately?

subhankarc (approved):

jayeshjk (approved):
LGTM

PR #535, opened 2019-01-03
Removing CF CC dependencies in extension APIs code flow except authentication and authorization

jayeshjk:
* Related backlog is: HCPCFS-2167.

* We can get rid of hard dependencies on Cloud Foundry CC by making use of ApiServer to fetch `space_guid`, `plan_id` and `context`.
 
* Filewise changes are as follows:

**1.  ServiceFabrikApiController**
* We fetch and store `context`, `space_guid` and `plan_id` at the start of processing of `/service_instances/:instance_id` requests, if not present already in the request.
* As the `space_guid` and `plan_id` would _always_ be present in requests now, the calls to CC to fetch them in `setPlan`, `verifyTenantPermission`(only `getServiceInstance` call is removed `getSpaceDeveloper` call still exists) and `startRestore` are removed.
* In `scheduleBackup`, information such as `org_name`, `space_name` etc is used in job parameters. Currently, this logic is slightly altered to fetch this information _conditionally_, i.e., if the platform is Cloud Foundry. 
* Similarly in `scheduleUpdate`, the `instance_name` will be fetched conditionally based on platform value.
* Above two cases can be alternatively dealt by removing this information from job data. Comments on this are requested.
* In `listBackupFiles` the plan details will be fetched from ApiServer instead of CC.

**2.  BackupReaperJob and ScheduleBackupJob**
* `isServiceInstanceDeleted` check will be done using ApiServer.

PR #1111, opened 2020-10-23
Replace Request with Axios in HTTP Clients

anoopjb:
## Using Axios in Clients

As of Feb 11th 2020, request is fully deprecated. For more information about why request is deprecated refer to [this issue](https://github.com/request/request/issues/3142).
Replace request with [Axios](https://github.com/axios/axios) in Clients.

### Difference in Options/Configurations in Request and Axios
|Request|Axios|Remarks|Comments|
|---|---|---|---|
|baseUrl|baseURL|The functionality is same.|`baseURL` is used|
|followRedirect|maxRedirects|The maximum redirects allowed by request module when the `followRedirect` option is `true` is 10. Axios allow any number of redirects. Couldn't find any hard limit. `maxRedirects: 0` disables redirects in Axios.|`maxRedirects` set to 10|
|rejectUnauthorized (or) strictSSL|axios relays on inbuilt http module in Node for SSL. There is no specific option to control this in axios.|This configuration is used to disable strict validation of SSL certificates in request module. It is set to `false` in request to allow using self signed certs.|We can create an https agent with `rejectUnauthorized: false` and pass to axios. Reference [#535](https://github.com/axios/axios/issues/535)|
|Supports bearer authorization|Axios supports only basic auth|Possible through HTTP headers|`{ authorization: 'Bearer 1234' }`|

#### Automatic transforms for JSON data

Axios tries to parse the response body to JSON.
Axios adds a default `transformResponse` handler to parse response body to JSON.
Reference from source is [here](https://github.com/axios/axios/blob/6642ca9aa1efae47b1a9d3ce3adc98416318661c/lib/defaults.js#L57).

So a call to `JSON.parse(response.data)` is redundant and causes failure.

#### Using Basic Auth and Bearer Authorization

We created a client with basic authentication configuration here,
```Javascript
const client = axios.create({
    baseURL: 'https://api.github.com',
    auth: {
        username: "test",
        password: "test"
    }
});
```

If we want to use bearer authentication in some of the requests,
we could disable the auth configuration already set in the client object by passing
`auth: false`. Now pass bearer token as authorization header.
```Javascript
client.request({
    url: '/user/emails',
    auth: false,
    headers: {
        authorization: `Bearer ${githubToken}`
    },
    method: 'POST',
    data: {
        emails: [
            'testemail@test.com'
        ]
    }
})
    .then(function (response) {
        console.log(response.data)
    });
```
Reference from [#1054](https://github.com/axios/axios/issues/1054).

#### Error handling in Axios

Any status code that falls out of the range of 2xx is considered as error in Axios.
A `catch` statement is required to handle these errors.
More information on error handling [here](https://github.com/axios/axios#handling-errors).

### Compatability with Nock
Nock v9.16 onwards supports Axios library.


vinaybheri:

vinaybheri:

vinaybheri:

vinaybheri:

vivekzhere (approved):

subhankarc (approved):

jayeshjk (approved):
LGTM

PR #1120, opened 2020-10-27
Restructured UTs of operators

visargsoneji:


broker/applications/operators/test/operators.BackupService.spec.js
```diff
@@ -199,6 +199,7 @@ describe('operators', function () {
         sandbox.restore();
         mocks.agent.getBackupLogs(fakeLogs);
         mocks.agent.lastBackupOperation(backup_state);
+        mocks.cloudProvider.auth(1);
```
visargsoneji, line 202: Due to change in the order of execution of UTs, the token was expired and so I have to add mock for authentication.

broker/applications/operators/test/operators.BackupService.spec.js
```diff
@@ -222,7 +223,8 @@ describe('operators', function () {
           .then(res => {
             expect(res.description).to.eql(`Backup deployment ${deployment_name} succeeded at ${finishDate}`);
             expect(res.state).to.eql('succeeded');
-            mocks.verify();
+            let ignoreNock = ["POST https://myopenstackcloud.com:5000/v3/auth/tokens"];
+            mocks.verify(ignoreNock);
```
visargsoneji, line 227: Passing the added mock as ignoreList while verifying the pending mocks. This is done to make sure that this UT is not dependent on order of execution. If token is not expired, this mock will not get hit and so it is passed as an optional(ignore) mock.

vivekzhere (approved):

jayeshjk (approved):
LGTM

PR #1123, opened 2020-10-29
Restructured UTs of Utils,SchedularJobs and Osb-broker

visargsoneji:


broker/core/scheduler-jobs/test/jobs.Scheduler.spec.js
```diff
@@ -284,6 +284,7 @@ describe('Jobs', function () {
     });
     after(function () {
       sandbox.restore();
+      logSpy.restore();
```
visargsoneji, line 287: In this UT, `logSpy` which is a sinon spy for logger(error) is not restored. This was creating an issue with other UT in `cf.TokenIssuer.spec.js` as it was trying to wrap the same function. Earlier, it was not creating issue as the order of execution was opposite for these two UTs and the cf UT has already restored the function.

broker/test/init
```diff
@@ -3,10 +3,10 @@
...
-        "test-cov")         babel-node ./node_modules/.bin/isparta cover _mocha -- "test/test_broker/**/*.js" "applications/**/test/**/*.js" --opts test/test_broker/mocha_broker.opts;;
+        "test")             _mocha "test/test_broker/**/*.js" "applications/**/test/**/*.js" "core/**/test/*.js" --opts test/test_broker/mocha_broker.opts;;
+        "test-ci")          babel-node ./node_modules/.bin/isparta cover --report cobertura _mocha -- "test/test_broker/**/*.js" "applications/**/test/**/*.js" "core/**/test/*.js" --opts test/test_broker/mocha_broker.opts;;
+        "test-coverall")    babel-node ./node_modules/.bin/isparta cover --report lcovonly _mocha -- "test/test_broker/**/*.js" "applications/**/test/**/*.js" "core/**/test/*.js" --opts test/test_broker/mocha_broker.opts;;
+        "test-cov")         babel-node ./node_modules/.bin/isparta cover _mocha -- "test/test_broker/**/*.js" "applications/**/test/**/*.js" "core/**/test/*.js" --opts test/test_broker/mocha_broker.opts;;
     esac
```
visargsoneji, line 10: Added the path to restructured `core` UTs

jayeshjk (approved):
LGTM

PR #1132, opened 2020-11-04
Restructured UTs of rest modules of data-access-layer

visargsoneji:
This PR contains restructured UTs of db, docker, event-logger, eventmesh, metering, oob-manager and service-agent modules of data-access-layer


jayeshjk (approved):
LGTM

PR #1134, opened 2020-11-05
Restructured UTs of deployment-hooks

visargsoneji:
This PR contains restructured UTs of `deployment_hooks`


broker/test/init
```diff
@@ -2,21 +2,30 @@
 
 test_broker(){
     rm -f ./logs/test.log
+
+    testBroker="test/test_broker/*.js" #for global.setup.js 
+    core="core/**/test/*.js"
+    dataAccessLayer="data-access-layer/**/test/*.js"
+    app=$(ls applications/**/test/*.js applications/**/test/**/*.js | grep -v "^applications/deployment_hooks") 
+
```
visargsoneji, line 10: We are running `deployment_hooks` separately and they are restructured to `applications` directory now. With the earlier way to run UTs, `deployment_hooks` will also run causing error. Here, I am selecting all the UTs of different `applications` except those of `deployment_hook`

jayeshjk (approved):
LGTM

PR #1139, opened 2020-11-11
Updating bluebird from 3.4.0 to 3.7.2

visargsoneji:
This PR contains changes required to upgrade `bluebird` to 3.7.2

### DIFF
--- a/broker/data-access-layer/bosh/test/bosh.BoshDirectorClient.spec.js
+++ b/broker/data-access-layer/bosh/test/bosh.BoshDirectorClient.spec.js
@@ -1325,10 +1325,14 @@ describe('bosh', () => {
 
         dummyBoshDirectorClient.makeRequestWithConfig({}, 200, directorConfig)
           .then(() => {
-            expect(requestStub).to.be.calledOnce;
+            expect(requestStub).to.be.calledTwice;
             sandbox.restore();
             config.directors = prevConfigDirectors;
             done();
+          })
+          .catch(() => {
+            sandbox.restore();
+            done(new Error('expected success but recieved error'));
           });
       });
     });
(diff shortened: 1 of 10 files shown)


broker/data-access-layer/bosh/test/bosh.BoshDirectorClient.spec.js
```diff
@@ -1325,10 +1325,14 @@ describe('bosh', () => {
 
         dummyBoshDirectorClient.makeRequestWithConfig({}, 200, directorConfig)
           .then(() => {
-            expect(requestStub).to.be.calledOnce;
+            expect(requestStub).to.be.calledTwice;
             sandbox.restore();
```
visargsoneji, line 1329: From the release 3.4.6 of bluebird they have fixed the issue of `Promise.map`. Earlier, it was performing the work synchronously whereas it was supposed to perform it asynchronously. Here, we are using `Promise.map` in the constructor of `BoshDirectorClient` and then we are creating the stubs. In the constructor itself, we are sending the `request` which was going to its actual function in earlier version of bluebird(3.4.0) but now as it is done asynchronously, the first call to `request` method is handled by stubs.

broker/data-access-layer/bosh/test/bosh.BoshDirectorClient.spec.js
```diff
@@ -1325,10 +1325,14 @@ describe('bosh', () => {
...
             config.directors = prevConfigDirectors;
             done();
+          })
+          .catch(() => {
+            sandbox.restore();
+            done(new Error('expected success but recieved error'));
           });
```
visargsoneji, line 1336: In case the promise is rejected, we need to handle that as well because when it fails it does not restores the sandbox and other UTs also fail.

jayeshjk (approved):

PR #1141, opened 2020-11-12
Operator-api and quotaapp deployed conditionally

vivekzhere:
- Deploy these apps only if they are enabled

### Fix metrics scrape in sister cluster
- The metrics svc of provisioner is not replicated
- Add prometheus annotations instead

### Make resources configurable per deployment
- Seperate resources values for each deployment
- Seperate replicaCount config for each deployment

### Requeue SFCluster for mcd periodically
- Requeue every 20 minutes for self healing
- Connection issues might close multi cluster watches

Feature: HCPCFS-2864

### DIFF
--- a/helm-charts/interoperator/values.yaml
+++ b/helm-charts/interoperator/values.yaml
@@ -81,6 +81,7 @@ interoperator:
     provisionerWorkerCount: 2
 
 operator_apis:
+  enabled: true
   port: 9297
   username: admin
   password: secret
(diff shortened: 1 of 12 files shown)


subhankarc (approved):

helm-charts/interoperator/values.yaml
```diff
@@ -68,6 +68,7 @@ interoperator:
     provisionerWorkerCount: 2
 
 operator_apis:
+  enabled: false
```
subhankarc, line 71: not sure if we should keep the default as true or false. Keeping it true would atleast be status quo. Or else, it will suddenly vanish for existing deployments.

subhankarc (approved):

helm-charts/interoperator/values.yaml
vivekzhere: Ok. Will update to true

jayeshjk (approved):

PR #1143, opened 2020-11-12
Revert "Adding stricter Sub account check for quota check"

subhankarc:
Reverts cloudfoundry-incubator/service-fabrik-broker#1107


jayeshjk (approved):
LGTM

PR #1160, opened 2020-11-27
Updating dockerode from 2.2.10 to 3.2.1

visargsoneji:
Breaking change:
Promises interface was standardized, output property was removed.

### DIFF
--- a/broker/applications/operators/src/docker-operator/DockerService.js
+++ b/broker/applications/operators/src/docker-operator/DockerService.js
@@ -790,7 +790,8 @@ class DockerService extends BaseService {
       .logsAsync({
         stdout: 1,
         stderr: 1,
-        timestamps: 1
+        timestamps: 1,
+        follow: true
       })
       .then(stream => demux(stream, {
         tail: 1000
--- a/broker/core/utils/src/commonFunctions.js
+++ b/broker/core/utils/src/commonFunctions.js
@@ -598,6 +598,10 @@ function demux(stream, options) {
     stream.on('readable', onreadable);
     stream.once('end', onend);
     stream.once('error', onerror);
+
+    setTimeout(function() {
+      stream.destroy();
+    }, 2000);
   });
 }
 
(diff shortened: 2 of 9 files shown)


broker/applications/operators/src/docker-operator/DockerService.js
```diff
@@ -790,7 +790,8 @@ class DockerService extends BaseService {
       .logsAsync({
         stdout: 1,
         stderr: 1,
-        timestamps: 1
+        timestamps: 1,
+        follow: true
```
visargsoneji, line 794: To get the logs of the container, we need to pass the `follow` parameter as true which is `false` by default.

broker/core/utils/src/commonFunctions.js
```diff
@@ -598,6 +598,10 @@ function demux(stream, options) {
     stream.on('readable', onreadable);
     stream.once('end', onend);
     stream.once('error', onerror);
+
+    setTimeout(function() {
+      stream.destroy();
+    }, 2000);
```
visargsoneji, line 604: If we are not destroying the stream, it gets stuck causing `502` Bad Gateway error. So, destroying it was a necessary step.

jayeshjk: @visargsoneji  can you pls add examples from dockerode which you referred for this as a comment. Might be helpful as future reference.

broker/applications/operators/src/docker-operator/DockerService.js
jayeshjk: Can you pls add a reference here as well/

visargsoneji:
Reference example from dockerode : https://github.com/apocas/dockerode/blob/master/examples/logs.js
`function containerLogs(container) {

  // create a single stream for stdin and stdout
  var logStream = new stream.PassThrough();
  logStream.on('data', function(chunk){
    console.log(chunk.toString('utf8'));
  });

  container.logs({
    follow: true,
    stdout: true,
    stderr: true
  }, function(err, stream){
    if(err) {
      return logger.error(err.message);
    }
    container.modem.demuxStream(stream, logStream, logStream);
    stream.on('end', function(){
      logStream.end('!stop!');
    });

    setTimeout(function() {
      stream.destroy();
    }, 2000);
  });
}`

jayeshjk (approved):
LGTM

subhankarc (approved):

PR #1193, opened 2020-12-21
distributing external deps across modules and migrating to yarn berry

jayeshjk:
As of now, we collect all the external deps in the package.json of root application modules. But to make each submodule self-contained, it is necessary to distribute these deps across all modules as needed. 
Apart from above changes, we also migrate to Yarn berry from Lerna as monorepo manager. This includes changes in dockerfile as well.
We also add yarn constraint for ensuring same external dependency versions across all yarn workspaces (https://yarnpkg.com/features/constraints).


vinaybheri:

broker/core/scheduler-jobs/src/Scheduler.js
```diff
@@ -36,6 +36,10 @@ class Scheduler {
     this.jobs = {};
     this.time = new Date();
     this.config = config.scheduler;
+    if(!this.config) {
```
jayeshjk, line 39: This check is needed because `const { BaseService } = require('@sf/provisioner-services');` now loads `DirectorService` also. In the jobs such VirtualHostOperator/PostgresqlMTOperator the configs for scheduler is not specified and this causes error when DirectorService also tries to load Scheduler module.

vinaybheri:

subhankarc (approved):

vivekzhere (approved):

PR #1236, opened 2021-02-03
reading primary cluster id from the configmap instead of hardcoded value

jayeshjk:
- As of now, cluster id of the master/primary cluster is hardcoded to `1` . This PR introduces changes so that it can be made configurable.
- To achive this we introduce a new label `interoperator.servicefabrik.io/primarycluster`, such that this label will have value `true` for sfcluster CR of the primary cluster.
- Further we add new config field `primaryClusterId`in the interoperator configmap, which will always store current primary cluster id.
- This configmap field is reconciled `provisioner_controller` in the mcd deployment. 
- Replicators in mcd will always fetch `primaryClusterId` from the configmap instead of relying on hardcoded value.


vivekzhere (approved):
LGTM

vinaybheri:

vinaybheri:

vinaybheri:

subhankarc (approved):

anoopjb (approved):

PR #1263, opened 2021-02-25
Added multiple region support for quota

anujbansal16:
Feature: HCPCFS-2886

This PR enhance the broker to support multiple regions. Enhancements are listed below:

- Extending the broker routes to support the region as a url parameter. 
   Example: `https://example-broker.com/region/eu10/v2/...`
- Based on the region provided as a part of url,  we extract the particular credential for quota validations.
- Tagging each sfserviceinstance based the the region so that it could be determined which region the service instance is from.

### DIFF
--- a/helm-charts/interoperator/values.yaml
+++ b/helm-charts/interoperator/values.yaml
@@ -28,6 +28,17 @@ broker:
     serviceDomain: https://my-tenant-onboarding.com
     username: quota_user
     password: quota_password
+    regions:
+      eu10:
+        oauthDomain: https://myauth-domain.com
+        serviceDomain: https://my-tenant-onboarding.com
+        username: quota_user
+        password: quota_password
+      eu20:
+        oauthDomain: https://myauth-domain.com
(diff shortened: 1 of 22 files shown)


helm-charts/interoperator/values.yaml
```diff
@@ -28,6 +28,17 @@ broker:
...
+      eu10:
+        oauthDomain: https://myauth-domain.com
+        serviceDomain: https://my-tenant-onboarding.com
+        username: quota_user
+        password: quota_password
+      eu20:
+        oauthDomain: https://myauth-domain.com
+        serviceDomain: https://my-tenant-onboarding.com
+        username: quota_user
+        password: quota_password
```
vivekzhere, line 41: Probably a good idea to comment out this section similar to image pull secrets in this file

anujbansal16: Done

subhankarc (approved):

jayeshjk (approved):
LGTM

PR #1286, opened 2021-03-19
Fix json error in Response Error

anujbansal16:
HCPCFS-2963

#### Details/Requirements: 
Currently, whenever the schema validation gets failed, an error response will be returned. We have used a error middleware which will return the error response in appropriate format based on the `Accept` header sent in the request.
Currently in the logic of our error response, we are setting the default format for the response to be in `json`.

Whenever any operation like provisioning etc. is initiated by the svcat , and that operation return error as a response, upon describing the instance it shows error like:

```shell
 svcat describe instance demo3
                                                                                                                                                                                                                                                                                                                                      
 Status:      Failed - Error provisioning ServiceInstance of ClusterServiceClass (K8S: "34731fb8- 7b84-5f57-914f-d3d55d793dd4" ExternalName: "postgresql-db") at ClusterServiceBroker "interoperator": Status: 400; ErrorMessage: <nil>; Description: <nil>; ResponseError: invalid character 's' looking for beginning of value @ 2021-02-26 09:35:53 +0000 UTC 
```

Upon debugging the issue it was found that, svcat doesn't send `Accept` header in the request.

According to the documentation of the `res.format(object)` method it was found that whenever the `Accept` header is not present in the request, the formatter invoke the first callback [Documentation.](http://expressjs.com/en/api.html#res.format) In our error response handler the first callback is of type `text`.

#### Findings:
The described error will occur when the expected response is `json` and actual was in `text`.

#### Fix
Changed the order of the formats in our error response handler i.e on [Line61](https://github.com/cloudfoundry-incubator/service-fabrik-broker/blob/63464875e8a4b9215ec85b503587b6eb73af14fb/broker/core/express-commons/src/middleware/index.js#L61)

```python
#previous ordering
const formats = options.formats || ['text', 'html', 'json']; 
#changed ordering
const formats = options.formats || ['json', 'text', 'html'];
```

As a result of above changes some of the UT's were getting failed because of mismatched response body.
As a quick fix currently I commented those matching.

### DIFF
--- a/broker/applications/extensions/test/acceptance/service-fabrik-api-2.0.backups.director.spec.js
+++ b/broker/applications/extensions/test/acceptance/service-fabrik-api-2.0.backups.director.spec.js
@@ -411,6 +411,11 @@ describe('service-fabrik-api-2.0', function () {
         });
         it('should return 410 Gone - when not found in both blobstore and apiserver', function () {
           const backupPrefix = `${space_guid}/backup`;
+          const expectedResponse = {
+            "description": "Backup does not exist or has already been deleted",
+            "error": "Gone",
+            "status": 410
+          };
           mocks.uaa.tokenKey();
           mocks.cloudController.getSpaceDevelopers(space_guid);
           mocks.apiServerEventMesh.nockGetResource(CONST.APISERVER.RESOURCE_GROUPS.BACKUP, CONST.APISERVER.RESOURCE_TYPES.DEFAULT_BACKUP,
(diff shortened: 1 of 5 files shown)


vinaybheri:

vinaybheri:

vinaybheri:

vinaybheri:

vivekzhere (approved):

vivekzhere (approved):

PR #1299, opened 2021-04-08
Recommended resource values for Large deployments

vivekzhere:
- Add a recomended set of resouce values to docs
- To be used on landscape with lots of instances

Bug: NGPBUG-146368


anoopjb (approved):

vivekzhere:
Moved the values inside the Interoperator.md doc

anoopjb (approved):

PR #1322, opened 2021-05-21
Create service_namespace via helm  if namespaced_separation is false

gowrisankar22:
*PR Contains*

- Right now, if the namespaced_separation is set to the false service instances created in the namespace mentioned in the `service_namespace` but this namespace should be created manually. This PR creates via helm automatically.
- Added another toggle  `create_services_namespace` to create the namespace.

### DIFF
--- a/helm-charts/interoperator/templates/namespace.yaml
+++ b/helm-charts/interoperator/templates/namespace.yaml
@@ -1,0 +1,6 @@
+{{- if not .Values.broker.enable_namespaced_separation }}
+apiVersion: v1
+kind: Namespace
+metadata:
+  name: {{ .Values.broker.services_namespace }}
+{{- end }}
--- a/helm-charts/interoperator/templates/namespace.yaml
+++ b/helm-charts/interoperator/templates/namespace.yaml
@@ -1,6 +1,8 @@
 {{- if not .Values.broker.enable_namespaced_separation }}
+{{- if .Values.broker.create_services_namespace }}
 apiVersion: v1
 kind: Namespace
 metadata:
   name: {{ .Values.broker.services_namespace }}
+{{- end }}
 {{- end }}
--- a/helm-charts/interoperator/values.yaml
+++ b/helm-charts/interoperator/values.yaml
@@ -14,6 +14,7 @@ broker:
   username: broker
   password: secret
   enable_namespaced_separation: true
+  create_services_namespace: false
   services_namespace: "services"
   allow_concurrent_operations: true
   allow_concurrent_binding_operations: true


anoopjb:
Looks good 👍 
The only downside is that if there is already a `services` namespace available and is not managed by helm, we cannot proceed.
```
➜  helm-charts git:(master) $ kubectl create namespace services
namespace/services created
➜  helm-charts git:(master) $ helm install --set cluster.host=xxxxxxxx --set broker.enable_namespaced_separation=false --set broker.services_namespace=services --namespace interoperator --wait interoperator interoperator

Error: rendered manifests contain a resource that already exists. 
Unable to continue with install: Namespace "services" in namespace "" exists and cannot be imported into the current release: invalid ownership metadata; 
label validation error: missing key "app.kubernetes.io/managed-by": must be set to "Helm"; 
annotation validation error: missing key "meta.helm.sh/release-name": must be set to "interoperator";
annotation validation error: missing key "meta.helm.sh/release-namespace": must be set to "interoperator"
```

gowrisankar22:
@anoopjb True, during the helm install, we have to provide the unique namespace but this is quite known. Helm upgrade should be ok.

anoopjb:
Yes. Helm upgrade should not have a problem. 👍 

Since Interoperator is stateless, it is possible to uninstall and then install the Interoperator without removing the service instances or binding. Some of the users might have this deployment strategy.

We feel may be we can introduce a toggle in `values.yaml`, something like `create_services_namespace: false`. Based on the toggle, we may choose to create the namespace.
@jayeshjk

gowrisankar22:
@anoopjb Make sense. I updated the PR with the changes.

jayeshjk (dismissed):

jayeshjk:
Please merge after pipeline is green

vinaybheri:

vinaybheri:

vinaybheri:

vinaybheri:

anoopjb (approved):

anoopjb:
I think there is a possibility of accidental deletion of namespace.
I got inspiration from https://github.com/helm/helm/issues/6794#issuecomment-589400266 and did some experiments.

I installed the interoperator with the changes,
```sh
$ helm install --set cluster.host=xxxx \
     --set broker.enable_namespaced_separation=false \
     --set broker.create_services_namespace=true \
     --set broker.services_namespace=services \
     --namespace interoperator --wait interoperator interoperator
```

It worked perfectly. The `services` namespace got created successfully.
The newly created service instances are all placed in the namespace.

```sh
bash-3.2$ kubectl get sfserviceinstances -n services
NAME                                   STATE       AGE   CLUSTERID
0ccfa7e0-d6da-4225-94df-4ce85a4f61e2   succeeded   40m   1
```

But we have a tricky situation during helm upgrade.
If the `create_services_namespace` is `false`, the rendered template will not have the namespace template.
Helm then deletes the namespace and thus terminating all the resources in it.
Thus the service instances in the `services` namespace is lost.
I feel this can happen, mostly accidentally, since we keep the `create_services_namespace` as `false` by default.

```sh
$ helm upgrade --set cluster.host=xxxx \
      --set broker.enable_namespaced_separation=false \
      --namespace interoperator --wait interoperator interoperator
```
```sh
kubectl get ns services 
NAME       STATUS        AGE
services   Terminating   61m
```
Another scenario that I can imagine is when downgrading to an older helm chart version where creating namespace is not implemented.
Looks like playing with namespaces is a risky business.

Let's dig a little deeper into this. I am parking this PR for further discussion.

jayeshjk:
Thanks @anoopjb for looking into it.
On a second look wrt above observation, I also feel that management of the services namespace should be kept independent of the interoperator deployment. 
The services namespace will contain metadata related to user service instances and it makes sense to separate it from stateless interoperator control plane.
@gowrisankar22 will it possible for you to try other workarounds for your scenario.

PR #1333, opened 2021-06-08
adding max attempts check in scheduled update job

jayeshjk:
- Currently, during the ServiceInstanceUpdate job, before triggering the actual update call, it is checked whether the instance is outdated or not. If it's not outdated the update is not triggered. 
- Once the actual update is triggered, the scheduled update job is rescheduled after some time(~30 mins). This can be seen as a retry mechanism. If the actual update didn't go through in earlier attempt, it will again be retried by this rescheduled job.
- This retrying stops after predefined threshold is reached. If this situation is reached, it would mean that, earlier updates did not go through successfully (as the deployment is still outdated).
- However, even in this case, the scheduled update is marked successful as the update trigger was successful. This PR handles this issue.
- If attempts exceed threshold, new update is not triggered and error is thrown.


vinaybheri:

saud89 (approved):

anoopjb (approved):

saud89 (approved):

PR #1404, opened 2021-08-18
Refresh Catalog Cache in Quota App

Pooja-08:
In the current code, if the given planId is not found in the catalog of quota_app we return 500- Internal server error which is a generic error. This PR modifies the code to instead return 404 NotFound in such case.
Resolves HCPCFS-3085.


anoopjb (approved):
LGTM

PR #1498, opened 2021-12-23
HCPCFS-3190 Minor improvement to avoid synchronous binding timeouts

jintusebastian:
- Removed duplicate watches on primary cluster
- Increased timeout period for replicator watches from 8 hours to 24 hours
- Increased controller runtime manager sync period from 10 hours to 24 hours

### DIFF
--- a/interoperator/pkg/constants/constants.go
+++ b/interoperator/pkg/constants/constants.go
@@ -28,7 +28,7 @@ const (
 
 	NamespaceLabelKey = "OWNER_INTEROPERATOR_NAMESPACE"
 
-	MultiClusterWatchTimeout = 28800 // 8 hours in seconds
+	MultiClusterWatchTimeout = 86400 // 24 hours in seconds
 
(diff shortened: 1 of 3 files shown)


interoperator/pkg/constants/constants.go
```diff
@@ -28,7 +28,7 @@ const (
 
 	NamespaceLabelKey = "OWNER_INTEROPERATOR_NAMESPACE"
 
-	MultiClusterWatchTimeout = 28800 // 8 hours in seconds
+	MultiClusterWatchTimeout = time.Hour * 24
```
anoopjb, line 31: I think the `time.Hour` resolution is on Nanoseconds. In that case the value will be  `3600000000000`.I go this impression from https://pkg.go.dev/time#Hour.

jintusebastian: Updated

anoopjb (approved):
LGTM

PR #1534, opened 2022-02-11
Remove finalizer for failed orphan binding

Pooja-08:
If the binding creation times out, broker tries to delete the binding. In some scenarios the deletion of such binding times out as well, leaving orphan bindings that need to be deleted manually since the binding still contains its finalizer. Since these bindings never got created, the user doesn't know about them and hence its safe to remove the finalizer to avoid having orphan bindings in the system.


anoopjb (approved):
Looks good to me. Please fix the eslint warnings.

PR #1549, opened 2022-02-24
library upgrade: Bump camelcase-keys version from  6.2.2 to 7.0.2

Pooja-08:


broker/applications/osb-broker/package.json
```diff
@@ -14,7 +14,7 @@
     "@sf/models": "*",
```
anoopjb, line 14: Since we include [@sf/eventmesh](https://github.com/cloudfoundry-incubator/service-fabrik-broker/blob/855027ec09081d71d8bdfeb8df5bc2ff94792684/broker/data-access-layer/eventmesh/package.json#L17), camelcase-keys will be indirectly included from there. I think it will be required to update in [eventmesh package.json](https://github.com/cloudfoundry-incubator/service-fabrik-broker/blob/855027ec09081d71d8bdfeb8df5bc2ff94792684/broker/data-access-layer/eventmesh/package.json#L17) as well.

anoopjb (changes requested):

broker/applications/osb-broker/package.json
Pooja-08: Thanks.. It was missed.. Done

anoopjb (approved):
LGTM 🚢

PR #1810, opened 2023-01-30
moving quota regions credentials to secret

pawan0306:
This is a security feature, here we are moving region based credentials to the secrets so that credentials shall not get exposed.

How we are doing it:

- Moving region details from values.yml excluding username and password to config
- Moving username and password to secrets of each region in a specific format to identify them
- Adding username and password in quota_app.yml as env so that it can be consumed


anoopjb (approved):
LGTM

PR #1890, opened 2023-07-04
[Metering] During plan update send start event after stop event

vinaybheri:
* During plan update the metering start event was sent before to metering stop event. As a result many instances after plan update got missing from the metering (escaped from charging).
* To fix this issue we have fixed the code to send stop event first and then start event during plan update also made sure that there is a gap 1 ms between each event.


anoopjb (approved):
LGTM :+1:

PR #1914, opened 2023-09-12
fix and reset service instance count in sfcluster

jintusebastian:
- fixed the issue in instance counter controler flow
- fixed the instance count in sfcluster in case of mismatch

### DIFF
--- a/interoperator/controllers/schedulers/updateinstancecount.go
+++ b/interoperator/controllers/schedulers/updateinstancecount.go
@@ -75,7 +75,7 @@ func updateInstanceCount(kubeConfig *rest.Config, scheme *runtime.Scheme, mapper
 		}
 		for _, sfserviceinstance := range sfserviceinstances.Items {
 			for _, finalizer := range sfserviceinstance.Finalizers {
-				if finalizer == constants.SFServiceInstanceCounterFinalizerName {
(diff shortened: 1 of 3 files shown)


vinaybheri (approved):

interoperator/controllers/schedulers/updateinstancecount.go
```diff
@@ -0,0 +1,110 @@
...
+	}
+
+	ctx := context.Background()
+	sfserviceinstances := &osbv1alpha1.SFServiceInstanceList{}
+	instanceOptions := &kubernetes.ListOptions{}
+
+	// Calculate the expected service instance count for each cluster
+	instanceCount := make(map[string]int)
+	for more := true; more; more = (sfserviceinstances.Continue != "") {
+		err := client.List(ctx, sfserviceinstances, instanceOptions, kubernetes.Limit(constants.ListPaginationLimit),
```
anoopjb, line 71: It is nice that we are only reading 100 (constants.ListPaginationLimit) sfserviceinstances at a time 👍 . This should prevent the possibility of an out of memory error in heavy landscapes. I think it will be a good idea to monitor memory usage after releasing this feature.
Here is a guide from Golang garbage collector, https://tip.golang.org/doc/gc-guide#Optimization_guide

anoopjb (approved):
LGTM 👍

